PR #8, opened 2015-04-30
Add connection pool

jakewins:


driver/src/main/java/org/neo4j/driver/internal/connector/socket/ChunkedInput.java
```diff
@@ -232,7 +232,19 @@ private ByteBuffer readNextChunk() throws IOException
     {
         int chunkSize = readChunkSize();
         byte[] buffer = new byte[chunkSize];
-        in.read( buffer );
+        int read = in.read( buffer );
+
+        if(read == -1)
```
zhenlineo: A question about the purpose of doing this `-1` test. I do not understand how this `-1` works in Socket InputStream. I would expect `int read = in.read( buffer );` will block until we get something from InputStream unless an exception is thrown. However if we get an exception, we will never get to this `if`.

driver/src/main/java/org/neo4j/driver/internal/connector/socket/ChunkedInput.java
```diff
@@ -242,9 +254,29 @@ private ByteBuffer readNextChunk() throws IOException
...
-        int chunkSize = ByteBuffer.wrap( buffer ).getShort();
-        return chunkSize;
+        int idx = 0;
+
+        for(;;)
+        {
+            int read = in.read( buffer, idx, 1 );
+            if ( read == -1 )
+            {
+                return throwConnectionLost();
```
jakewins: This return shouldn't be here

PR #9, opened 2015-05-12
Adding license headers

zhenlineo:


driver/src/main/java/org/neo4j/Neo4j.java
```diff
@@ -1,19 +1,19 @@
 /**
  * Copyright (c) 2002-2015 "Neo Technology,"
  * Network Engine for Objects in Lund AB [http://neotechnology.com]
- * <p>
+ *
```
chrisvest: Intellij won't add HTML tags when the license header comment is not a javadoc comment.

PR #10, opened 2015-05-13
ParameterIT failure

zhenlineo:
This PR is designed to help figure out the test failure on TC but not locally. 
On TC, ParameterIT fails with timeout to read next chunk. However this never happens locally. A possible cause is that we pack and unpack strings containing special characters wrongly. This PR should be reverted when we find and fix the bug.


zhenlineo:
I am doing a test on TC and will revert this PR once I find the bug, So now I merge this PR secretly.

PR #16, opened 2015-05-20
More logging

zhenlineo:
Added different levels of logging:
 o FINE      - Message level - the messages exchanged over connection
 o FINEST - Chunked data level - the chunked data exchanged over connection

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/logging/DevNullLogger.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/logging/DevNullLogger.java
@@ -19,10 +19,18 @@
  */
 package org.neo4j.driver.internal.logging;
 
+import java.util.logging.Level;
+
 import org.neo4j.driver.internal.spi.Logger;
 
 public class DevNullLogger implements Logger
 {
+    @Override
+    public void log( Level level, String message )
+    {
+
+    }
+
     @Override
     public void debug( String message )
     {
--- a/driver/src/main/java/org/neo4j/driver/internal/logging/JULogger.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/logging/JULogger.java
@@ -26,21 +26,22 @@ import org.neo4j.driver.internal.spi.Logger;
 public class JULogger implements Logger
 {
     private final java.util.logging.Logger delegate;
-    private final boolean debugEnabled;
 
     public JULogger( String name )
     {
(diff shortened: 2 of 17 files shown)


driver/src/main/java/org/neo4j/driver/internal/logging/JULogger.java
```diff
@@ -26,21 +26,22 @@
 public class JULogger implements Logger
 {
     private final java.util.logging.Logger delegate;
-    private final boolean debugEnabled;
 
     public JULogger( String name )
     {
         delegate = java.util.logging.Logger.getLogger( name );
-        debugEnabled = delegate.isLoggable( Level.FINE );
```
zhenlineo: If we keep this, then we only have the chance to enable and disable debug mode during initialization. I somehow added a FINEST level so this check seems not so meaningful at all.

jakewins: Fair enough. We should re-visit this later on then, not doing this check is *very* expensive, each debug log call effectively creates synchronization, this was a big bottleneck in the neo4j server a few years back.

driver/src/main/java/org/neo4j/driver/internal/logging/DevNullLogger.java
```diff
@@ -19,11 +19,19 @@
...
 package org.neo4j.driver.internal.logging;
 
+import java.util.logging.Level;
+
 import org.neo4j.driver.internal.spi.Logger;
 
 public class DevNullLogger implements Logger
 {
     @Override
+    public void log( Level level, String message )
```
jakewins: I'm not a fan of exposing JUL log level through this interface, the intent was to hide JUL behind it, so that we can plug in other log frameworks and testing if we like. Not a biggie, but perhaps we can follow up and add a `trace` method or something, rather than expose programmatic log level here.

driver/src/main/java/org/neo4j/driver/internal/logging/JULogger.java
chrisvest: You could investigate calling through MethodHandles from MutableCallsites. I think they're able to inline (if the MethodHandle reference is static final) so the cost of changing the log level (which is a quite rare event) would be a safe-point for deoptimising. I'm not completely sure if this is true, though.

PR #17, opened 2015-05-25
Global logging

zhenlineo:
A user could enable debug mode by setting system properties "org.neo4j.driver.loggingEnabled=true" and "org.neo4j.driver.loggingLevel=ALL".
This change makes debugging outside driver possible.


lassewesth:
@zhenlineo instead of using 'neo4j' as the key, why not something like 'org.neo4j.driver' - it is not Neo4j the product that you change logging level for, it is for the driver

zhenlineo:
This PR will handle this problem https://github.com/neo4j/neo4j-java-driver/pull/24

PR #22, opened 2015-06-03
Swap to channels API the whole way back to socket connect

jakewins:


driver/src/test/java/org/neo4j/driver/internal/connector/socket/SocketClientTest.java
```diff
@@ -26,14 +27,15 @@
...
 
 public class SocketClientTest
 {
     @Rule
     public ExpectedException exception = ExpectedException.none();
 
-    @Test
+    // TODO: This is not possible with blocking NIO channels, unless we use inputStreams, but then we can't use
+    // off-heap buffers. We need to swap to use selectors, which would allow us to time out.
+    @Test @Ignore
```
zhenlineo: Well, there is a trick http://stackoverflow.com/questions/2866557/timeout-for-socketchannel-doesnt-work, but this is not so important for now.

jakewins: There's a deal about that that though. Incoming data on the network can be read into "regular" ram using a system call, right. The java heap is not "regular" ram - since the GC will be moving things around in there, we can't ask the OS to put stuff directly in the heap. Instead, if we use regular heap buffers (eg. `new byte[]` or `ByteBuffer.allocate()` instead of `ByteBuffer.allocateDirect()`) the JVM will malloc a temporary buffer outside the heap for each read and each write, tell the OS to read/write to that, and then copy that buffer byte-by-byte onto the heap.

To illustrate, lets say we ask `InputStream` to read into a `byte[]` for us, from the network buffer. What will happen is something like:

    Us: ourBuffer = new byte[1024]
    Us: input.read( ourBuffer )
    JVM: tmp = malloc( 1024 )
    JVM: os.copyFromNetworkBufferInto( tmp )
    JVM: for byte in tmp: 
                ourBuffer.add( byte );
    JVM: free(tmp)

This gets even worse, since we're using `ByteBuffer.allocateDirect()` at the top layer, so that `byte[]` will *also* be a temporary buffer. Once the JVM has copied all the data into that buffer, it'll then copy it back byte-by-byte off the heap again into our direct buffer.

Beyond the direct overhead of all this copying, imagine the overhead in GC from allocating these big temporary buffers over and over, it'll cause heap fragmentation quite badly.

Contrast that to using the Channels API all the way through:

    Us: ourBuffer = ByteBuffer.allocateDirect( 1024 );
    Us: ch.read( ourBuffer )
    JVM: os.copyFromNetworkBufferInto( ourBuffer )

You'll see these optimizations here: http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/sun/nio/ch/IOUtil.java

Notice how, in the read and write methods, it checks if the buffer is a "DirectBuffer", and if it does it bypasses all the copying. That's what we want :)

PR #24, opened 2015-06-10
Rename Neo4j to Driver and introduce Config in Driver

zhenlineo:
Renamed the top class for database connection management from Neo4j to Driver. All the functions to establish a session with a Neo4j instance are now in Driver.
Introduced configuration in Driver for config per connection pool created in Driver.
Refined logging system - Only handle debug and trace when they are enabled


driver/src/main/java/org/neo4j/driver/internal/connector/socket/SocketClient.java
```diff
@@ -146,7 +140,9 @@ private SocketProtocol negotiateProtocol() throws IOException
 
         while ( buf.remaining() > 0 )
         {
+            int start = buf.position();
             channel.read( buf );
+            logger.trace( "S: " + BytePrinter.hex( buf, start, buf.position() ) );
```
jakewins: We can't do these like this - this will print every bit of the buffer even if `trace` is off. 

I'd suggest we do this the same way you did it with the Input/Output streams - introduce classes that implement WritableByteChannel/ReadableByteChannel, and can wrap the regular channels. Then, when we create the real channels, we do something like:

    if( logger.isTraceEnabled() )
    {
        input = new LoggingInputChannel( logger, input );
        output = new LoggingOutputChannel( logger, output );
    }

That way, we don't have to intermingle these log statements with the rest of the code, and there's zero overhead if trace logging is not enabled.

driver/src/main/java/org/neo4j/driver/internal/connector/socket/SocketResponseHandler.java
```diff
@@ -75,7 +74,7 @@ public void handleRecordMessage( Value[] fields )
             collector.record( fields );
         }
 
-        logger.debug( new RecordMessage( fields ).toString() );
+        logger.debug( "S: RecordMessage{%s}", Arrays.asList( fields ) );
```
jakewins: Same with these as above - we could move this to a separate MessageHandler, LoggingMessageHandler, and only include that if debug logging is enabled.

driver/src/main/java/org/neo4j/driver/internal/pool/StandardConnectionPool.java
```diff
@@ -71,25 +65,19 @@
...
     private final Clock clock;
+    private final Config config;
 
-    public StandardConnectionPool( Logging logging )
-    {
-        this( loadConnectors( logging ) );
-    }
-
-    public StandardConnectionPool( Collection<Connector> conns )
+    public StandardConnectionPool( Config config )
```
jakewins: I would actually vote for not letting Config go all the way down here - have the surface, `Driver`, ask for Config, but past that pass in component-specific arguments. So rather than asking for config here, we ask for the parameters we need (`connectionsPerDatabase` and `minIdleTimeBeforeConnectionTest`). Config would only be used by `Driver`, keeping the rest of the system from having to know about all the other config options, and making it possible to test these components without creating a config object.

driver/src/main/java/org/neo4j/driver/internal/util/InputStreams.java
```diff
@@ -32,18 +33,20 @@
...
             {
-                throw new ClientException(
+                throw new ClientException( String.format(
                         "Connection terminated while receiving data. This can happen due to network " +
-                        "instabilities, or due to restarts of the database. Expected " + into.limit() + "bytes, " +
-                        "recieved " + BytePrinter.hex( into ) + "." );
+                        "instabilities, or due to restarts of the database. Expected %s bytes, received %s.",
+                        into.limit(), BytePrinter.hex( into ) ) );
             }
+            logger.trace( "S: " + BytePrinter.hex( into, start, read ) );
```
jakewins: Same as with the channel thing - if we put the logging on the channel level, we wont need this

driver/src/main/java/org/neo4j/Driver.java
```diff
@@ -0,0 +1,192 @@
...
+ * (The default connection pool size could be found in {@code Config}.)
+ * A connection could be returned to the connection pool by invoking {@link Session#close() close} when the work over
+ * a session is done.
+ */
+public class Driver implements AutoCloseable
+{
+    /**
+     * A static class to create a driver
+     */
+    public static class GraphDatabase
```
jakewins: I would want to move this out, so it's not an inner class; that way we get:

    Driver driver = GraphDatabase.driver(..);

instead of:

    Driver driver = Driver.GraphDatabase.driver(..);

driver/src/main/java/org/neo4j/driver/internal/pool/StandardConnectionPool.java
lassewesth: If you do create longish parameter lists, consider making them microtypes so that you do not get lost in ints and Strings

zhenlineo: I kind of disagree of passing real parameters instead of Config. Because how we use Config is something like:
```
ClassA
{
  String param1;
  ClassB b;
  public ClassA( Config config )
  {
    param1 = config.param1;
    classB = new ClassB( config )
  }
}
ClassB 
{
  String param2;
  public ClassB( Config config )
  {
    param2 = config.param2;
  }
}
```

lassewesth: Why not inject class B?

driver/src/main/java/org/neo4j/GraphDatabase.java
```diff
@@ -0,0 +1,80 @@
...
+ * // Establish a session with a Neo4j instance
+ * Session session = driver.session();
+ * ** Do some work with the database... **
+ *
+ * // Release all the resources
+ * driver.close();
+ * }
+ * </pre>
+ */
+public class GraphDatabase
```
cleishm: This will conflict with https://github.com/neo4j/neo4j/pull/4674 -- and I'm not sure it's the best place anyway. Why not use factory methods in `org.neo4j.Driver` ?

driver/src/main/java/org/neo4j/driver/Config.java
```diff
@@ -0,0 +1,118 @@
...
+ *               ** .setProperties() **
+ *                  .build();
+ * }
+ * </pre>
+ */
+public class Config
+{
+
+    /** A config with all default settings */
+    public static final Config DEFAULT_CONFIG = Config.builder().build();
```
cleishm: `.builder().build()` is kind of awkward. Might want to think about the naming there.

driver/src/main/java/org/neo4j/driver/Config.java
```diff
@@ -0,0 +1,118 @@
...
+        this.soTimeout = soTimeout;
+    }
+
+    public static ConfigBuilder builder()
+    {
+        return new ConfigBuilder();
+    }
+
+    public static class ConfigBuilder
+    {
```
cleishm: These inner-class builders should have a private constructor (something Misha called me out on recently)

driver/src/main/java/org/neo4j/driver/internal/pool/StandardConnectionPool.java
zhenlineo: Well, that was a bad example, so what about this?
```
ClassA
{
  String param1;
  String paramThatAShouldNotKnow;
  public ClassA( String param1, String param2 )
  {
    this.param1 = param1;
    this.paramThatAShouldNotKnow = param2;
  }
  void method()
  {
    ClassB classB = new ClassB( paramThatAShouldNotKnow );
  }
}
ClassB 
{
  String param2;
  public ClassB( String param2 )
  {
    this.param2 = param2;
  }
}

lassewesth: Same thing: creation of class B doesn;t belong in class A precisely because of what you said: it is not the responsibility of A to know about creation of class B. In particular wrt config, but in general also

driver/src/main/java/org/neo4j/driver/Config.java
zhenlineo: Added constructor

driver/src/main/java/org/neo4j/driver/Config.java
zhenlineo: I am following the same naming patten used in park-bench. Could you suggest some good names for these methods if you strongly would like to change them? I feel they are okay to me

cleishm: Here's some suggestions:

For this specific case (the default), you could introduce a factory method for this case, something like `Config.empty()` or `Config.default()`? I'd also suggest using that method wherever it's needed, instead of a public static field.

I'd suggest renaming `Config#builder()` to `Config#build()`, so that it's a verb -- indicating the start of a sentence like "build with blah, with blah, and with blah". Likewise, perhaps use `withXX()` vs `setXX()` (we do this in some [other builders](https://github.com/neo4j/neo4j/blob/2.3/community/logging/src/main/java/org/neo4j/logging/FormattedLogProvider.java#L73)).

Finally, you might like to rename `ConfigBuilder#build()` to `ConfigBuilder#toConfig()`, just so that it reads something like `Config.build().withLogging(...).withConnectionPoolSize(...).toConfig()`, which fits the sentence structure "build with logging, and with pool size, and then give me the Config object".

Just suggestions :)

jakewins: FWIW, I'm +1 on all those suggestions.

PR #30, opened 2015-06-22
Roll back tx on client error

zhenlineo:
This PR is a response to the PR for server neo4j/neo4j#4771
When a `client error` is returned from the server, the transaction will be left open and can be used again after the `FAILURE` has been acknowledged. This is mainly to support user-driven queries, where a database administrator may have built up a large transaction, and we do not want a simple spelling mistake to roll it all back.


driver/src/test/java/org/neo4j/driver/integration/ErrorIT.java
```diff
@@ -52,7 +52,7 @@ public void shouldThrowHelpfulSyntaxError() throws Throwable
     }
 
     @Test
-    public void shouldNotAllowUsingTransactionAfterError() throws Throwable
+    public void shouldNotAllowUsingTransactionAfterRecoverableError() throws Throwable
```
nigelsmall: Is this name the right way round? A transaction should surely be _usable_ after a _recoverable_ error but _unusable_ after an _unrecoverable_ one...

zhenlineo: Well, currently we mark all `ClientError` as recoverable errors, which indicates the transaction is still open and a user could do something in the same transaction. The method name could be changed to anything for sure, such as `shouldNotAllowUsingTxAfterClientError`

PR #39, opened 2015-07-08
Only read message ending after all bytes in message content have been read

zhenlineo:
Fix a bug where the ChunkedInput will try to read message ending 00 00 from channel when more bytes left unread in the message content buffer.
Refined DumpMessage error message when decoding messages from hex string to plain text

### DIFF
--- a/driver/src/test/java/org/neo4j/driver/internal/connector/socket/ChunkedInputTest.java
+++ b/driver/src/test/java/org/neo4j/driver/internal/connector/socket/ChunkedInputTest.java
@@ -26,8 +26,10 @@ import java.nio.channels.Channels;
 import java.nio.channels.ReadableByteChannel;
 import java.util.Arrays;
 
+import org.neo4j.driver.exceptions.ClientException;
 import org.neo4j.driver.util.RecordingByteChannel;
 
+import static junit.framework.Assert.fail;
 import static junit.framework.TestCase.assertEquals;
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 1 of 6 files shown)


driver/src/test/java/org/neo4j/driver/internal/connector/socket/ChunkedInputTest.java
```diff
@@ -150,4 +152,30 @@ public void canReadAllIntegerSizes() throws Exception
         assertEquals( 0,                 in.readInt() );
         assertEquals( Integer.MIN_VALUE, in.readInt() );
     }
-}
\ No newline at end of file
+
+    @Test
+    public void shouldNotReadMessageEndingWhenByteLeftInBuffer()
```
jakewins: This test is awesome, no need to change it, but noting that Nigel showed me an even awesomer way to write these kinds of tests, using the `ExpectedException` JUnit rule, you can do something like;

    @Rule public ExpectedException exception = ExpectedException.none();

    @Test public void blah() throws Exception {
        // given
        some stuff

        // expect
        exception.expect( "It didn't work oh no");

        // when
        ch.messageBoundaryHook().run();
    }

zhenlineo: The experience of using `exception.expect` is not so good for me. Sometimes if we have many places that might throw an exception then the `exception.expect` could catch any of them. So I usually just use the simple `try-fail-catch-message` to make sure that the exception is the one I want and it really comes from the line I want.

lassewesth: Massive +1 to that @zhenlineo

jakewins: In this case they are equivalent though - you only have a single line after the expect, so the test will fail if anything throws an exception before that, and it'll fail if that line throws any exception other than the one you expect. True though that there if there are cases where it can be unclear, try/finally is safer.

PR #41, opened 2015-07-30
TLS connections with the server

zhenlineo:
Enabled the driver to establish TLS (handshake) and close TLS connections with the server
Ensured basic data transfer using TLS
Enabled the user to choose from no-TLS, TLS with `trust-first-cert`, and TLS with `only-trust-cert-provided`.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/Config.java
+++ b/driver/src/main/java/org/neo4j/driver/Config.java
@@ -50,12 +50,16 @@ public class Config
     /** Connections that have been idle longer than this threshold will have a ping test performed on them. */
     private final long idleTimeBeforeConnectionTest;
 
+    private final boolean isTLSEnabled;
+
     private Config( ConfigBuilder builder )
     {
         this.logging = builder.logging;
 
         this.connectionPoolSize = builder.connectionPoolSize;
         this.idleTimeBeforeConnectionTest = builder.idleTimeBeforeConnectionTest;
+
+        this.isTLSEnabled = builder.isTLSEnabled;
     }
 
     /**
@@ -86,6 +90,15 @@ public class Config
         return idleTimeBeforeConnectionTest;
     }
 
+    /**
+     * If TLS is enabled in all socket connections
+     * @return
+     */
+    public boolean isTLSEnabled()
+    {
+        return isTLSEnabled;
+    }
+
     public static ConfigBuilder build()
     {
         return new ConfigBuilder();
@@ -107,6 +120,7 @@ public class Config
         private Logging logging = new JULogging( Level.INFO );
         private int connectionPoolSize = 10;
         private long idleTimeBeforeConnectionTest = 200;
+        public boolean isTLSEnabled = false;
 
         private ConfigBuilder()
         {
@@ -145,6 +159,16 @@ public class Config
         public ConfigBuilder withMinIdleTimeBeforeConnectionTest( long milliSecond )
         {
             this.idleTimeBeforeConnectionTest = milliSecond;
+            return this;
+        }
+
+        /**
+         * Enable TLS in all connections with server
+         * @param value
+         */
+        public ConfigBuilder withTLSEnabled( boolean value )
+        {
+            this.isTLSEnabled = value;
             return this;
         }
 
--- a/driver/src/main/java/org/neo4j/driver/Config.java
+++ b/driver/src/main/java/org/neo4j/driver/Config.java
@@ -18,6 +18,7 @@
  */
 package org.neo4j.driver;
 
+import java.io.File;
 import java.util.logging.Level;
 
 import org.neo4j.driver.internal.logging.JULogging;
@@ -51,6 +52,7 @@ public class Config
     private final long idleTimeBeforeConnectionTest;
 
     private final boolean isTLSEnabled;
+    private final File trustedCertificate;
 
     private Config( ConfigBuilder builder )
     {
@@ -60,6 +62,7 @@ public class Config
         this.idleTimeBeforeConnectionTest = builder.idleTimeBeforeConnectionTest;
 
         this.isTLSEnabled = builder.isTLSEnabled;
+        this.trustedCertificate = builder.trustedCertificate;
     }
 
     /**
@@ -99,6 +102,15 @@ public class Config
         return isTLSEnabled;
     }
 
+    /**
+     * Return the trusted certificate file. If it is not specified, then this method will return null.
+     * @return
+     */
+    public File trustedCertificate()
+    {
+        return trustedCertificate;
+    }
+
     public static ConfigBuilder build()
     {
         return new ConfigBuilder();
@@ -120,7 +132,8 @@ public class Config
         private Logging logging = new JULogging( Level.INFO );
         private int connectionPoolSize = 10;
         private long idleTimeBeforeConnectionTest = 200;
-        public boolean isTLSEnabled = false;
+        private boolean isTLSEnabled = false;
+        private File trustedCertificate = null;
 
         private ConfigBuilder()
         {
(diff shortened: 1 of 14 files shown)


driver/src/main/java/org/neo4j/driver/internal/connector/socket/SSLSocketChannel.java
```diff
@@ -0,0 +1,518 @@
...
+     * Add the server cert to the client's trusted cert list.</li>
+     * </ul><p>
+     * For the 2nd or 3rd option, what we could do is to add the cert to the trusted store used by the client. The
+     * trusted store could be created by {@code keytool} with the server cert. The trusted store could be loaded here
+     * or be set with {@code javax.net.ssl.trustStore} system property.
+     *
+     * @throws GeneralSecurityException
+     * @throws IOException
+     */
+    private void initSSLContext() throws GeneralSecurityException, IOException
```
lassewesth: @zhenlineo I am probably reading this wrong, but it looks like you initialise the context to never have keys, and to never bother with trust. Does that get configured later somehow? It there a test that proves users _can_ use keys and trust chains?

driver/src/main/java/org/neo4j/driver/Config.java
```diff
@@ -149,6 +163,16 @@ public ConfigBuilder withMinIdleTimeBeforeConnectionTest( long milliSecond )
         }
 
         /**
+         * Enable TLS in all connections with server
+         * @param value
+         */
+        public ConfigBuilder withTLSEnabled( boolean value )
```
lassewesth: @zhenlineo From what we discussed, I thought there was going to be three levels: full TLS, TLS without trust, or no TLS at all. Did I misunderstand?

driver/src/main/java/org/neo4j/driver/internal/connector/socket/SSLSocketChannel.java
zhenlineo: I have not yet add cert-trust handling in this pr. Your understanding of the current behaviour is 100% correct. The cert-trust part is coming soon.

driver/src/main/java/org/neo4j/driver/Config.java
zhenlineo: We will probably only have two levels: no TLS, TLS trust-on-first-use. Our server also use the same strategy, so does the client.

jakewins: I like the combination of your two comments: Full TLS (never trust certificates from unknown CAs), Trust-on-first-use (default), and off.

driver/src/main/java/org/neo4j/driver/internal/connector/socket/SSLSocketChannel.java
```diff
@@ -0,0 +1,455 @@
...
+                    // TODO
+                    throw new ClientException( "Failed to enlarge a buffer: plainIn.size = " + oldSize +
+                                               ", cipherIn.size = " + cipherIn.capacity() +
+                                               ". The maximum size for a buffer is " + bufferMax );
+                }
+                int newSize = Math.min( oldSize * 2, bufferMax );
+                ByteBuffer newPlainIn = ByteBuffer.allocateDirect( newSize );
+                bufferCopy( plainIn, newPlainIn );
+                plainIn = newPlainIn;
+                logger.debug( "Enlarged application input buffer from " + oldSize + " to " + newSize +
```
jakewins: Since this is in a super low-level loop of the driver, could we do an explicit check for isDebugEnabled here before we construct the string? Otherwise we'll create a stringBuilder, chararray and a string object twice every time a network packet is pulled in

driver/src/main/java/org/neo4j/driver/internal/connector/socket/TrustOnFirstUseTrustManager.java
```diff
@@ -0,0 +1,167 @@
...
+class TrustOnFirstUseTrustManager implements X509TrustManager
+{
+    /**
+     * A list of pairs (known_server_ip, certificate) are stored in this file.
+     * When establishing a SSL connection to a new server, we will save the server's IP and its certificate in this
+     * file.
+     * Then when we try to connect to a known server again, we will authenticate the server by checking if it provides
+     * the same certificate as the one saved in this file.
+     */
+    static final String KNOWN_CERTS_FILE_PATH = "known_certs";
```
jakewins: We should think some on where we put this file.. putting it directly in the working dir is probably the most sensible option, assuming we then provide a mechanism to change the location (eg. using Config or something). The alternative would be to put it in `System.getProperty("user.home") + "/.neo4j/known_certs"` or some such. 

I'd suggest that independent of where we put it, we should have a default name that makes it clear what it is. `neo4j_known_hosts` would be stronger in signalling that it's for Neo and that it's basically the same as the known_hosts file for SSH. To be further helpful, having a comment on the first line in the file to explain why it's there might be useful;

    # This file contains trusted certificates for Neo4j servers, it's created by Neo4j drivers. 
    # You can configure the location of this file by setting the NEO4J_TRUSTED_HOSTS_PATH environment var

Second, permissions for this file become important. I'm not sure how good java is at cross-platform permissions here, but if there's some way we can make the file readable/writeable only by the user that created the file, that'd be brilliant.

jakewins:
@zhenlineo this looks absolutely marvelous, great job. I had some thoughts/comments inline, let me know what you think.

driver/src/main/java/org/neo4j/driver/Config.java
lassewesth: I don't think first-trust should be default. It seems a special snowflake. Is there prior art?

Also I think turning off TLS and relying on firewalls or AWS security groups is a valid use case.

akollegger: We've learned that irritating users is detrimental to adoption and can even increase friction with our allies, like driver authors. There is always a balance between security and convenience. 

First-trust is a nice alternative to manually propagating certificates. That's you shouting a password to me across a room filled with people we both trust, which I can then take with me to use in a setting where I don't trust the neighbors. More concretely, I can connect to Neo4j while I'm at work with little fuss, then take my laptop home and still feel comfortable that I'm not compromising business information.

That seems pretty nice. 

Equally convenient would be no-TLS, but then I'd have no benefits. Less convenient would be full-TLS, but that'd be most secure. 

Turning off TLS could have lots of useful scenarios, including a client-side which does not have write permission to a file system (which seems to be a requirement), and happens to not care about security.

driver/src/main/java/org/neo4j/driver/internal/connector/socket/TrustOnFirstUseTrustManager.java
akollegger: +1 Great suggestions. 

Is it possible for a client to connect using TLS _without_ having a writable filesystem available?

driver/src/main/java/org/neo4j/driver/internal/connector/socket/SSLSocketChannel.java
```diff
@@ -0,0 +1,493 @@
...
+                {
+                    throw new ClientException(
+                            String.format( "Failed ro enlarge application buffer from %s to %s, as the maximum " +
+                                           "buffer size allowed is %s. The content in the buffer is: %s\n",
+                                    curAppSize, newAppSize, appSize * 2, BytePrinter.hex( plainIn ) ) );
+                }
+                ByteBuffer newPlainIn = ByteBuffer.allocateDirect( newAppSize );
+                newPlainIn.put( plainIn );
+                plainIn = newPlainIn;
+                logger.debug( "Enlarged application input buffer from %s to %s. " +
```
zhenlineo: I did not move this under `if( logger.debugEnabled() )` as we expect we will only get `OK` when invoking `wrap` and `unwrap`. If we got this debug info, it is probably an indication of a potential error in the code, e.g. forgot to clean the buffer after use etc.

PR #42, opened 2015-08-06
Update driver to use new compact path format

jakewins:


driver/src/main/java/org/neo4j/driver/internal/util/CertificateTool.java
```diff
@@ -147,6 +163,43 @@ public static void loadX509Cert( Certificate cert, String certAlias, KeyStore ke
     {
         keyStore.setCertificateEntry( certAlias, cert );
     }
+
+    private static X509Certificate generateSelfSignedCertificate(X500Name name, int bits, long validSecs)
```
zhenlineo: Not sure why we build this method. If it is to remove the warning while compiling mvn that `XX is internal proprietary API and may be removed in a future release`, then I think all the classes under `sun.security.x509` might not be supported in future versions. So this change will not help. But what we could do is to move this `genX509Cert` method into a test class as generating a cert is only used in tests. Then we might use `bouncycastle`

jakewins: It was a java7/java8 issue - the original class used here had been moved in Java 8, so it can't be compiled on Java 8. +1 on not using the internal classes though, this was just a quick fix. 

However, since this is just for testing, maybe we can follow up with a commit that moves this tool to tests, and we can pop in a test dependency that can generate these certificates?

driver/src/main/java/org/neo4j/driver/internal/messaging/PackStreamMessageFormatV1.java
```diff
@@ -484,76 +522,111 @@ private Value unpackValue() throws IOException
...
+            // Path sequence
+            int length = (int) unpacker.unpackListHeader();
+
+            // Knowing the sequence length, we can create the arrays that will represent the nodes, rels and segments in their "path order"
+            Path.Segment[] segments = new Path.Segment[length / 2];
+            Node[] nodes = new Node[segments.length + 1];
+            Relationship[] rels = new Relationship[segments.length];
+
+            Node prevNode = uniqNodes[0], nextNode; // Start node is always 0, and isn't encoded in the sequence
+            nodes[0] = prevNode;
```
zhenlineo: Proposal to the fix for test failure MessageFormatTest.shouldGiveHelpfulErrorOnMalformedNodeStruct. Though I do not totally understand the math to calculate the size of array here.

PR #47, opened 2015-09-07
Using TLSAuthenticationConfig

zhenlineo:
to select using either full-auth or trust-on-first-use to authenticate the server when establishing TLS connections

### DIFF
--- a/driver/src/test/java/org/neo4j/driver/integration/SSLSocketChannelIT.java
+++ b/driver/src/test/java/org/neo4j/driver/integration/SSLSocketChannelIT.java
@@ -22,13 +22,19 @@ import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
 import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
 import java.io.IOException;
+import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.URI;
 import java.nio.channels.SocketChannel;
 import java.security.cert.X509Certificate;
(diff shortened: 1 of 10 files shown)


driver/src/test/java/org/neo4j/driver/integration/SSLSocketChannelIT.java
```diff
@@ -92,22 +111,73 @@ public void shouldPerformTLSHandshake() throws Throwable
...
+            sslChannel = new SSLSocketChannel( "localhost", 7687, channel, mock( Logger.class ),
+                    usingKnownCerts( knownCerts ) );
+            sslChannel.close();
+        }
+        catch ( SSLHandshakeException e )
+        {
+            assertEquals( "General SSLEngine problem", e.getMessage() );
+            assertEquals( "General SSLEngine problem", e.getCause().getMessage() );
+            assertTrue( e.getCause().getCause().getMessage().startsWith(
+                    "The certificate received from the server is different from the one we've known in file" ) );
```
jakewins: From the perspective of a user that is not well-versed with TLS or trust-on-first-use, it would be awesome to make tailor this message to helping the user with how to fix the problem. Eg, could we change this message to be something more along the lines of:

    Unable to connect to neo4j at `neo4j://blahblah`, because the encryption certificate the server uses has changed. This is a security feature to protect against man-in-the-middle attacks. 

    If you trust the certificate the server uses now, simply remove the line that starts with <blah> in the file <bleh>. The certificate used by the server is: <cert info>.

I know that's a really long exception message, but I don't know that that's really an issue - independent of which, main point being that we give the user a hand in explaining how to resolve it.

jakewins:
Looks really good! Left a comment on expanding an error message

PR #48, opened 2015-09-08
Introduce javadoctest, fix compile errors etc. in javadocs

jakewins:


driver/src/test/java/org/neo4j/driver/util/Neo4jRunner.java
```diff
@@ -83,9 +83,10 @@ public Neo4jRunner() throws IOException
         {
             if ( neo4jHome.exists() )
             {
-                System.out.println( "Found an old Neo4j server in: " + neo4jHome.getAbsolutePath() + ". Deleting to " +
-                                    "use a new one." );
-                FileTools.deleteRecursively( neo4jHome );
+                return;
+//                System.out.println( "Found an old Neo4j server in: " + neo4jHome.getAbsolutePath() + ". Deleting to " +
+//                                    "use a new one." );
+//                FileTools.deleteRecursively( neo4jHome );
```
zhenlineo: If we want to default to use the existing server rather than downloading from the server again, we should make changes to revert this commit https://github.com/neo4j/neo4j-java-driver/commit/b61d475ff9d8192342f5c7039b66e57169d740c0

The reason to change to always download a new server is to avoid using the old server by mistake when running tests locally in IDE

But maybe it might be better to change it back to the old behaviour and call
`FileTools.deleteRecursively( neo4jHome );` inside `private void stopOnExit()`

lassewesth:
@jakewins I would love it if we did not repeat the same mistake we have over in the product where we have docs stuff sitting in between product code.

Any examples we have that need to go into the manual can _at least_ sit in a separate module - if not in a separate repo - so that we do not conflate the two, and suffer drag.

driver/src/test/java/org/neo4j/driver/util/Neo4jRunner.java
jakewins: Oh, this was actually by accident - I just added it temporarily to speed up the test. It makes sense to have, I think, but it'd be nice to be able to turn it off as well, or to have it check if the downloaded file it has is the same as the file thats up remotely. Wonder if we can get S3 to give us 304-not-modified responses..

In any case, it's a separate thing, I'll undo this.

jakewins:
@lassewesth this is actually somewhat in reaction to that - this is trying to put examples directly in the documentation without automated cut-n-paste at build time. So these examples are all examples that go in the javadocs, like here: http://alpha.neohq.net/docs/java-driver/

This just adds test coverage for those examples, without moving them from their javadocs. Does that make sense, or am I misunderstanding you?

lassewesth:
@jakewins My protest is with having `driver/src/test/java/org/neo4j/driver/DriverDocIT.java`, I think it ought to be in `driver/src/somethingsomethingdocs/java/org/neo4j/driver/DriverDocIT.java` or maybe `driversomethingsomethingdocs/src/test/java/org/neo4j/driver/DriverDocIT.java`.

And then, of course, to ensure that when looking in `driver/src/test/java/**/*{Test, IT}.java` you see complete functional coverage, without including the DocITs.

jakewins:
@lassewesth to be clear; that test is not generating documentation like the 'DocIT' tests are in the neo code base. It's testing the examples in the javadocs. Given that, I think it makes sense to have it in the same module, since it's testing code that lives in the `main` directory. If we split it into a different module, it'd just be a module of tests that test code examples that live elsewhere.

Eg; this example snippet: https://github.com/neo4j/neo4j-java-driver/blob/doctest/driver/src/main/java/org/neo4j/driver/Driver.java#L37

Is represented by 'DocSnippet' and tested here: https://github.com/neo4j/neo4j-java-driver/blob/doctest/driver/src/test/java/org/neo4j/driver/DriverDocIT.java#L25

PR #55, opened 2015-09-25
Added isOpen

nigelsmall:


zhenlineo:
It will be great if you have some test to verify the behaviour change

driver/src/main/java/org/neo4j/driver/internal/StandardSession.java
```diff
@@ -70,6 +70,12 @@ public Result run( String statement )
     }
 
     @Override
+    public boolean isOpen()
+    {
+        return connection.isOpen();
```
jakewins: I'm not sure this is correct - wasn't the purpose of `isOpen` to determine if the session object had been closed? If so, this will not work - the connections remain open ideally for the duration of the users application, closing a session object simply returns the connection to the pool to be reused for another session object. 

Ergo; perhaps we can just have a flag in StandardSession that we set on close and check in this method? That'd simplify the code tracking if the underlying connection is open as well.

jakewins:
Would love a test that highlights the use case here!

driver/src/main/java/org/neo4j/driver/internal/pool/PooledConnection.java
```diff
@@ -34,10 +34,13 @@
     private Consumer<PooledConnection> release;
     private boolean unrecoverableErrorsOccurred = false;
 
+    boolean acquired;
```
zhenlineo: How this var is used in this method? I feel it is harmless without it too?

driver/src/main/java/org/neo4j/driver/Session.java
```diff
@@ -47,6 +47,13 @@
      */
     Transaction newTransaction();
 
+    /**
+     * Detect whether the session is currently open or not.
+     *
+     * @return true if the session is open, false if it has been {@link #close() closed}.
+     */
+    boolean isOpen();
```
boggle: I think all this should move up to StatementRunner, valid for transactions, too, right?

driver/src/main/java/org/neo4j/driver/internal/pool/PooledConnection.java
jakewins: +1, missed removing this.

driver/src/main/java/org/neo4j/driver/Session.java
jakewins: Fairynuff

zhenlineo: Interesting, when googling `fairynuff`, I got "Fairy Nuff is a fairy who studies medicine in Zanaris, found in her house north of the bank. She...", then I realized it is fair enough :)

boggle: That was funny

jakewins:
@zhenlineo @boggle updated. I guess this will be red now tho until the identity PR goes in

PR #56, opened 2015-09-28
Doc to return null when accessing non-existing values in Result

zhenlineo:
Minor doc improvements

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/SimpleRecord.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/SimpleRecord.java
@@ -57,7 +57,16 @@ public class SimpleRecord implements Record
     @Override
     public Value get( String fieldName )
     {
-        return fields[fieldLookup.get( fieldName )];
+        Integer fieldIndex = fieldLookup.get( fieldName );
+
+        if ( fieldIndex == null )
+        {
+            return null;
+        }
+        else
+        {
+            return fields[fieldIndex];
+        }
     }
 
     @Override
--- a/driver/src/main/java/org/neo4j/driver/internal/SimpleResult.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/SimpleResult.java
@@ -76,6 +76,12 @@ public class SimpleResult implements Result
     @Override
     public Value get( String fieldName )
     {
+        if( current == null )
+        {
+            throw new ClientException(
+                    "In order to access fields of a record in a result, " +
+                    "you must first call next() to point the result to the next record in the result stream." );
+        }
         return current.get( fieldName );
     }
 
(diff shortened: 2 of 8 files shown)


driver/src/main/java/org/neo4j/driver/internal/SimpleRecord.java
```diff
@@ -57,7 +58,16 @@ public Value get( int fieldIndex )
     @Override
     public Value get( String fieldName )
     {
-        return fields[fieldLookup.get( fieldName )];
+        Integer fieldIndex = fieldLookup.get( fieldName );
+
+        if ( fieldIndex == null )
+        {
+            throw new ValueNotFoundException( fieldName );
```
jakewins: How come we have a discrepancy between this get() method and the property() method on the Entity interface? It seems it'd be easier for the user if accessing a non-existent field were the same in both cases.

@boggle argued we should return nulls in these cases, since that is what cypher does. I'm inclined to agree - a null check is nicer than exceptions for control flows, and if the user code is not equipped to deal with it they will get an exception anyway. Eg: I argue we should just return nulls from all of these rather than throw an exception, and update the javadocs and add tests accordingly.

driver/src/main/java/org/neo4j/driver/internal/SimpleResult.java
```diff
@@ -76,7 +77,24 @@ public Value get( int fieldIndex )
...
     public Value get( String fieldName )
     {
-        return current.get( fieldName );
+        if( current == null )
+        {
+            throw new ClientException(
+                    "The current record is null. " +
+                    "It is either because the result contains no more record, or the record has not been set. " +
+                    "You could call `next()` to test if there are more records in this result " +
+                    "and/or to iterate over result records." );
```
jakewins: I actually think it'd be enough to explain the case of not having called `next` - this will only happen if the user did not call `next`, or did not check the result of `next` being true. If you've reached the end of the stream, the result will point to the last record (assuming there was at least one record).

Eg. maybe this can be altered to highlight that scenario more - `In order to access fields of a record in a result, you must first call next() to point the result to the next record in the result stream.`

driver/src/main/java/org/neo4j/driver/internal/SimpleRecord.java
```diff
@@ -57,7 +58,16 @@ public Value get( int fieldIndex )
     @Override
     public Value get( String fieldName )
```
jakewins: This same fix should apply to get( int ) as well, I think, so that we are consistent.

driver/src/main/java/org/neo4j/driver/internal/SimpleRecord.java
zhenlineo: The strategy to guide the changes made in this PR is to make it easier for a end user to find their error. So I feel a client exception at the Record interface is easier than a NPE for a user to find the error. But if the cypher behaviour is already defined to return null, we definitely have no reason to change that. 

I will update the pr to change to return null and add doc instead.

PR #57, opened 2015-09-29
Add methods in Value to return as a list of java generic types

zhenlineo:
such as List<Long>

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/Value.java
+++ b/driver/src/main/java/org/neo4j/driver/Value.java
@@ -18,6 +18,9 @@
  */
 package org.neo4j.driver;
 
+import java.util.List;
+import java.util.function.Function;
+
 /**
  * Represents a value from Neo4j.
  * <p>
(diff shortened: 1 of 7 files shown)


driver/src/main/java/org/neo4j/driver/Value.java
```diff
@@ -84,9 +86,32 @@
...
+    /**
+     * If the value represents a number, this method will return true if the number is not equals to 0.
+     * If the value represents a collection, this method will return true if the collection is not empty.
+     * If the value represents a string, this method will return true, if the string is not empty.
+     * @return the value as a Java boolean, if possible.
+     */
     boolean javaBoolean();
 
+    /** @return the value as a java List<String>, if possible */
+    List<String> javaStringList();
```
jakewins: I'm wondering if we could take a different approach here - this is a really sensible design, and easy for users to find, but it will not help in more complex cases, for instance if I want a `List<Map<String, List<Integer>>>` or something crazy like that. 

What if we had a method called `javaList` that took a converter object, like a `java.util.Function` or some such? Then we could have an API like:

    <T> List<T> javaList( java.util.Function<Value, T> map );

With default converter functions in the `Values` helper class;

    private static Function<Value, String> STRING = new Function<>() {
        public String apply( Value val ) { return val.javaString(); }
    }

And thus usage like

    List<String> stringList = myVal.javaList( STRING );

PR #74, opened 2015-10-26
Return an empty stats if no stat is included in summary

zhenlineo:


zhenlineo:
@boggle We need to improve DIP to include what might happen when empty result summary returned.

boggle:
I'm happy with this code in this PR though I'd like to understand _why_ no summary was included. I though that we should always get one back? Could you please investigate the current behavior of Cypher, i.e. does db.execute(query) always give us update statistics? If so, remoting should behave the same. Or is this an optimization where we choose to not send an empty (all zero) statistics from the server side?

zhenlineo:
@boggle 
We optimized it to only send stats when there are updates:
```
if ( delegate.getQueryStatistics().containsUpdates() )
{
...
``` 
https://github.com/neo4j/neo4j/blob/3.0/community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/internal/CypherAdapterStream.java#L73

In other word, if we got no updates, then nothing is sent and on the client we should give them 0s by default.

jakewins:
Looks good! Needs rebase, happy to merge this

PR #86, opened 2015-11-21
New results api [DONT MERGE YET]

boggle:
o Draft; needs more tests
o And talking through a couple of things


jexp:
How often will that package version be updated, only when the actual driver API changes or also when the bolt protocol changes? Or when a new version of the driver is released? 

If the latter this will be very annoying to users. If the former there will be a disconnect between bolt-versions on the wire and driver versions.

Where does the idea of changing the packages to be versioned come from? I have not seen that before anywhere, it will be a big pain for our users.

jexp:
A fairly big update, but still no implementation of streaming results / larger datasets?

jakewins:
@jexp I've got a branch that has the baseline in place for streaming results, and we've got a card tracking it :) Don't worry!

I tend to agree on `Result` - if we ever decide to introduce another way to get a hold of cursors over records, we can extract a common interface then, rather than being generic now?

boggle:
Two things, I introduced ResultCursor vs RecordCursor purely here to see if we could abstract that way. Happy to kill one of them.  Regarding naming, the DIP says `RecordCursor`, we could simplify that by just saying `Records`, I still like being explicit in the name about what is being returned.

boggle:
@jexp The idea here is to aim to NEVER have to do a version change but to be able to do so if we must in a way that allows a driver to support two APIs in the same release.

boggle:
Sorry for the rebase.. had to cherry pick atop merge commit since the first half of this was already merged.

boggle:
The plan is to merge this and then finalize the API; we have some more changes queued up, some of this stuff is untested but we'll make sure we cover things before we proceed.

driver/src/main/java/org/neo4j/driver/v1/RecordView.java
```diff
@@ -0,0 +1,61 @@
...
+     * @param index the index of the field
+     * @throws IndexOutOfBoundsException if an invalid index was given
+     * @return the field's value
+     */
+    Value value( int index );
+
+    /**
+     * @return an immutable copy of the currently viewed record
+     */
+    Record record();
```
lutovich: Return type should probably be `RecordView`

driver/src/main/java/org/neo4j/driver/v1/RecordView.java
```diff
@@ -0,0 +1,61 @@
...
+ * A field is a pair of a string key and a value.
+ *
+ * A record view must not contain multiple fields with the same key.
+ */
+public interface RecordView
+{
+    /**
+     * @return an ordered sequence of all field's keys
+     */
+    Iterable<String> keys();
```
lutovich: If the sequence is ordered than maybe return type should be `List`?

driver/src/main/java/org/neo4j/driver/v1/RecordView.java
```diff
@@ -0,0 +1,61 @@
...
+     *
+     * @param key the field's key
+     * @return the field's value or null if this record view does not contain a field with the given key
+     */
+    Value value( String key );
+
+    /**
+     * @return the number of fields accessible from this record view
+     */
+    int numberOfFields();
```
lutovich: `size` feels like smth more conventional

driver/src/main/java/org/neo4j/driver/v1/RecordView.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Retrieve the value of the field at the given index
+     *
+     * @param index the index of the field
+     * @throws IndexOutOfBoundsException if an invalid index was given
+     * @return the field's value
+     */
+    Value value( int index );
```
lutovich: Why one would use this method instead of `#value(String)`?

driver/src/main/java/org/neo4j/driver/v1/Result.java
```diff
@@ -18,29 +18,40 @@
...
+    int position();
+
+    /**
+     * Test if the cursor is positioned at the last stream record of if the stream is empty.
      *
-     * @return {@link ReusableResult}
+     * @return true if the cursor is at the last record or the stream is empty.
      */
-    ReusableResult retain();
+    boolean atEnd();
```
lutovich, line 62: Doesn't this method give us information already inferable from results returned by `#next()`, `#skip(int)`, `#first()`, `#last()` ?

driver/src/main/java/org/neo4j/driver/v1/Result.java
```diff
@@ -50,49 +61,47 @@
...
 
     /**
-     * Get an ordered sequence of the field names in this result.
+     * Move to the first record if possible and verify that it is the only record
      *
-     * @return field names
+     * @return true if the cursor was successfully placed at the single first and only record
      */
-    Iterable<String> fieldNames();
+    boolean single();
```
lutovich, line 92: Isn't this a combination of `#first()` and `#next()` calls? If so than imho better remove it to keep the interface as small as possible.

driver/src/main/java/org/neo4j/driver/v1/Result.java
```diff
@@ -103,4 +112,7 @@
      * @return a summary for the whole query
      */
     ResultSummary summarize();
+
+    @Override
+    void close();
```
lutovich: Would be nice to document that implementations might throw if `Result` is already closed.

driver/src/main/java/org/neo4j/driver/v1/Result.java
boggle, line 92: Almost, it also asserts that the first is the last. This turns out to be surprisingly practical in my experience, (not only) but also for writing tests.

driver/src/main/java/org/neo4j/driver/v1/Result.java
boggle, line 62: All other methods will move the cursor, this one will not.

driver/src/main/java/org/neo4j/driver/v1/RecordView.java
boggle: You think? Isn't iterable ordered implicitly? We could change that, number of keys will never be super large so this doesn't have to be lazy.

driver/src/main/java/org/neo4j/driver/v1/RecordView.java
boggle: We would prefer that but it's confusing in `Result.size()`, that would look like number of rows, not number of columns, latest suggestion was `fieldCount()`.

driver/src/main/java/org/neo4j/driver/v1/RecordView.java
boggle: No, `Record` is immutable, `RecordView` is not (Cursors are RecordViews).

driver/src/main/java/org/neo4j/driver/v1/RecordView.java
boggle: It's a micro-micro-performance concern for tight loops (lookup index first and then key by that when using an array representation).

More importantly perhaps it aligns with the Value interface. Don't have a strong opinion on this one.

driver/src/main/java/org/neo4j/driver/v1/Result.java
boggle: I believe that's documented on the super, I'll copy it down.

boggle:
Status: We will go over this once more, after that it hopefully is good to go.

driver/src/main/java/org/neo4j/driver/v1/RecordView.java
lutovich: I do not think ordering for `Iterable` is guaranteed. Javadoc for `List` says: " An ordered collection (also known as a sequence)".

driver/src/main/java/org/neo4j/driver/v1/RecordView.java
lutovich: There is a suspicious cyclic knowledge between `RecordView` and `Record`.
`Record` extends `RecordView` but at the same time `RecordView` returns `Record` from one of it's methods. That is why I've asked.

driver/src/main/java/org/neo4j/driver/v1/RecordView.java
lutovich: If this method would not exist than `#numberOfFields()` would not be needed as well, right?

driver/src/main/java/org/neo4j/driver/v1/Result.java
lutovich, line 62: Why there is no `#atStart()` method?
Could you please maybe describe a use-case for this method?

driver/src/main/java/org/neo4j/driver/v1/Record.java
```diff
@@ -19,33 +19,8 @@
...
-
-    /**
-     * Get an ordered sequence of the field names in this result.
-     *
-     * @return field names
-     */
-    Iterable<String> fieldNames();
-
+public interface Record extends RecordView {
+    // intentionally empty
```
lutovich: Would it always be empty or this is temporary state?
It feels strange that immutable container extends mutable container.
In my mind mutable container would extend immutable one and add mutating methods.

driver/src/main/java/org/neo4j/driver/v1/RecordView.java
```diff
@@ -0,0 +1,61 @@
...
+package org.neo4j.driver.v1;
+
+/**
+ * A record view provides access to an ordered, finite collection of zero or more fields.
+ *
+ * A field is a pair of a string key and a value.
+ *
+ * A record view must not contain multiple fields with the same key.
+ */
+public interface RecordView
```
lutovich: What methods allow users to mutate `RecordView`?

driver/src/main/java/org/neo4j/driver/v1/RecordView.java
boggle: This cyclic dependency is going away soon

driver/src/main/java/org/neo4j/driver/v1/RecordView.java
boggle: Yes, I mean technically `numberOFields` can be derived from `keys.size`; but let's wait with that discussion until the new changes have happened.

driver/src/main/java/org/neo4j/driver/v1/Result.java
boggle, line 62: `atStart` is covered by `position() == -1`; Do you think we should have it as a shortcut? We felt that name may be misleading (is it position() == -1 or == 0?). `atEnd` is the only condition not detectable via `position()` hence we added the extra method. It is used by `single`; without `atEnd()` we couldn't implement that reliably without navigating past the `single` record. Also it may be useful for a future `peek` use case that we discussed.

driver/src/main/java/org/neo4j/driver/v1/RecordView.java
boggle: This is only done by methods on implementations; conceptually the contract of record view is just that it allows you look at a record, which record that is may change between invocations of any of its methods but of course in a controlled fashion (e.g. by the methods on result)

driver/src/main/java/org/neo4j/driver/v1/Record.java
boggle: To me that is super strange because it breaks substitution. If the super type is immutable but the sub type is mutable then what contract holds if I have a value of the super type that references a value of the sub type? On the other hand, when starting from (potential) mutability there is a clear contract, and a sub type then can narrow this down to immutability.

driver/src/main/java/org/neo4j/driver/v1/Result.java
boggle, line 62: Also, `atEnd` allows to check if the result is empty without advancing the cursor if its not (think something that throws if empty but otherwise passes on to an otherwise unsuspecting method that wants to just iterator via `next`)

driver/src/main/java/org/neo4j/driver/v1/Record.java
lutovich: I do not think it breaks substitution and it is pretty much impossible to narrow interface down to immutability.

```
interface Int {
    int get();
}

interface MutableInt extends Int {
    void set(int value);
}
```

So if one has `Int` that it is definitely guaranteed that it is possible to call `#get()`.

driver/src/main/java/org/neo4j/driver/v1/RecordView.java
lutovich: "done by methods on implementations" sounds very strange because it means we'll expose actual implementations. So we'll make it harder for ourselves to change the implementation in future.
If `Record` is immutable than it should only have methods for reading and if `RecordView` is mutable than it should inherit reading methods from `Record` and add writing methods.

driver/src/main/java/org/neo4j/driver/v1/Result.java
lutovich, line 62: Ok, no strong opinion on `#atStart()`. It is just a bit of asymmetry in the API that caught my eye.

zhenlineo, line 62: How about `isLast()` to make it more looks like a simple boolean method and to differentiate it from the methods that move the cursor?

lutovich, line 62: Sounds good to me. `java.sql.ResultSet` has both `#isFirst()` http://docs.oracle.com/javase/7/docs/api/java/sql/ResultSet.html#isFirst() and `#isLast()` http://docs.oracle.com/javase/7/docs/api/java/sql/ResultSet.html#isLast()

driver/src/main/java/org/neo4j/driver/v1/Record.java
boggle: I think you look at this from a different end - this is not so much about mutation operations, this is about guarantees when reading.

Assume:

```
interface MapAccess {
  Object value( String key )
}

interface Record extends MapAccess
```

If I now have a `MapAccess m` and then

```
Object o1 = m.value( "a" );
Object o2 = m.value( "a" );
```

there is no guarantee for `o1.equals(o2)` (because mutable, by other methods outside of the interface implemented by inheritors).

However if I have a record `Record r` and

```
Object o1 = r.value( "a" );
Object o2 = r.value( "a" );
```

then it always holds `o1.equals(o2)`. 

In the end, this is not about what methods are made available, this is about what contract you get when reading. If the inheritance order is flipped, as in `MapAccess extends Record`, then if I had a `Record r = new MapAccess()` my record would not be guaranteed to be immutable (by contract) anymore.

boggle: I think it is really wrong to make mutable classes inherit from immutable interfaces if one cares about read semantics across multiple invocations of reading methods (scala gets that even more right by distinguishing between shared superclass, immutable, and mutable interfaces).

driver/src/main/java/org/neo4j/driver/v1/RecordView.java
boggle: Think it like this: `RecordView` is not mutable, it is an interface with certain read methods and no contract regarding isolation between repeated reads, `Record` extends that interface and provides the contract of repeated read isolation. `Result` also extends `RecordView` and adds mutating operations. Makes more sense this way?

boggle:
Ping; we think we nailed the interface now, some tests still missing and not all javadoc is up to date.

lutovich:
@boggle I have a more general question. Now we have `Result` that gives ability to iterate over `Record` that exposes `Value` that exposes the actual java value.
Why do we need `Value` class? `Record` could just have methods like `#getString(int index)`, `#getInt(int index)` etc.
With such design we would not need fairly complex `<T> List<T> values( Function<Value, T> mapFunction )` methods but instead smth a bit simpler directly on `Record` class `#getList(int index, Class<T> elementClass)`.

driver/src/main/java/org/neo4j/driver/v1/Result.java
```diff
@@ -50,49 +66,40 @@
...
+     * Retrieve and store the entire remaining result stream (including the current record).
+     * This can be used if you want to iterate over the stream multiple times or to store the
+     * whole result for later use.
      *
-     * @return a single record from the stream
-     * @throws ClientException if the stream is empty
+     * Calling this method exhausts the result cursor and moves it to the last record
      */
-    Record single();
+    List<Record> retain();
```
lutovich, line 103: Why not simply `#toList()` ?
_Update:_ or maybe even `#asList()` to be consistent with `Value`

driver/src/main/java/org/neo4j/driver/v1/RecordView.java
lutovich: Ok, I understand the difference now.
Does this mean that driver would need to execute same request multiple times to provide `RecordView` guarantee but only once and cache the result to provide `Record` guarantee?

lutovich:
Now both `Result` and `Record` extend `RecordAccess`. This feels strange. It basically removes the need for `Record` class completely and makes `Result` more like conventional `java.sql.ResultSet`. I'd vote for having `Result` that gives ability to iterate over `Record`s which directly provide access to java values. With such layout we'll separate navigation over the result set and view of the result values.

driver/src/main/java/org/neo4j/driver/v1/Result.java
boggle, line 103: because I felt it is confusing that asList in one case returns a list of values but a list of records in another. The interfaces of Value and Result are already close so let's try to be helpful and use `retain` which is more in line with the naming of all the other methods of `Result`.  There's also an actual semantic difference: `retain` consumes the result, `asList` does not consume a value.

boggle:
No there still is a difference:

- RecordAccess: Read methods without isolation guarantees
- Record: Actually immutable record (i.e. isolation guarantees)
- Result: Read and write methods

boggle:
Result is intentionally not an Iterable of Record; we want to avoid the object construction if possible.

lutovich:
I'm not saying `Result` should be an `Iterable<Record>` but now when you mentioned possible reuse of `Record` objects I feel like documentation about this is missing.

My questions are:
 1) `RecordAccess` has no isolation guarantees. Does this mean driver can execute same query multiple times for two `Result#record()` calls?
 2) Why do we need `Value` since all methods for getting java values could be on `Record`?

boggle:
Regarding (2), there are a couple of reasons:

* Not all methods could be on record; you have to first specify the key (column name). That's why Result implements ListAccess and MapAccess in the latest commit here, so you call 
`result.value("name").asString()`; `result.asString()` makes no sense.
* Without a `Value` type it becomes difficult to just store an `uninterpreted` Cypher value in client code; you have to immediately project back to Java.  That is not always desirable, a query tool might rather want to work with Cypher values, for example to pass them on as received in a follow-up query as a parameter.
* We have (arbitrarily nested values like lists, maps, and entities. We need a way to go deep into these values. The `Value` type provides that. 
* It is nice to make the concern "extraction of Java values from Cypher types" the concern of a single abstraction (`Value`) from the viewpoint of the single responsibility principle 
* We could (as a performance optimization) support things like `asString( columnName )` but I'd rather first see good numbers that indicate we need that. And I can think of other ways of providing an alternative means of accessing values that avoids the boxing.

boggle:
Regarding (1), I really meant the term isolation much simpler. Imagine two pieces of client cold (called A and B) hold onto the same reference `r`, unaware of each other.

o Assume `r` is a `RecordAccess`. If (A) reads `r` , the (B) gets to execute, then (A) cannot rely that executing the same read of `r` will yield the same result (because B could mutate it in between).
o Assume `r` is a `Record`. If (A) reads `r` , the (B) gets to execute, then (A) can rely that executing the same read of `r` will yield the same result (because a `Record` is immutable).

This distinction leads to the proposed hierarchy:

* `RecordAccess`: Shared super interface with no isolation as described above
* `Record`: Sub interface with isolation (immutability) guaranteed by contract
* `Result`: Sub interface that defines some actual mutation operations

I hope I explained it better now, if not, perhaps let's have a call on this, this is getting a rather lengthy discussion on github.

driver/src/main/java/org/neo4j/driver/v1/RecordAccess.java
```diff
@@ -0,0 +1,48 @@
...
+ * limitations under the License.
+ */
+package org.neo4j.driver.v1;
+
+import java.util.List;
+
+/**
+ * Access the fields of an underlying ordered map like data structure by key and index
+ */
+public interface RecordAccess extends ListAccess, MapAccess
```
jakewins: I don't like the `XXAccess` naming - that makes it sound like this is an interface for accessing `Records` (eg. something that has `Record get();` as a method.

Why not name it `XXView`?

driver/src/main/java/org/neo4j/driver/v1/Path.java
```diff
@@ -61,7 +61,7 @@
     Node end();
 
     /** @return the number of segments in this path, which will be the same as the number of relationships */
-    long length();
+    int length();
```
jakewins, line 65: 2B is a lot for sure - but we've hit this again and again in other places, why constrain the API to that?

driver/src/main/java/org/neo4j/driver/v1/FieldAccess.java
```diff
@@ -0,0 +1,56 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.neo4j.driver.v1;
+
+/**
+ * Access to the interior fields of an underlying collection like
+ * data structure
+ */
+public interface FieldAccess
```
jakewins: Lists and maps do not have "fields", and those two are the only ones using this interface. If this is a view of a collection, why not call it `CollectionView`?

driver/src/main/java/org/neo4j/driver/v1/Values.java
```diff
@@ -268,44 +275,64 @@ public static Value value( final Map<String,Object> val )
         return VALUE;
     }
 
-    public static Function<Value,String> valueToString()
+    public static Function<Value,String> valueAsString()
```
jakewins, line 279: Nitpicking now, and really entirely unrelated to this PR (and should not be resolved in this PR), I don't like the `valueAsXX` because it reads odd when this is used:

```
value.asList( valueAsString() );
```

I think it'd be nicer to have it be `ofString` or `asString` here:

```
value.asList( ofString() );
```

driver/src/main/java/org/neo4j/driver/v1/Values.java
```diff
@@ -27,6 +27,7 @@
 import java.util.Map;
 
 import org.neo4j.driver.v1.exceptions.ClientException;
+import org.neo4j.driver.v1.internal.util.Extract;
```
jakewins: This is a circular dependency from the API to the impl - which means we could not swap out the implementation of the API. We don't need to fix this here, but I'd like to resolve it after this PR goes in. I've got a branch with a build plugin that fails the build if we have dependencies from the API to impl.

driver/src/test/java/org/neo4j/driver/v1/integration/ResultStreamIT.java
```diff
@@ -83,17 +86,17 @@ public void shouldGiveHelpfulFailureMessageWhenCurrentRecordHasNotBeenSet() thro
         }
     }
 
-    @Test
+    @Ignore("revisit")
```
jakewins: Assuming these will be turned back on before PR is good to go?

driver/src/main/java/org/neo4j/driver/v1/RecordAccess.java
boggle: Several people objected to that naming; also we might introduce actual views later on in Cypher in which case this perhaps gets confusing.  Scala uses "Like" as the suffix for classes like that. I'm happy either way.

driver/src/main/java/org/neo4j/driver/v1/Path.java
boggle, line 65: It gets unnatural from a Java POV (Java arrays are int based for example). By choosing longs we force the mental overhead of taking care there's no overflow etc to users instead of us taking care of that.  We talked this over and think if we need/want long indexing we'd rather duplicate the methods and have additional `longLength()` and `value( long )` methods.  These two should be enough. wdyt?

driver/src/main/java/org/neo4j/driver/v1/FieldAccess.java
boggle: +1 on calling it `CollectionXYZ` though the idea is still that Field is the umbrella term for map entry and list element. We need to use something.

driver/src/test/java/org/neo4j/driver/v1/integration/ResultStreamIT.java
boggle: They should

driver/src/main/java/org/neo4j/driver/v1/Path.java
boggle, line 65: Oh wait I misread.. this is about Path not FieldAccess...hmm I guess same reasoning applies here.

driver/src/main/java/org/neo4j/driver/v1/Values.java
boggle: I don't see how we would do that in a convenient way for a class like `Values` that has static methods. This surely contains implementation code. For all non-utility classes I'm happy to try separate between interface and implementation as much as possible.

driver/src/main/java/org/neo4j/driver/v1/Path.java
jakewins, line 65: Hmm, true that the technical aspects of this do become tricky, since all the array types are `int` in Java. Yeah, fine, this will be easier for users - and if they are dealing with 2B+ long paths we'll likely need specialized path cursor types anyway.

driver/src/main/java/org/neo4j/driver/v1/Values.java
jakewins: I really think we should enforce strict separation here - the API should never explicitly depend on the implementation. If we take that stance, it follows that we'll need to change `Values` somehow.

One way to do that is to make the `Values` utility into an interface, and allowing users to get instances of it from `Driver` or `GraphDatabase`. 

Another is that we could move all the simple value classes into the API as well - SimpleNode, SimpleMap, SimpleBlah all become part of the API.

Yet another is to make `Values` be an implementation utility - it's there, but it is not part of the API. That seems less appealing though, since it'd kill the portability of pretty much any app using the API. 

Maybe there are other options as well, we should look at how the Java File System APIs do this, since they heavily use static utility functions - in any case, one of the biggest problems with the current Core API is that it is circularly dependent on its implementation. We should enforce in the build that that is not the case for the new API.

driver/src/main/java/org/neo4j/driver/v1/Record.java
```diff
@@ -19,33 +19,8 @@
 package org.neo4j.driver.v1;
 
 /**
- * A record is a collection of named fields, and is what makes up the individual items in a {@link
- * Result}
+ * A record is an immutable copy of an ordered map
  */
-public interface Record
+public interface Record extends RecordAccess
```
jakewins: IMO, if we want this, we should explicitly name it `ImmutableRecord`

driver/src/main/java/org/neo4j/driver/v1/RecordAccess.java
jakewins: Fair enough. I'd be +1 to `RecordLike` etc over `Access` in that case.

driver/src/main/java/org/neo4j/driver/v1/FieldAccess.java
jakewins: In my mind a `field` is specific to Records - it has key, a position and a value. I don't think it makes sense to say that a list contains 'fields' - it contains values, likewise, the entries in a map does not have positions.

boggle: How would you call the generic notion of "contained element"? Regarding ordering, note that RecordAccess overrides everything to return lists to make that more clear.

zhenlineo: How about sth like:

```
interface CollectionView
{
  int size();
  boolean isEmpty();
  Iterable<Value> values();
  <T> Iterable<T> values( Function<Value, T> mapFunction );
}
```

Then for any XXView that might introduced in Cypher, we call them `XXCypherView` in the future?

jakewins: Zhen: I like that API - but I agree with Stefan that `View` might end up
being confusing to use. I can't really think of anything better than `Like`
or `Access` as an alternative.

Stefan: Aren't they just `Values`? A collection of Values?

technige:
Please re-raise this against the 1.0 branch

boggle:
Closing an re-opening against new 1.0 branch.

PR #88, opened 2015-11-30
Added support for CypherTypes in driver

pontusmelke:
Added ability to do `CypherTypes.typeOf( value )` to figure out the
Cypher type of a given value.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/v1/CoarseCypherType.java
+++ b/driver/src/main/java/org/neo4j/driver/v1/CoarseCypherType.java
@@ -1,0 +1,37 @@
+/**
+ * Copyright (c) 2002-2015 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.neo4j.driver.v1;
+
+/**
+ * The coarse Cypher type of a {@link Value}
+ *
+ * A coarse Cypher type is the type of a value that can be inferred without inspecting its elements, e.g.
+ * the coarse type of a LIST OF INTEGER is LIST OF ANY?
+ */
+public interface CoarseCypherType extends CypherType
+{
+    /**
+     * Test if a value has this Cypher type
+     *
+     * @param value the value
+     * @return <tt>true</tt> if the value is a value of this Cypher type otherwise <tt>false</tt>
+     */
+    boolean isTypeOf( Value value );
+}
+
--- a/driver/src/main/java/org/neo4j/driver/v1/CypherType.java
+++ b/driver/src/main/java/org/neo4j/driver/v1/CypherType.java
@@ -1,0 +1,35 @@
+/**
+ * Copyright (c) 2002-2015 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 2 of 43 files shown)


driver/src/main/java/org/neo4j/driver/v1/CypherTypes.java
```diff
@@ -0,0 +1,130 @@
...
+     * @return the smallest coarse Cypher type of the value
+     */
+    public static CoarseCypherType typeOf( Value value )
+    {
+        if ( value == null )
+        {
+            return NULL;
+        }
+
+        TypeConstructor constructor = ((ValueAdapter) value).typeConstructor();
```
jakewins: This is a circular dependency from API to impl - it means we couldn't implement this API with an alternative implementation. 

Why not add `type()` to `Value` directly?

driver/src/main/java/org/neo4j/driver/v1/internal/types/TypeConstructor.java
```diff
@@ -0,0 +1,161 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.neo4j.driver.v1.internal.types;
+
+import org.neo4j.driver.v1.Value;
+import org.neo4j.driver.v1.internal.value.ValueAdapter;
+
+public enum TypeConstructor
```
jakewins: I don't understand this enum - could we add a javadoc to describe its purpose?

driver/src/main/java/org/neo4j/driver/v1/CypherType.java
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.neo4j.driver.v1;
+
+/**
+ * The Cypher type of a {@link Value}
+ */
+public interface CypherType
```
jakewins: Kind of uncomfortable encoding `Cypher` as part of the API - could we make this just be `Type` instead, or `Neo4jType`?

driver/src/main/java/org/neo4j/driver/v1/CypherType.java
```diff
@@ -0,0 +1,35 @@
...
+{
+    /**
+     * @return the name of the Cypher type (as defined by Cypher)
+     */
+    String name();
+
+    /**
+     * @return the erased Cypher type of this type
+     */
+    CoarseCypherType erasedType();
```
jakewins: I'd love this JavaDoc to have a short description of what this means, because I don't think that's obvious to people who are not used to type theory.

driver/src/main/java/org/neo4j/driver/v1/CoarseCypherType.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+package org.neo4j.driver.v1;
+
+/**
+ * The coarse Cypher type of a {@link Value}
+ *
+ * A coarse Cypher type is the type of a value that can be inferred without inspecting its elements, e.g.
+ * the coarse type of a LIST OF INTEGER is LIST OF ANY?
+ */
+public interface CoarseCypherType extends CypherType
```
jakewins: What is the use case for this vs. CypherType? The only reason we have to inspect the inner values of a list is because we don't encode the nested type signature in Bolt - wouldn't it be better to just always return `List<Any>` for now, and then specialize that once Bolt supports the full type signatures?

driver/src/main/java/org/neo4j/driver/v1/CypherTypes.java
boggle: We could guarn against that by using an InternalType interface

boggle: Regarding why: null is Java null and hence can't have a method. Also doing it this way keeps it completely orthogonal and replaceable with a different type system.

driver/src/main/java/org/neo4j/driver/v1/CypherType.java
boggle: 50/50 on this: JDK has a "Type" class.

driver/src/main/java/org/neo4j/driver/v1/CoarseCypherType.java
boggle: Coarse types are a subset of all cypher types. All types returned and exposed by this PR are coarse types and `CypherTypes.LIST` is already `LIST OF ANY`as you requested. Another detail is that `isTypeOf` might be expensive for arbitray list values hence we only allow this operation on coarse types for now.  

This approach allows us to add richer type introspection later, even on the client side (by walking the Value) or by having Cypher send the richer types for query data.  So this is about future proofing things.

driver/src/main/java/org/neo4j/driver/v1/CypherTypes.java
jakewins: How about representing `null` with a `Value`?

driver/src/main/java/org/neo4j/driver/v1/CoarseCypherType.java
jakewins: I'm not smart enough here :) Can you give me a code example? Once we expose a nested type signature, that should be trivial to add to `ListType`, or no?

    Type t = value.type();
    Type inner = t.innerType();

    // Or, if we don't want `innerType` on the root Type interface:

    Type t = ((ListType)t).innerType();

jakewins:
Closing this to open a new one with one amended commit since I don't have push rights to @pontusmelke branch - more details in that PR!

PR #93, opened 2015-12-03
fix NPE if 'null' as parameter value is used

sarmbruster:


jakewins:
Nice! One thing - do you mind closing this PR and re-raising it against the `1.0` branch? We're in the middle of swapping over to use the same branching scheme as the main product repo..

PR #97, opened 2015-12-14
New results api, take 2

boggle:


driver/src/main/java/org/neo4j/driver/v1/Entry.java
```diff
@@ -21,11 +21,12 @@
...
  * Immutable pair of a key and a value
  *
- * @see PropertyMapAccessor
+ * @see MapAccessor
+ * @see RecordAccessor
  * @param <V> the Java type of the contained value
  */
 @Immutable
-public interface Property<V>
+public interface Entry<V>
```
pontusmelke: Doesn't the name clash with `java.util.Map.Entry` make it cumbersome to use?

boggle: Maybe.. I usually don't import it and just use Map.Entry directly in my Java code; that was the thinking here but I realize now that is not a good thing to assume by default. What would be a good generic alternative? We could just have a type `Pair<K, V>`? Wdyt?

boggle: Wikipedia agrees with that idea: https://en.wikipedia.org/wiki/Associative_array

boggle:
Something is weird with the PR build, it looks like there is a hanging db server. Let's see if this clears up tomorrow.

driver/src/main/java/org/neo4j/driver/v1/Entry.java
pontusmelke: I almost always import `Entry` but I am not sure if I am representative either. `Pair`, is fine by me.

driver/src/test/java/org/neo4j/driver/v1/TypePlayTest.java
```diff
@@ -0,0 +1,24 @@
+package org.neo4j.driver.v1;
+
+import org.junit.Test;
+
+public class TypePlayTest
+{
```
zhenlineo: ? What's the purpose of this test?

driver/src/main/java/org/neo4j/driver/internal/util/Format.java
```diff
@@ -33,9 +31,9 @@ private Format()
         throw new UnsupportedOperationException();
     }
 
-    public static <V> String formatProperties( Function<V, String> printValue,
-                                               int propertyCount,
-                                               Iterable<Property<V>> properties )
+    public static <V> String formatEntries( Function<V, String> printValue,
```
zhenlineo: Do we still keep the idea of an `entry`?
Shall it be called `formatPairs`

driver/src/main/java/org/neo4j/driver/internal/util/Format.java
```diff
@@ -33,9 +31,9 @@ private Format()
...
+                                            int propertyCount,
+                                            Iterable<Pair<String, V>> Entries )
```
zhenlineo: `Entries` -> `entries` or `pairs`?

driver/src/main/java/org/neo4j/driver/internal/util/Format.java
boggle: Yep, that's an omission. Good catch.

pontusmelke:
licenses...

PR #99, opened 2015-12-17
Added Cucumber for driver compliance tests

RagnarW:
Now runs Cucumber TCK tests

Step file for Bolt Type System Feature included

### DIFF
--- a/driver/src/test/java/org/neo4j/driver/v1/tck/BoltTypeSteps.java
+++ b/driver/src/test/java/org/neo4j/driver/v1/tck/BoltTypeSteps.java
@@ -1,0 +1,269 @@
+/**
+ * Copyright (c) 2002-2015 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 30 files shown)


driver/src/test/java/org/neo4j/driver/v1/tck/BoltTypeSystemTestIT.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.io.IOException;
+
+import org.neo4j.driver.v1.util.TestNeo4jSession;
+
+/**
+ * The base class to run all cucumber tests
+ */
+@RunWith( DriverCucumberAdapter.class )
+@CucumberOptions( features = {"target/resources/features/BoltTypeSystem.feature", "target/resources/features/BoltChunkingAndDechunking.feature"} )
```
zhenlineo: The current name of the class indicates this is not the correct class to test Chunking/Dechunking, isn't it? Shall we consider either create another class `ChunkingDechunkingIT` or rename this class to some meaningful place for all cucumber tests? Not sure what is the best practice in this situation.

driver/src/test/java/org/neo4j/driver/v1/tck/BoltTypeSteps.java
```diff
@@ -0,0 +1,445 @@
...
+        stringRunner = null;
+        runners = new ArrayList<>();
+    }
+
+    @After
+    public void print( Scenario sc )
+    {
+        if ( sc.isFailed() )
+        {
+            System.out.println( "Failed: " + sc.getName() );
```
zhenlineo: This is a great method, can we print more info for the failed tests? e.g. the expected result is a `5` of bolt `Integer`. (It would really helpful if we could get all the info, e.g. even `value` and `type` for the bug directly). Meanwhile, I would also suggest to skip the printout for success methods.

jakewins:
Merged in latest changes into this and opened a new PR, this LGTM and is now green locally given we merged the chunking fix. 

https://github.com/neo4j/neo4j-java-driver/pull/99

PR #100, opened 2015-12-17
Update config API for TLS to match DIP.

jakewins:
This modifes the Config API to talk about encryption rather than TLS, as
well as renaming the on/off and the tls auth config to EncryptionLevel and
TrustStrategy, respectively.


zhenlineo:
How come the second commit related to this PR?
Besides, I would assume the certificates should not be included in committed files, right?

jakewins:
+1, this ended up really messy. Cleaner PR with just the TLS changes here: https://github.com/neo4j/neo4j-java-driver/pull/111

PR #101, opened 2015-12-18
keys should never return null

pontusmelke:
For cypher statements that has no results, `resultCurson.keys` should return
an empty list rather than `null`.


driver/src/test/java/org/neo4j/driver/internal/InternalSessionTest.java
```diff
@@ -78,4 +78,34 @@ public void shouldBeAbleToOpenTxAfterPreviousIsClosed() throws Throwable
         // Then we should've gotten a transaction object back
         assertNotNull( tx );
     }
+
+    @Test
+    public void shouldNotBeAbleToUseSessionWhileOngoingTransaction() throws Throwable
```
pontusmelke, line 83: These two tests has nothing to do with the fix, I just thought they were missing.

PR #106, opened 2016-01-04
Use id for equality

pontusmelke:
Both nodes and relationships should only use `id` in `equals` and `hashCode`.


lassewesth:
@pontusmelke is this documented somewhere, what equality means for _unified_ drivers?

pontusmelke:
@lassewesth it is undefined since custom equality is not supported by all languages. We want to start with the simplest possible equality representation in all the drivers.

PR #111, opened 2016-01-04
Update config API for TLS to match DIP.

jakewins:
This modifes the Config API to talk about encryption rather than TLS, as
well as renaming the on/off and the tls auth config to EncryptionLevel and
TrustStrategy, respectively.


jakewins:
This will be red until https://github.com/neo4j/neo4j/pull/6134 makes its way through the pipeline

driver/src/main/java/org/neo4j/driver/internal/connector/socket/TrustOnFirstUseTrustManager.java
```diff
@@ -156,11 +158,29 @@ public void checkServerTrusted( X509Certificate[] chain, String authType )
...
     }
 
+    /**
+     * Calculate the certificate fingerprint - simply the SHA-1 hash of the DER-encoded certificate.
+     */
+    public static String fingerprint( X509Certificate cert ) throws CertificateException
+    {
+        try
+        {
+            MessageDigest md = MessageDigest.getInstance( "SHA-1" );
```
zhenlineo: How about let's move to SHA-2 directly as SHA-1 could already be reverted nowadays?

pontusmelke:
needs rebase

zhenlineo:
https://github.com/neo4j/neo4j-java-driver/pull/126

PR #113, opened 2016-01-05
Handle applications not closing sessions better.

jakewins:
- Introduces basic finalizer to the InternalSession object, which closes
  the underlying connection if the session was not closed by the user,
  and prints a warning to stderr.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/InternalSession.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/InternalSession.java
@@ -21,6 +21,7 @@ package org.neo4j.driver.internal;
 import java.util.Map;
 
 import org.neo4j.driver.internal.spi.Connection;
+import org.neo4j.driver.internal.spi.Logger;
 import org.neo4j.driver.internal.summary.ResultBuilder;
 import org.neo4j.driver.internal.types.InternalTypeSystem;
 import org.neo4j.driver.v1.ResultCursor;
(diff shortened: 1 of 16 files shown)


driver/src/main/java/org/neo4j/driver/internal/InternalSession.java
```diff
@@ -121,6 +121,27 @@ public TypeSystem typeSystem()
...
+                // force the connction to be closed, rather than returned to the connection pool.
+                // This can be removed once we have a good way in Bolt to reset a session, then we won't need to close it like this.
+                connection.forceClose();
+            }
+            else
+            {
+                connection.close();
+            }
+        }
+    }
```
chrisvest, line 142: Finalizers can run concurrently with other methods on the object, and in a different thread. May want to make the `isOpen` field volatile to ensure the write in `close` is visible to the finalizer thread.

jakewins, line 142: Good call. Also - WDYT about using the logging framework here rather than stderr? I'm kind of scared of finalizers, I was worried it'd risk leading down a rabbit hole of broken object graphs to call "too far out". Do you think it'd be safe to keep a `final Log` field on the Session and use that to print the warning?

chrisvest, line 142: Yeah, having a `final Log` type field would be better. The fields won't be nulled out until after the finalizer has completed. If the `Log` type has a finalizer method, then I'm not sure if you can rely on any call order between the two, though. In Stormpot I just let the objects leak. Whatever the case, there is definitely (assuming a bug-free Java driver) a bug in user-code if you ever find the need to put out that log line.

PR #121, opened 2016-02-07
Allow calling first() and single() multiple times as long as not pass the first record

boggle:


pontusmelke:
needs rebase to go green

PR #125, opened 2016-02-14
Streaming cursor

technige:
This PR adds streaming to the Java Driver. This passes control of message retrieval to the ResultCursor class and, as such, alters slightly where expections may be thrown from. In addition, each ResultCursor instance now has a recordBuffer (instead of a static list of records) which is used in the case where the cursor needs to "soak up" records from the network due to a later cursor traversing its records first.

Note that the ConnectionPoolIT test is disabled for now. From what I can tell, the occasional failures seen with this test are connected to the new startup scripts which can return before the server port is available. Fixing this should probably be a separate issue and not something to hold up this PR.


driver/src/main/java/org/neo4j/driver/internal/InternalResultCursor.java
```diff
@@ -18,33 +18,153 @@
...
-        this.keys = keys;
-        this.iter = new PeekingIterator<>( body.iterator() );
-        this.summary = summary;
+        this.connection = connection;
+
+        Map<String, Value> unmodifiableParameters =
+                (parameters == null) || (parameters.isEmpty()) ? NO_PARAMETERS : unmodifiableMap( parameters );
+        final SummaryBuilder summaryBuilder = new SummaryBuilder( new Statement( statement, unmodifiableParameters ) );
+
+        this.runResponseCollector = new StreamCollector()
```
pontusmelke: this could maybe be extracted as a method

driver/src/main/java/org/neo4j/driver/internal/InternalResultCursor.java
```diff
@@ -18,33 +18,153 @@
...
+            @Override
+            public void done()
+            {
+                if ( keys == null )
+                {
+                    keys = new ArrayList<>();
+                }
+            }
+        };
+        this.pullAllResponseCollector = new StreamCollector()
```
pontusmelke: and this as well

driver/src/main/java/org/neo4j/driver/internal/InternalResultCursor.java
```diff
@@ -18,33 +18,178 @@
...
     {
-        this.keys = keys;
-        this.iter = new PeekingIterator<>( body.iterator() );
-        this.summary = summary;
+        return new StreamCollector()
+        {
+            @Override
+            public void keys( String[] names )
+            {
+                keys = new ArrayList<>( Arrays.asList( names ) );
```
jexp: no need for an ArrayList

driver/src/main/java/org/neo4j/driver/internal/InternalRecord.java
```diff
@@ -41,11 +42,19 @@
...
         this.values = values;
+
+        int numFields = keys.size();
+        Map<String, Integer> fieldLookup = new HashMap<>( numFields );
+        for ( int i = 0; i < numFields; i++ )
+        {
+            String name = keys.get(i);
+            fieldLookup.put( name, i );
+        }
+        this.keyIndexLookup = fieldLookup;
```
jexp: for the field-counts we're dealing with (10s), keys.indexOf() should be good enough

driver/src/main/java/org/neo4j/driver/internal/InternalResultCursor.java
```diff
@@ -77,6 +222,9 @@ public int index( String key )
 
     public List<String> keys()
     {
+        while (keys == null && !done) {
```
jexp, line 225: can this become an endless loop?

driver/src/main/java/org/neo4j/driver/internal/InternalResultCursor.java
```diff
@@ -113,16 +261,32 @@ public long position()
...
+        {
+            return false;
+        }
+        else if (done)
+        {
+            return true;
+        }
+        else
+        {
+            while ( recordBuffer.isEmpty() && !done )
```
jexp, line 274: busy wait?

driver/src/main/java/org/neo4j/driver/internal/InternalResultCursor.java
technige, line 225: Only if the server responds incorrectly and in that case and Exception will be thrown, breaking out of the loop.

nigelsmall: Special message for Jim.

PR #126, opened 2016-02-15
Update config API for TLS to match DIP

zhenlineo:


driver/src/main/java/org/neo4j/driver/v1/Config.java
```diff
@@ -241,32 +241,44 @@ public Config toConfig()
     }
 
     /**
-     * A configuration to configure TLS authentication
+     * Control the level of encryption to require
+     */
+    public enum EncryptionLevel
+    {
+        /** With this level, the driver will only connect to the server if it can do it without encryption. */
+        REJECTED,
```
boggle: This name feels off. @nigelsmall?

technige: Agreed. Not even sure we need two levels of settings (security on/off + security mode). We should probably be able to make do with something like SECURITY_NONE, SECURITY_TOFU and SECURITY_FULL.

cleishm: `FULL` is misleading also. I presume you mean certificate authority based? So perhaps `SECURITY_CERT`?

Which, I think, makes it the same as the TrustStrategy.

driver/src/main/java/org/neo4j/driver/internal/connector/socket/TrustOnFirstUseTrustManager.java
```diff
@@ -168,9 +188,17 @@ private File createKnownCertFileIfNotExists() throws IOException
             File parentDir = knownCerts.getParentFile();
             if( parentDir != null && !parentDir.exists() )
             {
-                parentDir.mkdirs();
+                if(!parentDir.mkdirs()) {
+                    throw new IOException( "Failed to create directories for the known hosts file in " + knownCerts.getAbsolutePath() + ". This is usually " +
+                                           "because you do not have write permissions to the directory. Try configuring the Neo4j driver to use a file " +
```
cleishm: It may not be a failure due to write permissions.

driver/src/test/java/org/neo4j/driver/internal/ConfigTest.java
```diff
@@ -26,14 +26,12 @@
...
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertThat;
-import static org.junit.Assert.assertTrue;
 
 public class ConfigTest
 {
-    private static final File DEFAULT_KNOWN_CERTS = new File( System.getProperty( "user.home" ),
-            "neo4j/neo4j_known_certs" );
+    private static final File DEFAULT_KNOWN_CERTS = new File( System.getProperty( "user.home" ), ".neo4j/neo4j_known_certs" );
```
cleishm: I think @jakewins confirmed that this should be changed to `.neo4j/known_hosts`?

driver/src/main/java/org/neo4j/driver/v1/Config.java
boggle: Why not just ENCRYPTED and PLAIN_TEXT? Isn't that what this is about? And the class could be something like "EncryptionPolicy"

cleishm: +1. BTW, libneo4j-client handles this with a single `INSECURE` flag, which is essentially a Boolean, since there is only two different "levels" anyway.

jakewins: The point of REQUIRED/REJECTED was that it allows adding OPTIONAL later, so if you're connecting to a server that may or may not require TLS, the driver will work either way. This is a nice option if you want to centralize control of this policy - you only have to configure the server to say `REQUIRE`, and as long as all the drivers are set to `OPTIONAL` they will follow suit.

Without that option it becomes hard, I think, to change the policy without full system downtime as well. This is how Oracle defines their encryption policy. (actually, they have a fourth, something like `DESIRED` which solves an ambiguity if both sides are configured as `OPTIONAL`, but I can't see a real use case for that, just default to using encryption if both sides are set to `OPTIONAL`)

Obviously, we don't have OPTIONAL on the driver side here, but the naming of the other two was intended to leave a slot for that in a later version.

driver/src/test/java/org/neo4j/driver/internal/ConfigTest.java
jakewins: +1, but I forgot to sync with team on that and didn't update the DIP. If nobody minds, I'm +1 to changing this to `known_hosts` as per above.

driver/src/main/java/org/neo4j/driver/v1/Config.java
jakewins: Also, note that this policy is orthogonal to how the driver determines _trust_. This policy is choosing if encryption is required, optional or rejected. How trust is established is a separate variable.

cleishm: Right. But it's probably a bit speculative to leave an option for OPTIONAL on the client (driver) side. Indeed, it's not even possible in the current protocol, since the client starts the conversation and has to choose one or the other with no knowledge or negotiation.

jakewins: It's certainly possible - the driver would try encrypted first, and degrade to non-encrypted. `OPTIONAL` was part of the design and was originally in the DIP, but we agreed to wait to include it due to time constraints. Since this is a useful operational feature, and the overhead of leaving space here (calling this `EncryptionPolicy.REQUIRED/REJECTED` vs `EncryptionPolicy.SECURE/INSECURE` as you suggest) is minimal, I think the alternative paints us into a corner for very little benefit while this leaves a nice door open to add this feature later.

cleishm: Ah, you mean try two independent connections? Yes, that'd work. I was referring to the inability to decline a TLS handshake, but of course that can be done by closing the connection and starting again.

PR #128, opened 2016-02-19
Draft Authentication API

boggle:
NOT READY FOR MERGE!

This is a draft of a new driver API for authentication support


driver/src/main/java/org/neo4j/driver/v1/Config.java
```diff
@@ -67,6 +72,7 @@ private Config( ConfigBuilder builder )
 
         this.encryptionLevel = builder.encruptionLevel;
         this.trustStrategy = builder.trustStrategy;
+        this.authToken = builder.authToken;
```
boggle: The reason we didn't add AuthToken to Config was that that makes having a default config kind-of pointless unless we have add a defaultConfigWithAuthToken factory method.

driver/src/main/java/org/neo4j/driver/v1/Config.java
```diff
@@ -113,6 +119,11 @@ public TrustStrategy trustStrategy()
         return trustStrategy;
     }
 
+    public AuthToken authToken()
+    {
+        return authToken;
+    }
+
```
boggle: Should anyone with access to Config be able to obtain the AuthToken?

jakewins: Well, if you're on the JVM and you have access to this part of the object graph I don't think there's any way to keep that from happening anyway, if it's stored somewhere on the heap you can get it. We could look into techniques to obfuscate this of course.. but the JVM does not provide good help for that.

IMO this part of the API is a bit wrong - the end-user cares about the config builder, not the Config itself. I'd love to find a way to make all the methods here non-public

driver/src/main/java/org/neo4j/driver/v1/Config.java
jakewins: But isn't that what we have - `Config.build().withAuthToken(..)`?

boggle: Isn't  connecting to a URL with AuthToken _the_ most common thing todo with a driver? Why does a user have to care about setting up a config to achieve that goal?

driver/src/main/java/org/neo4j/driver/v1/Config.java
boggle: Messing with the heap can be restricted with a security policy but so can the set of callable methods so I guess that's a fair point.

boggle:
Also since we only have two classes in it and they need to be fairly easy to find it feels like we should move everything from the `auth` package one level up again.

driver/src/main/java/org/neo4j/driver/v1/Config.java
jakewins: Actually, yeah you're right. I'll change it.

jakewins:
@boggle hmm, I'm concerned about the number of things in that namespace - but as per our discussion today and your point above, this thing is something _every_ driver user will use, meaning it makes sense to put it top-level. If we add more complex functionality later (eg. expand back towards the full object model of the underlying model you outlined) we can add that in a security package.

boggle:
On that note, I'd love to separate the driver API into a part shared between connector and driver and a driver specific part. Perhaps that would help a bit with this.

jakewins:
@boggle IMO we shouldn't do that unless we decide to make the Connector API layer public, then it'd make sense to have classes that are shared between them in a third component, agreed - although IMO I think I'd prefer a clean separation, so the Connector layer has it's own representation of what an AuthToken is, but maybe that is overly separatey.

boggle:
I was more thinking about types like Session and Value that might be shared. We need to sort out dependencies between driver, api, connector, and product.

jakewins:
I disagree the concrete classes should be shared - while they share the abstract concepts, a Session is exposed in a different way on the connection layer than on the end-user layer. Same for `Value` - the needs we have internally for an API for values is different from what we'd want to commit to in a public API, I think.

But - this is all just my subjective opinion, like with so many other things :p 

In any case, I think that's orthogonal to this PR, lets keep exploring the API layering tho.

PR #129, opened 2016-02-23
Test for and handle failures on commit

technige:
This fix allows the driver to gracefully handle failures when they occur at commit time rather than at execution time.


jakewins:
@nigelsmall bump

pontusmelke:
@nigelsmall This needs update. It fails on driver `INIT` so most probably missing auth

PR #136, opened 2016-03-09
1.0 iterate

jakewins:
Two large API changes - one swapping Cursor for Iterator (see commit message for details), and one reorganizing the final API after that change (see commit message for details).

Because this moves quite a few classes in order to simplify onboarding into the API, the commits are quite large. Suggest you read them each on their own, rather than this whole PR in one go.

As far as I know, this now contains all pending API changes for 1.0.


driver/src/main/java/org/neo4j/driver/internal/InternalResultStream.java
```diff
@@ -28,26 +28,25 @@
...
+import org.neo4j.driver.v1.util.Function;
+import org.neo4j.driver.v1.util.Functions;
+import org.neo4j.driver.v1.value.Value;
 
 import static java.lang.String.format;
 import static java.util.Collections.emptyList;
-import static org.neo4j.driver.v1.Records.recordAsIs;
 
-public class InternalResultCursor extends InternalRecordAccessor implements ResultCursor
+public class InternalResultStream implements ResultStream
```
technige: We should discuss naming. I think we all have different ideas here.

jakewins: Agree, @boggle noted the same thing. I just picked something :) @boggle suggested `RecordResult`. Ideas welcome! Maybe even just `Result`?

boggle:
Looks good.. noticing in passing that Values.asMap() misses a variant that takes a function for mapping over the values

jakewins:
@boggle fixed above. Imma merge this then, since it resolves two outstanding blocking issues.

PR #147, opened 2016-03-21
Handle failures after closed transaction

### LINKED ISSUE #146: NullPointerException at org.neo4j.driver.internal.InternalSession$2.run when an invalid Cypher query is executed
With 1.0.0-M05, this exception is thrown

```
Exception in thread "main" java.lang.NullPointerException
    at org.neo4j.driver.internal.InternalSession$2.run(InternalSession.java:140)
    at org.neo4j.driver.internal.pool.PooledConnection.onDelegateException(PooledConnection.java:197)
    at org.neo4j.driver.internal.pool.PooledConnection.receiveOne(PooledConnection.java:146)
    at org.neo4j.driver.internal.InternalStatementResult.tryFetching(InternalStatementResult.java:344)
    at org.neo4j.driver.internal.InternalStatementResult.hasNext(InternalStatementResult.java:190)
```

when you attempt to execute an invalid Cypher query using session.run. This only happens when a previous transaction has been carried out in the same session.

```
Session session = driver.session();

//Must have a transaction first
StatementResult result = tx.run("CREATE (n) RETURN n");
if (result.hasNext()) {
    result.next();
}
result.consume();
tx.success();
tx.close();

//Now the bad query
StatementResult badResult = session.run("CREAT (n) RETURN n");
if (badResult.hasNext()) {
    badResult.next();
}
```

Without the previous `tx`, `badResult` is executed and produces
`Exception in thread "main" org.neo4j.driver.v1.exceptions.ClientException: Invalid input ' ': expected 'e/E' (line 1, column 6 (offset: 5))` as expected.

If `badResult` was a valid query `badResult = session.run("CREATE (n) RETURN n");` then both execute fine.

pontusmelke:
We should not run `onError` of a transaction after the transaction
has been successfully closed.

Fixes #146

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/InternalSession.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/InternalSession.java
@@ -137,9 +137,13 @@ public class InternalSession implements Session
             @Override
             public void run()
             {
-                currentTransaction.markAsRolledBack();
-                currentTransaction = null;
-                connection.onError( null );
+                //must check if transaction has been closed
+                if (currentTransaction != null)
+                {
+                    currentTransaction.markAsRolledBack();
+                    currentTransaction = null;
+                    connection.onError( null );
+                }
             }
         });
         return currentTransaction;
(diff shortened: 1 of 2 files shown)


technige:
:+1:

PR #149, opened 2016-03-29
Buffer channel reads

pontusmelke:
Keeps a buffer instead of reading individual bytes from channel.

Benchmarks show that this makes reading 5 times faster when reading 1 million rows from the database.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/connector/socket/BufferingChunkedInput.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/connector/socket/BufferingChunkedInput.java
@@ -1,0 +1,430 @@
+/**
+ * Copyright (c) 2002-2016 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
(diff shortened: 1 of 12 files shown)


jexp:
Yay! Thanks @pontusmelke

jakewins:
This looks awesome, the code is quite readable, much better than what it was before as well. So win-win in more efficient buffer usage and more maintainable code :) 

I'm missing one thing - when we've changed things around this stuff before, it's almost always introduced tricky edge cases. We rooted a lot of those cases out on the server side by using this test: https://github.com/neo4j/neo4j/blob/3.1/community/bolt/src/test/java/org/neo4j/bolt/v1/transport/socket/FragmentedMessageDeliveryTest.java#L66

Maybe we could copy that test into the driver (and edit it as appropriate) as part of this commit, as way to show that there's no fragmented delivery case that we do not handle with this new state machine approach?

pontusmelke:
@jakewins Ported the test

pontusmelke:
Use https://github.com/neo4j/neo4j-java-driver/pull/150 instead

PR #152, opened 2016-04-05
Don't rely on unsupported configuration

pontusmelke:
Instead of relying on `unsupported.dbms.security.auth_store.location`  to
control where the auth store is kept, we now use `dbms.directories.data`.


zhenlineo:
It looks good to me and I think it could be merged as it is a test improvement to make tests more reliable.

PR #153, opened 2016-04-05
Clarify semantics of statement result methods

boggle:
As per discusssion w Nigel / same-named trello card

Also cleans up InternalStatementResult impl quite a bit which was using recursion in weird places.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/InternalStatementResult.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/InternalStatementResult.java
@@ -246,19 +246,22 @@ public class InternalStatementResult implements StatementResult
     public Record peek()
     {
         assertOpen();
-        Record nextRecord = recordBuffer.peek();
-        if ( nextRecord != null )
-        {
-            return nextRecord;
-        }
-        else if ( done )
-        {
-            return null;
-        }
-        else
+
+        while ( true )
         {
             tryFetching();
-            return peek();
+            Record nextRecord = recordBuffer.peek();
+            if ( nextRecord == null )
+            {
+                if ( done )
+                {
+                    throw new NoSuchRecordException( "Cannot peek past the last record" );
+                }
+            }
+            else
+            {
+                return nextRecord;
+            }
         }
     }
 
--- a/driver/src/main/java/org/neo4j/driver/internal/InternalStatementResult.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/InternalStatementResult.java
@@ -269,12 +269,19 @@ public class InternalStatementResult implements StatementResult
     @Override
     public ResultSummary consume()
     {
-        while ( !done )
+        if ( done )
         {
-            connection.receiveOne();
             recordBuffer.clear();
         }
-        recordBuffer.clear();
+        else
+        {
+            do
+            {
+                connection.receiveOne();
+                recordBuffer.clear();
+            }
+            while ( !done );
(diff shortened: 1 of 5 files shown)


driver/src/main/java/org/neo4j/driver/internal/InternalStatementResult.java
```diff
@@ -274,46 +248,34 @@ else if ( done )
     @Override
     public <T> List<T> list( Function<Record, T> mapFunction )
     {
-        if ( isEmpty() )
-        {
-            assertOpen();
-            return emptyList();
-        }
-        else if ( position == -1 && hasNext() )
+        if ( hasNext() )
```
pontusmelke, line 251: couldn't this be rewritten to a `while( hasNext() )...` instead of `if ( hasNext() ) ... do {...} while ( hasNext() )...`?

driver/src/main/java/org/neo4j/driver/internal/InternalStatementResult.java
```diff
@@ -274,46 +248,34 @@ else if ( done )
...
     {
-        if(!open)
-        {
-            return summary;
-        }
-
         while ( !done )
         {
             connection.receiveOne();
+            recordBuffer.clear();
```
pontusmelke, line 274: why do we need to clear the queue multiple times?

driver/src/main/java/org/neo4j/driver/internal/InternalStatementResult.java
boggle, line 251: This is so that for the empty case we can use emptyList() which returns a constant. For the non-empty case we then already know that hasNext() is true which is why that case uses a do while loop

driver/src/main/java/org/neo4j/driver/internal/InternalStatementResult.java
boggle, line 274: I think we have two issues here:
- We don't want to fill up the buffer during consumption as that would cause unnecessary memory pressure, hence we need to empty the buffer in the inner loop
- We also want to make sure we empty the buffer in any case - imagine consume being called while there are still records in the buffer but done is already true - hence it gets emptied one more time after the loop

Is there a better way? How about:

```
    if ( done )
    {
        recordBuffer.clear();
    }
    else 
      do {
        connection.receiveOne();
        recordBuffer.clear();
      } while ( !done );

    return summary;
```

Would you like that better?

pontusmelke, line 274: Why is it not enough to just clear the queue after the loop?

boggle, line 274: Because then we could build up a prohibitively large queue during consumption and run out of memory during looping.

technige, line 274: Agree with @boggle. If we're draining records off the network, we shouldn't use the buffer so as to avoid unnecessary memory allocation.

The alternative (which might be a bit cleaner) would be to replace `receiveOne` with something like `discardOne` for this purpose. This would read the message off the network but not put it anywhere, thereby removing the need to purge the buffer each time. But unless this is quick and easy to do, we should probably stay with what we have for now and consider that alternative as an optimisation.

zhenlineo:
What's the status of this PR? It is good to me. Shall we merge?

PR #157, opened 2016-04-15
fixed wrong usage of Short.MAX_VALUE

ikwattro:
When packing an element, if the size of the value is beween `2^15` and `(2^16)-1`, a header marker `0xda` will be used instead of `0xd9` because of the usage of `Short.MAX_VALUE` which is `32767` and not `65535`.

I reported the same issue for neo4j itself here : https://github.com/neo4j/neo4j/issues/6959


pontusmelke:
Thanks awesome! Could you please close this PR and reopen against 1.0, sorry for the inconvenience.

ikwattro:
Sure !

PR #159, opened 2016-04-18
Add a TrustManager that trusts all certificates for testing purposes

hectorlf:
InProcessServer changes the server certificate on each restart, so none of the current strategies fit (without cumbersome scripting work, that is).

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/connector/socket/SSLContextFactory.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/connector/socket/SSLContextFactory.java
@@ -71,6 +71,9 @@ class SSLContextFactory
         case TRUST_ON_FIRST_USE:
             trustManagers = new TrustManager[]{new TrustOnFirstUseTrustManager( host, port, authConfig.certFile(), logger )};
             break;
+        case TRUST_ALL:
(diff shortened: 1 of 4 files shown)


jakewins:
Hey @hectorlf - I'm super happy to see contributions like this where you've come across an issue and are able to take the time to help others not stumble on the same thing. But, and I hope this doesn't discourage you from further contributions, I'm not sure we will accept this, because we very much want to avoid having a public API option that disables trust but makes it look like the connection is still encrypted. The reason for that is that might encourage people to use that option in production, making them think they are safely using encryption when they are really not.

For the case of unit testing; is there a reason you can't turn encryption off to get the same results?

```
Driver driver = GraphDatabase.driver(.., Config.build().withEncryptionLevel(NONE).toConfig() );
```

hectorlf:
Oh, please, I can't be discouraged when your argument is absolutely correct and I perfectly understand the reasoning behind it :)

I went this route because I recalled bolt didn't work on unencrypted connections, but I've just tested your solution and works fine. So, I don't know, my bad!!

I'm closing this. Thank you!!!

jakewins:
:) Can I ask which resources, if any, you used to find out how to set up the Bolt driver with the InProcessServer thing? It'd be good to update any docs resources to include this fix

PR #167, opened 2016-04-22
Resolve issue where pooled connections could leak if exception occurred

jakewins:
during close.
- close in PooledConnection calls reset and then returns the session to
  the pool. However, returning the session was not in a finally clause,
  so reset throwing an exception would leak the connection, eventually
  draining the pool from available sessions.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/pool/PooledConnection.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/pool/PooledConnection.java
@@ -156,11 +156,14 @@ public class PooledConnection implements Connection
         {
             reset( StreamCollector.NO_OP );
             sync();
-            release.accept( this );
         }
         catch (Exception ex)
         {
             dispose();
+        }
+        finally
+        {
+            release.accept( this );
         }
     }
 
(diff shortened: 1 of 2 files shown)


driver/src/main/java/org/neo4j/driver/internal/pool/PooledConnection.java
```diff
@@ -156,12 +156,15 @@ public void close()
...
             sync();
-            release.accept( this );
         }
         catch (Exception ex)
         {
             dispose();
         }
+        finally
+        {
+            release.accept( this );
```
technige, line 166: From the code, it looks like `release.accept` can fail with at least an `IllegalStateException`. Now this doesn't have a `catch` protecting it, what happens if/when this is thrown?

jakewins, line 166: That'll bubble up to the application layer - but this is good, if the pool rejects the connection return, that's a fatal problem that should get reported to the user. IllegalStateException happens whenever the pool state is corrupted, so this is good.

technige, line 166: In that case, :+1:

technige:
LGTM

PR #170, opened 2016-04-25
Message boundary bug

pontusmelke:
When the last byte in the buffer is the next chunk size we should not fetch more data until that byte is also consumed. First commit fixes the actual issue whereas the second two handles other feedback.
- Do not delegate in the state machine since it might lead to stack overflow errors
- Better handling in `readBytes` for large messages

Note that this PR addresses the same issue https://github.com/neo4j/neo4j-java-driver/pull/169


driver/src/test/java/org/neo4j/driver/internal/connector/socket/BufferingChunkedInputFuzzTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+            assertThat( dst, equalTo( expected ) );
+        }
+    }
+
+    @Test
+    public void messageSizeFuzzTest() throws IOException
+    {
+        int maxSize = 1 << 16;
+        Random random = new Random(11);
```
thobe: This should probably not be a fixed seed.

PR #178, opened 2016-05-26
Avoid marking objects as both live and disposed

pontusmelke:
In the `ThreadCachingPool` whenever an object is picked from the
`ThreadLocal` we cannot add it to the `disposed` queue directly since the
object will already be present in the `live` queue (or will be on release). Doing
that means that the object will be present both in the  `live` and `disposed`
queue and also means that on reallocation we can multiple copies of the same
reference and can lead to that the `live` queue blows up in size.

### DIFF
--- a/driver/src/test/java/org/neo4j/driver/internal/pool/ThreadCachingPoolTest.java
+++ b/driver/src/test/java/org/neo4j/driver/internal/pool/ThreadCachingPoolTest.java
@@ -18,17 +18,24 @@
  */
 package org.neo4j.driver.internal.pool;
 
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
-
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.reflect.Field;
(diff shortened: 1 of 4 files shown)


driver/src/test/java/org/neo4j/driver/internal/pool/ThreadCachingPoolTest.java
```diff
@@ -256,10 +268,126 @@ public void shouldRecovedDisposedItemReallocationFailing() throws Throwable
...
+            }
+        } );
+        executor.awaitTermination( 2, TimeUnit.SECONDS );
+        return succeeded.get();
+    }
+
+
+    //This is terrible hack, but I really want to keep the queues private in
+    //ThreadCachingPool
+    private static MethodHandle queueGetter( String name )
```
jakewins, line 375: It is indeed :) This seems acceptable to me, since it's in a test - but it does smell, and it smells like a missing abstraction. If we find ourselves digging deeper into this hole, one thing we could explore is to make the queues in the pool into some sort of service, so the pool takes an "ItemStorageService" or something that exposes the domain methods the pool is using the queues for. Then the test can inject its own "counting item storage" or whatever.

Something something back in my day we never did hacks.

pontusmelke, line 375: Yeah I found this code, https://github.com/neo4j/neo4j/commits/3.1/community/bolt/src/main/java/org/neo4j/bolt/v1/packstream/utf8/SunMiscUTF8Encoder.java?author=jakewins, on teh internets and ran with it. Seemed legit.

jakewins:
I'll forward-merge to 1.1 as well

PR #186, opened 2016-06-09
Block when negotiating versions

pontusmelke:
When the negotiating happens we must do blocking read and writes


pontusmelke:
Built on top of https://github.com/neo4j/neo4j-java-driver/pull/185 so that must be merged first

PR #188, opened 2016-06-10
Change to a simpler session pool

pontusmelke:
This PR removes the fixed size session pool and replaces it with a simple queue.


driver/src/main/java/org/neo4j/driver/v1/Config.java
```diff
@@ -84,9 +87,19 @@ public Logging logging()
...
     {
-        return connectionPoolSize;
+        return maxIdleConnectionPoolSize;
+    }
+
+    /**
+     * Max number of idle connections per URL for this driver.
+     * @return the max number of connections
+     */
+    public int maxIdleConnectionPoolSize()
```
jakewins, line 100: Since we're deprecating the other `connectionPoolSize`, and the config option on the builder uses the word `session`, could we stealthily swap to use `session` here as well?

jakewins:
Nice. 

Later on, we should explore allowing passing in a pool interface through the config; so by default you get this and don't have to think about pooling; but if you want to plug something highly configurable like c3p0 or really fast like stormpot in, you can do that from the outside.

PR #191, opened 2016-06-21
Added EncryptionLevel.REQUIRED_NON_LOCAL

technige:


driver/src/main/java/org/neo4j/driver/v1/Driver.java
```diff
@@ -72,6 +73,11 @@
 public interface Driver extends AutoCloseable
 {
     /**
+     * Return a flag to indicate whether or not encryption is used for this driver.
+     */
+    boolean encrypted();
```
lutovich: I think this method should be named `#isEncrypted()` to signify that it is a configuration option check and not some soft of a fluent interface.

driver/src/main/java/org/neo4j/driver/internal/util/AddressUtil.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Return true if the host provided matches "localhost" or "127.x.x.x".
+     *
+     * @param host the host name to test
+     * @return true if localhost, false otherwise
+     */
+    public static boolean isLocalhost( String host )
+    {
+        return LOCALHOST.matcher( host ).matches();
```
lutovich: This is not a good way to check for localhost. There can be multiple network cards, interfaces and what not. Plus regex matches things like `127.99999.99.123123123123`. Class `InetAddress` has methods like `#isAnyLocalAddress()`, `#isLoopbackAddress()`. It seems to be useful to check if address is defined on any interface like this `NetworkInterface.getByInetAddress(address) != null`. I guess we can adopt smth similar to http://stackoverflow.com/a/2406819/5517896

technige: Multiple network cards don't apply here. We're talking about the address to which the client connects and whether that is equal to `localhost` or `127.x.x.x`.

I'm also not sure that `127.99999.99.123123123123` would actually be a valid IP address anyway. But I'll check out the utility methods you mention. We do need to make sure we catch `localhost` here as well as the raw IP addresses so I'll work out which one(s) do that.

PR #206, opened 2016-07-26
[DO NOT MERGE YET] Driver changes for discovery

technige:
- Adds `ClusterDriver` implementation
- Uses `StubServer` for some tests (experimental)

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/v1/GraphDatabase.java
+++ b/driver/src/main/java/org/neo4j/driver/v1/GraphDatabase.java
@@ -18,6 +18,7 @@
  */
 package org.neo4j.driver.v1;
 
+import org.neo4j.driver.internal.ClusterDriver;
 import org.neo4j.driver.internal.DirectDriver;
 import org.neo4j.driver.internal.ConnectionSettings;
 import org.neo4j.driver.internal.net.BoltServerAddress;
@@ -168,6 +169,8 @@ public class GraphDatabase
         {
         case "bolt":
             return new DirectDriver( address, connectionSettings, securityPlan, poolSettings, config.logging() );
+        case "bolt+discovery":
+            return new ClusterDriver( address, connectionSettings, securityPlan, poolSettings, config.logging() );
         default:
             throw new ClientException( format( "Unsupported URI scheme: %s", scheme ) );
         }
--- a/driver/src/test/java/org/neo4j/driver/v1/util/StubServer.java
+++ b/driver/src/test/java/org/neo4j/driver/v1/util/StubServer.java
@@ -1,0 +1,75 @@
+/**
+ * Copyright (c) 2002-2016 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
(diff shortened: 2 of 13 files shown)


driver/src/test/java/org/neo4j/driver/v1/util/StubServer.java
```diff
@@ -0,0 +1,75 @@
...
+import static java.lang.Thread.sleep;
+import static java.util.Arrays.asList;
+import static java.util.Collections.singletonList;
+
+public class StubServer
+{
+    // This may be thrown if the driver has not been closed properly
+    public static class ForceKilled extends Exception {}
+
+    private static final int DEFAULT_PORT = 7687;
```
zhenlineo, line 35: Might need to default to an different port so that we could both have a real server and a stub server running at the same time

driver/src/test/java/org/neo4j/driver/v1/util/StubServer.java
```diff
@@ -0,0 +1,75 @@
...
+    private Process process = null;
+
+    private StubServer( String script, int port ) throws IOException, InterruptedException
+    {
+        List<String> command = new ArrayList<>();
+        command.addAll( singletonList( "boltstub" ) );
+        command.addAll( asList( Integer.toString( port ), script ) );
+        ProcessBuilder server = new ProcessBuilder().inheritIO().command( command );
+        process = server.start();
+        sleep( 500 );  // might take a moment for the socket to start listening
```
zhenlineo, line 46: This might be a lot of waiting time if we would have the script as a very common tool. Wondering if we could also keep the script running in background?

driver/src/main/java/org/neo4j/driver/v1/GraphDatabase.java
```diff
@@ -168,6 +169,8 @@ public static Driver driver( URI uri, AuthToken authToken, Config config )
         {
         case "bolt":
             return new DirectDriver( address, connectionSettings, securityPlan, poolSettings, config.logging() );
+        case "bolt+discovery":
```
zhenlineo, line 172: Personal preference to avoid `+` or other strange symbols in a url.
e.g.
`https://github.com/neo4j/neo4j-java-driver/pull/206/files` 
is far more better than 
`http+ssl://6ithub.com?!+neo4jneo4j+java\\driver+pull@@206~files`.

technige, line 172: The plus sign is explicitly mentioned in [RFC3986 §3.1](https://tools.ietf.org/html/rfc3986#section-3.1) as a valid character within the scheme part of a URI. Often this is used as a "link" between two related scheme parts and read as "with" or "over" such as [`git+https`](https://pip.pypa.io/en/stable/user_guide/#requirements-files).

I think here it reads more clearly than trying to combine two terms into a single word.

driver/src/test/java/org/neo4j/driver/v1/util/StubServer.java
technige, line 35: While the default port for Bolt connections will still be 7687, I agree there would be a big benefit in informally reserving a range that we use for testing. I've personally been using ports 900x for this. I think quality tools typically use 1000x.

PR #209, opened 2016-08-02
Move map up for Clojure map which is both a map and iterable.

zhenlineo:
In java, all maps are not iterable as it is undefined what to iterate so it is safe to assume nothing will be changed after this change.


technige:
:+1:

tuddman:
@zhenlineo : thanks!  any reason(s) this was merged into `1.0` and not `1.1` branch?  will this change be picked up in next release?

zhenlineo:
@tuddman I treat this change as a bug in the 1.0 code. Getting the change to 1.0 is also faster to have the problem fixed too. 

1.1 would probably not production ready until GA, as 1.1 mainly focus on new features and might be broken here and there while we are testing on new features.

PR #210, opened 2016-08-02
Kill queries

zhenlineo:
Added `session#reset` to stop long running statements and reset the session to a clean state


driver/src/main/java/org/neo4j/driver/v1/Session.java
```diff
@@ -58,7 +58,10 @@
      */
     Transaction beginTransaction();
 
-
+    /**
+     * Stop running more statements in this session and rest the session to a clean state.
+     */
+    void kill();
```
technige: As per earlier conversation, this should be named `reset` and could be described as follows:

```
/**
 * Reset the current session. This sends an immediate RESET signal to the server which both interrupts
 * any statement that is currently executing and ignores any subsequently queued statements. Following
 * the reset, the current transaction will have been rolled back and any outstanding failures will
 * have been acknowledged.
 */
void reset();
```

PR #211, opened 2016-08-08
Make sessioin.reset thread-safe

zhenlineo:
- Sync on `enqueue` and `flush` to avoid sending `reset` and other messages at the same time. 
- Also changed response handler stream collector queue to be concurrent queue so that adding into the queue (sending) and removing from the queue (receiving) can be done safely concurrently.
- Keep `ack_fail` muted when `reset` is already send but not replied yet.
- Sync on `tx.markToClose` and `tx.run` to enforce no more statement in the current tx, if tx is already failed (reset).


zhenlineo:
Shall we forbidden any more interaction with the session after session.reset or we just give user the fresh session and allow them to run more statement/Tx?

Issue #172, opened 2016-05-09
Please restore default values to call to get a node's property


crichey:
With Neo3 and bolt the method for getting a property value from a node lost the ability to set a default value in case the property doesn’t exist in the node. That was a VERY useful feature. With the current implementation you end up taking code that effectively was (prior to Bolt) as simple and clear as:
`String myValue = node.get(“myValue”,””)`
to

```
String myValue = “"
if(node.containsKey(“myValue”)) { myValue = node.get(“myValue”).asString()}
```

Obviously its not a huge deal if you have to do it once, but when you have to do it frequently (as I am when migrating 2.3 code to bolt) then it is a bit of a pain and a source of potential errors.

technige:
Thanks for the feedback. That would indeed be a useful addition to the API.

technige:
We have added this to our TODO list for implementation.

zhenlineo:
I am closing this issue as the changes already included with the new release [1.1.0-M02](https://github.com/neo4j/neo4j-java-driver/wiki/1.1-changelog#110-m02)

PR #214, opened 2016-08-16
Adding new API for accessing records and mapped values with a default value

zhenlineo:
This change enables 
`record.get( "key", defalutValue )`
and
`node.get( "propKey", defalutPropValue )`
`rel.get( "propKey", defalutPropValue )`
where defalutValue could be the types that we've defined `asXX` method already


zhenlineo:
For https://github.com/neo4j/neo4j-java-driver/issues/172

PR #216, opened 2016-08-17
Added permission check to the known_host file

zhenlineo:
to give user more information about why we failed to read/write to the known_host file

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/connector/socket/TrustOnFirstUseTrustManager.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/connector/socket/TrustOnFirstUseTrustManager.java
@@ -30,9 +30,10 @@ import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 import javax.net.ssl.X509TrustManager;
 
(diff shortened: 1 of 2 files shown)


driver/src/main/java/org/neo4j/driver/internal/connector/socket/TrustOnFirstUseTrustManager.java
```diff
@@ -107,12 +110,38 @@ private void saveTrustedHost( String fingerprint ) throws IOException
...
+            ) );
+        }
+    }
+
+    private void assertKnownHostFileWritable() throws IOException
+    {
+        if( !knownHosts.canWrite() )
+        {
+            throw new IOException( format(
+                    "Failed to write certificates to file %s as you have no write permissions to it. " +
```
spacecowboy: Would appreciate if you could add a `\n` after the first sentence in these messages, just like the ones below for CertificateException.

zhenlineo: Alright

PR #220, opened 2016-09-05
Backport server version and timing functionality to 1.0

pontusmelke:
Make this functionality available in 1.0 as well.

NOTE: Do a null merge when forward merging


zhenlineo:
Uncertain about if we want this back to 1.0, otherwise +1

PR #231, opened 2016-09-20
Automatically close on session.reset()

pontusmelke:
In the event of a reset the transaction should be completely closed and the
session should be ready to create new transactions without explicitly closing
any transactions.


zhenlineo:
So the synchronized modifier is to enforce changes to `currentTx` in `session` are all sequential. I can hardly think of any simpler way to make it more thread-safe.

PR #235, opened 2016-09-28
Unignore all boltkit tests

pontusmelke:
Infrastructure has been updated to handle tests using boltkit so these tests
should no longer ignored


pontusmelke:
~~Apparently infrastructure is not ready yet, I'll leave this PR open as an indicator of when boltkit is ready to go.~~

We are good to go

PR #236, opened 2016-09-30
Better error message when running statements against incompatible server

zhenlineo:
The connection will also be closed when putting back to pool.


driver/src/main/java/org/neo4j/driver/internal/pool/PooledConnection.java
```diff
@@ -224,9 +225,16 @@ public void dispose()
...
     {
-        if ( !isClientOrTransientError( e ) || isProtocolViolationError( e ) )
+        if ( isUnrecoverableErrorsOccurred( e ) )
         {
             unrecoverableErrorsOccurred = true;
+            if( isClusterError( e ) )
+            {
+                e = new ClientException( "The driver and server are incompatible. " +
+                        "1.0 driver does not support running statements against a 3.1+ core edge cluster. " +
+                        "Please upgrade to 1.1 driver and use `bolt+cluster` scheme to run statements directly" +
```
pontusmelke: should be `bolt+routing`

pontusmelke (changes requested):

pontusmelke (approved):

PR #249, opened 2016-10-13
Fixed expect strust but get xx bug

zhenlineo:
that is caused by forgetting to flip a buffer back after connection error.
Close the socket channel directly if server already mark the end of the channel


lutovich:
@zhenlineo it would be really nice to have tests for flipping the buffer on connection error and logs for empty `best effort` catch blocks.

zhenlineo:
@lutovich Logging for this specific best effort channel.close error is not very helpful. As this is not the real error. The error that the driver cannot read/write more data from the channel is the real error that a user should worry about. Whether I close this socket connection correctly or the socket just crashed is not really important as next session, they will get another new connection.

zhenlineo:
@lutovich Be free to find me if you still disagree. I merge this now as it is an important bug fix.

PR #250, opened 2016-10-13
Proper handling of cluster failures for transactions

pontusmelke:
We weren't checking for the cluster failures when running on transactions.
All operations should check for leader-switches and connection failures and
handle them appropriately.


driver/src/main/java/org/neo4j/driver/internal/RoutingTransaction.java
```diff
@@ -0,0 +1,143 @@
...
+    public TypeSystem typeSystem()
+    {
+        return delegate.typeSystem();
+    }
+
+
+    @Override
+    public void success()
+    {
+
```
zhenlineo: ?? why this and the following ones could be left empty?

pontusmelke: sorry, missed a push...

PR #252, opened 2016-10-19
Close all connections on driver.close()

pontusmelke:
Connections/Sessions that were in flight, i.e. acquired from the pool, were not
closed when shutting down the driver, connections remained open until GC
happened. The expected behavior is that `driver.close()` closes down all
sessions it has created.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
@@ -139,7 +139,7 @@ public class NetworkSession implements Session
     @Override
     public boolean isOpen()
     {
-        return isOpen.get();
+        return isOpen.get() && connection.isOpen();
     }
 
     @Override
@@ -176,10 +176,6 @@ public class NetworkSession implements Session
         try
         {
             connection.sync();
-        }
-        catch ( Throwable t )
-        {
-            throw t;
         }
         finally
         {
@@ -314,7 +310,7 @@ public class NetworkSession implements Session
 
     private void ensureSessionIsOpen()
     {
-        if ( !isOpen() )
+        if ( !isOpen.get() )
         {
             throw new ClientException(
                     "No more interaction with this session is allowed " +
--- a/driver/src/main/java/org/neo4j/driver/internal/net/pooling/BlockingPooledConnectionQueue.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/net/pooling/BlockingPooledConnectionQueue.java
@@ -1,0 +1,121 @@
+/**
+ * Copyright (c) 2002-2016 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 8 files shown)


driver/src/main/java/org/neo4j/driver/internal/net/pooling/BlockingPooledConnectionQueue.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * Offer a connections back to the queue
+     *
+     * @param pooledConnection the connection to put back to the queue
+     * @return <code>true</code> if connections was accepted otherwise <code>false</code>
+     */
+    public boolean offer( PooledConnection pooledConnection )
+    {
+        boolean offer = queue.offer( pooledConnection );
```
pontusmelke, line 61: Is it necessary to synchronize here, i.e. making the move from `queue` to `acquiredConnections` atomic? It feels like you should but I am failing to come up with a scenario where it matters, so I opted for less synchronization.

zhenlineo: The order should be first remove from `acquiredConn`, then put back to queue to avid the following scenario happening:
T1: queue.offer(connA)
T2: acquire() -> get connA // connA = queue.poll, acquiredConn.Add(connA)
T1: acquiredConn.remove(connA)

NOTE: _Remove first, then add back_. This rule applies to `acquire` too

zhenlineo: Also, whenever you _add_ into the `queue`/`acquiredConn`, you probably want to test if the pool already terminated. if it is already terminated, then drop the conn directly.

driver/src/main/java/org/neo4j/driver/internal/net/pooling/BlockingPooledConnectionQueue.java
```diff
@@ -0,0 +1,121 @@
...
+    public boolean contains( PooledConnection pooledConnection )
+    {
+        return queue.contains( pooledConnection );
+    }
+
+    /**
+     * Terminates all connections, both those that are currently in the queue as well
+     * as those that have been acquired.
+     */
+    public void terminate()
```
zhenlineo, line 123: You probably want a atomic var here at the very beginning of the method, to flag it that we've started termination. If any connection created after this, close it right now. If any connection want to _add_ to the queue or map, drop them directly.

driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
```diff
@@ -139,7 +139,7 @@ public synchronized void reset()
     @Override
     public boolean isOpen()
     {
-        return isOpen.get();
+        return isOpen.get() && connection.isOpen();
```
zhenlineo, line 142: I am not sure if we should add connection.isOpen test here. Though it is handle, however this method before is just to test if `session.close` is ever called. If you want to change this to `if this session could run more queries` or something else, then make sure that other methods who calling this `isOpen` is not broken.

driver/src/main/java/org/neo4j/driver/internal/net/pooling/BlockingPooledConnectionQueue.java
zhenlineo: Ref: https://github.com/neo4j/neo4j-dotnet-driver/blob/1.1/Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
We used lock because we do not have a concurrent container before. We probably could have a lock free impl for java driver

zhenlineo (changes requested):

zhenlineo (approved):

PR #257, opened 2016-10-21
Remove ConnectionFailureException

technige:
The `ServiceUnavailableException` should be used in all places in place of `ConnectionFailureException` as this can apply to both `bolt` and `bolt+routing` schemes and is not named in terms of the connection to which the user does not have direct access.


pontusmelke:
`ConnectionFailureException` was introduced in 1.1 so there is no need to deprecate them. In 1.0 they are all `ClientException` but we introduced them in order to figure out when a server has gone missing so we can throw the appropriate exception (`SessionExpired` or `ServiceUnavailable`) to the client.

technige:
Oops. Hadn't spotted that. And it was right there in the docstring too....

PR #260, opened 2016-10-25
Fix trust on first use for cluster

zhenlineo:
The driver creates a serverId-fingerprint map and pass this map to be shared by all connections. Each connection creates its own sslContext and trustManager at the connection starts. All connections refer to the map when getting/adding a server fingerprint to the file


technige:
Closing as no longer necessary due to #268

PR #262, opened 2016-10-25
Add tabular formatting to StatementResult

jexp:
- this is a PoC for feedback
- to be usable in testing and debugging
- to be used in cypher-shell and other shells

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/ResultFormatter.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/ResultFormatter.java
@@ -1,0 +1,28 @@
+/**
+ * Copyright (c) 2002-2016 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.neo4j.driver.internal;
+
+import java.io.PrintWriter;
+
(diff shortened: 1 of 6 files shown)


driver/src/main/java/org/neo4j/driver/internal/ResultFormatter.java
```diff
@@ -0,0 +1,224 @@
...
+import org.neo4j.driver.v1.types.Path;
+import org.neo4j.driver.v1.types.Relationship;
+import org.neo4j.driver.v1.util.Function;
+
+import java.io.PrintWriter;
+import java.util.*;
+
+import static org.neo4j.driver.internal.types.InternalTypeSystem.TYPE_SYSTEM;
+
+public class ResultFormatter {
```
thobe: This class looks stateless (apart from `serializeFunction` which looks like it never changes and could thus be made `static final`), so it could be either a singleton or have all it's methods be static.

thobe (commented):
Looks like a pretty nice addition.

driver/src/main/java/org/neo4j/driver/internal/ResultFormatter.java
lassewesth: s/could/should to reveal intent?

jexp: good point @thobe I kept it non-static, b/c otherwise all methods have to be static and that adds a lot of clutter.

thobe: Then I'd propose making it a singleton.

technige:
Good addition but we should pause until implementations have been prepared for all drivers.

technige:
For reference, similar functionality is provided by the `CypherWriter` class in py2neo:

https://github.com/nigelsmall/py2neo/blob/v3/py2neo/database/cypher.py#L28

We can pilfer some of this code for the official Python driver. Assuming the author agrees. Which he does.

driver/src/main/java/org/neo4j/driver/internal/ResultFormatter.java
jexp: can do

jexp:
Addressed comments now it's a singleton but offers an interface method.
Fixed code-style.

akollegger:
Probably someone could describe why this is appropriate as a driver feature. Please do so before merging this.

technige:
Closing after decision made not to include this functionality in the drivers at this time.

PR #263, opened 2016-10-26
Uniform handling of ENCRYPTION_NON_LOCAL

pontusmelke:
Although the old check, which looked for a loopback address was in someway
_more correct_. We want a well-defined uniform across driver. So the driver
will only consider a connection to be local if it matches `localhost`,
`127.x.y.z`, or `::1` (and variants thereof)

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/net/BoltServerAddress.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/net/BoltServerAddress.java
@@ -35,7 +35,7 @@ public class BoltServerAddress
     public static final int DEFAULT_PORT = 7687;
     public static final BoltServerAddress LOCAL_DEFAULT = new BoltServerAddress( "localhost", DEFAULT_PORT );
     private static final Pattern LOCALHOST =
-            Pattern.compile( "^(localhost|127(\\.\\d+){3})$", Pattern.CASE_INSENSITIVE );
+            Pattern.compile( "^localhost$|^127(\\.\\d+){3}$|^(?:0*\\:)*?:?0*1$", Pattern.CASE_INSENSITIVE );
 
 
     public static BoltServerAddress from( URI uri )
@@ -61,7 +61,7 @@ public class BoltServerAddress
 
     public BoltServerAddress( String host )
     {
-        int colon = host.indexOf( ':' );
+        int colon = host.lastIndexOf( ':' );
         if ( colon >= 0 )
         {
             this.port = Integer.parseInt( host.substring( colon + 1 ) );
--- a/driver/src/main/java/org/neo4j/driver/internal/net/BoltServerAddress.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/net/BoltServerAddress.java
@@ -16,7 +16,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.neo4j.driver.internal.net;
 
 import java.net.InetAddress;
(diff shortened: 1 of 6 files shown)


driver/src/main/java/org/neo4j/driver/internal/net/BoltServerAddress.java
```diff
@@ -138,21 +141,12 @@ public int port()
...
-        {
-            // confirmed to work as desired with both "localhost" and "127.x.x.x"
-            return InetAddress.getByName( host ).isLoopbackAddress();
-        }
-        catch ( UnknownHostException e )
-        {
-            // if it's unknown, it's not local so we can safely return false
-            return false;
-        }
+        return LOCALHOST.matcher( host ).matches();
```
MishaDemianenko, line 150: Why do we use regexp to know that that address belongs to local instead of InetAddress methods?

thobe: Because this behaviour can be implemented uniformly across all different drivers on all different platforms, such that we have the same behaviour in all drivers.

driver/src/main/java/org/neo4j/driver/internal/net/BoltServerAddress.java
```diff
@@ -34,6 +34,9 @@
 {
     public static final int DEFAULT_PORT = 7687;
     public static final BoltServerAddress LOCAL_DEFAULT = new BoltServerAddress( "localhost", DEFAULT_PORT );
+    private static final Pattern LOCALHOST =
+            Pattern.compile( "^(localhost|127(\\.\\d+){3})$", Pattern.CASE_INSENSITIVE );
```
thobe: What about IPv6 addresses (`::1`)?

technige: Good catch. We have no tests currently testing IPv6. We should have.

pontusmelke: Yes I actually thought about that as well, I'll add some tests.

pontusmelke:
@thobe now supports `::1` and variants thereof as well. Added some tests for connecting to the database with ipv6 as well and fixed a bug in how we create `BoltServerAddress` from a `host:port` string that wasn't working for ipv6.

driver/src/main/java/org/neo4j/driver/internal/net/BoltServerAddress.java
```diff
@@ -34,6 +34,9 @@
 {
     public static final int DEFAULT_PORT = 7687;
     public static final BoltServerAddress LOCAL_DEFAULT = new BoltServerAddress( "localhost", DEFAULT_PORT );
+    private static final Pattern LOCALHOST =
+            Pattern.compile( "^localhost$|^127(?:\\.[0-9]+){0,2}\\.[0-9]+$|^(?:0*\\:)*?:?0*1$", Pattern.CASE_INSENSITIVE );
```
thobe: This regex does not look quite right to me. In IPv4, can you really skip the zeroes in the middle?
And do we care about this matching invalid IP addresses?

pontusmelke: I thought `127.0000.0.1` was supported but lets not think about that, I changed it back to a simpler pattern.

technige:
Closing as no longer necessary due to #268

PR #265, opened 2016-10-27
Driver.transact

technige:
Add the `Driver.transact` method to remove the boilerplate code required to perform retries after receiving a `SessionExpiredException`.


driver/src/main/java/org/neo4j/driver/internal/DriverContract.java
```diff
@@ -0,0 +1,22 @@
+package org.neo4j.driver.internal;
```
pontusmelke, line 20: missing license header

driver/src/main/java/org/neo4j/driver/v1/RetryLogic.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public class RetryLogic
+{
+    public static final RetryLogic TRY_UP_TO_3_TIMES_WITH_5_SECOND_PAUSE = new RetryLogic( 3, 5_000 );
+    public static final RetryLogic DEFAULT_RETRY_LOGIC = TRY_UP_TO_3_TIMES_WITH_5_SECOND_PAUSE;
+
+    private final int attempts;
+    private final int pause;
+
+    public RetryLogic( int attempts, int pause )
```
pontusmelke: Maybe add a third argument, `TimeUnit timeUnit`, and then do `this.pause = timeUnit.toMillis(pause)` in order to avoid confusion over what unit we are dealing with. If not, the documentation must make it super clear what unit `pause` is in.

driver/src/main/java/org/neo4j/driver/v1/RetryLogic.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Encapsulates details of retry logic for operations that can be retried after failure.
+ */
+public class RetryLogic
+{
+    public static final RetryLogic TRY_UP_TO_3_TIMES_WITH_5_SECOND_PAUSE = new RetryLogic( 3, 5_000 );
+    public static final RetryLogic DEFAULT_RETRY_LOGIC = TRY_UP_TO_3_TIMES_WITH_5_SECOND_PAUSE;
+
+    private final int attempts;
+    private final int pause;
```
pontusmelke: pause is in milliseconds, should probably be a `long`.

driver/src/main/java/org/neo4j/driver/internal/BaseDriver.java
```diff
@@ -43,4 +56,69 @@ public boolean isEncrypted()
...
+                        {
+                            tx.close();
+                        }
+                        catch ( Exception ex )
+                        {
+                            // ignore errors if we've already failed as
+                            // we already know this connection is problematic
+                        }
+                    }
+
```
pontusmelke: weird extra line

driver/src/main/java/org/neo4j/driver/v1/RetryLogic.java
pontusmelke: `long pause`

driver/src/main/java/org/neo4j/driver/v1/RetryLogic.java
```diff
@@ -0,0 +1,48 @@
...
+        this.attempts = attempts;
+        this.pause = pause;
+    }
+
+    public int attempts()
+    {
+        return attempts;
+    }
+
+    public int pause()
```
pontusmelke: `public long pause()`

pontusmelke (dismissed):
Nice, some minor comments and you need to do `mvn license:format`

driver/src/main/java/org/neo4j/driver/internal/DriverContract.java
technige: Should be there now

driver/src/main/java/org/neo4j/driver/internal/BaseDriver.java
```diff
@@ -43,4 +54,68 @@ public boolean isEncrypted()
...
+            try
+            {
+                Thread.sleep( logic.pauseMillis() );
+            }
+            catch ( InterruptedException e )
+            {
+                throw new NotCommittedException( format( "Interrupted after %d attempts", logic.attempts() - remaining ) );
+            }
+        }
+        throw new NotCommittedException( format( "Unable to commit transaction after %d attempts", logic.attempts() ) );
```
zhenlineo, line 107: Is it really not committed or we do not know if it is committed? Is it possible that it actually get committed but the connection crashed before we get a reply to the client?

driver/src/main/java/org/neo4j/driver/v1/Driver.java
```diff
@@ -88,6 +92,13 @@
 
     Session session(AccessMode mode);
 
+    <T> T transact( RetryLogic logic, AccessMode mode, Function<Transaction, T> work )
```
thobe, line 95: Needs javadoc.

driver/src/main/java/org/neo4j/driver/v1/Driver.java
```diff
@@ -88,6 +92,13 @@
...
+            throws NotCommittedException, ServiceUnavailableException;
+
+    <T> T read( Function<Transaction, T> work ) throws NotCommittedException, ServiceUnavailableException;
```
thobe, line 98: Needs javadoc.

driver/src/main/java/org/neo4j/driver/v1/Driver.java
```diff
@@ -88,6 +92,13 @@
...
+
+    <T> T write( Function<Transaction, T> work ) throws NotCommittedException, ServiceUnavailableException;
```
thobe, line 100: Needs javadoc.

driver/src/main/java/org/neo4j/driver/v1/Driver.java
```diff
@@ -88,6 +92,13 @@
 
     Session session(AccessMode mode);
```
thobe, line 93: Needs javadoc. (not part of this commit, but still true)

driver/src/main/java/org/neo4j/driver/v1/exceptions/NotCommittedException.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.neo4j.driver.v1.exceptions;
+
+/**
+ * An <em>NotCommittedException</em> indicates that a unit of work could not successfully be committed.
+ * @since 1.1
+ */
+public class NotCommittedException extends Neo4jException
```
thobe, line 25: We should document the fact that there are cases where it could have committed, but we just were unable to find out that it committed...

driver/src/main/java/org/neo4j/driver/internal/BaseDriver.java
```diff
@@ -43,4 +54,68 @@ public boolean isEncrypted()
...
+    }
```
thobe, line 108: I feel like the code in here is more complex than it needs to be, but this can be addressed later.

thobe (commented):
I find the documentation somewhat lacking.

driver/src/main/java/org/neo4j/driver/internal/BaseDriver.java
thobe: yes, that is possible.

driver/src/main/java/org/neo4j/driver/internal/BaseDriver.java
```diff
@@ -43,4 +54,68 @@ public boolean isEncrypted()
         return securityPlan.requiresEncryption();
     }
 
+    public <T> T transact( RetryLogic logic, AccessMode mode, Function<Transaction, T> work )
```
zhenlineo, line 57: It there any reason to put `RetryLogic` the first parameter? Otherwise I would suggest to put it the final parameter. As in C# this method would be more flexible if I give a default value

```
public <T> T transact(AccessMode mode, Function<Transaction, T> work, RetryLogic logic=null)
```

If a user do not supply a value when calling this method, then I could default to `contract.retryLogic`

driver/src/main/java/org/neo4j/driver/internal/BaseDriver.java
technige: Possibly. I went longhand for clarity of algorithm which makes for easier porting.

PR #268, opened 2016-11-01
Deprecate TOFU, add TRUST_ALL and remove E_NON_LOCAL

technige:
- Adds `TRUST_ALL_CERTIFICATES` and sets as default trust mode
- Deprecates `TRUST_ON_FIRST_USE`
- Explicitly blocks `TRUST_ON_FIRST_USE` in combination with `bolt+routing`
- Removes `ENCRYPTION_NON_LOCAL`

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/v1/GraphDatabase.java
+++ b/driver/src/main/java/org/neo4j/driver/v1/GraphDatabase.java
@@ -235,12 +235,12 @@ public class GraphDatabase
                 // intentional fallthrough
             // END OF DEPRECATED CASES //
 
-            case TRUST_ALL_CERTIFICATES:
-                return SecurityPlan.forAllCertificates();
             case TRUST_CUSTOM_CA_SIGNED_CERTIFICATES:
                 return SecurityPlan.forCustomCASignedCertificates( config.trustStrategy().certFile() );
             case TRUST_SYSTEM_CA_SIGNED_CERTIFICATES:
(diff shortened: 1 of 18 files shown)


driver/src/main/java/org/neo4j/driver/v1/GraphDatabase.java
```diff
@@ -215,25 +214,33 @@ private static SecurityPlan createSecurityPlan( BoltServerAddress address, Confi
...
                 logger.warn(
                         "Option `TRUST_SIGNED_CERTIFICATE` has been deprecated and will be removed in a future " +
-                        "version " +
-                        "of the driver. Please switch to use `TRUST_CUSTOM_CA_SIGNED_CERTIFICATES` instead." );
-                //intentional fallthrough
+                        "version of the driver. Please switch to use `TRUST_CUSTOM_CA_SIGNED_CERTIFICATES` instead." );
+                // intentional fallthrough
+            // END OF DEPRECATED CASES //
+
+            case TRUST_ALL_CERTIFICATES:
```
zhenlineo: Shall we default to `TRUST_ALL` for old `TRUST_SIGNED`? isn't it more reasonable to fall through to `TRUST_CUSTOM_CA`? The `log.warn` at line 233 suggests it would be `TRUST_CUSTOM_CA`

technige: Yes, you're completely right, I messed up the ordering there.

zhenlineo:
+1

PR #272, opened 2016-11-15
Move server info into result summary

zhenlineo:
Replaced session.server with resultSummary.server.version
Replaced session.address with resultSummary.server.address

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/net/SocketConnection.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/net/SocketConnection.java
@@ -32,11 +32,13 @@ import org.neo4j.driver.internal.messaging.RunMessage;
 import org.neo4j.driver.internal.security.SecurityPlan;
 import org.neo4j.driver.internal.spi.Collector;
 import org.neo4j.driver.internal.spi.Connection;
+import org.neo4j.driver.internal.summary.InternalServerInfo;
 import org.neo4j.driver.v1.Logger;
 import org.neo4j.driver.v1.Logging;
 import org.neo4j.driver.v1.Value;
 import org.neo4j.driver.v1.exceptions.ClientException;
 import org.neo4j.driver.v1.exceptions.Neo4jException;
+import org.neo4j.driver.v1.summary.ServerInfo;
 
 import static java.lang.String.format;
 import static org.neo4j.driver.internal.messaging.AckFailureMessage.ACK_FAILURE;
@@ -50,7 +52,7 @@ public class SocketConnection implements Connection
     private final SocketResponseHandler responseHandler;
     private AtomicBoolean isInterrupted = new AtomicBoolean( false );
     private AtomicBoolean isAckFailureMuted = new AtomicBoolean( false );
-    private final Collector.InitCollector initCollector = new Collector.InitCollector();
+    private InternalServerInfo serverInfo;
 
     private final SocketClient socket;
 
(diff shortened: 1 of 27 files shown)


driver/src/main/java/org/neo4j/driver/internal/net/SocketConnection.java
```diff
@@ -69,15 +72,34 @@ public SocketConnection( BoltServerAddress address, SecurityPlan securityPlan, L
...
+        this.socket.start();
+    }
+
+    // for mocked socket testing
+    SocketConnection( SocketClient socket, Logger logger )
+    {
+        this.socket = socket;
+        this.logger = logger;
+
+        if( logger.isDebugEnabled() )
```
lutovich, line 80: Maybe extract this `if-else` in a method like `createResponseHandler(Logger)`, otherwise it is duplicated in both constructors.

driver/src/main/java/org/neo4j/driver/internal/net/SocketConnection.java
```diff
@@ -69,15 +72,34 @@ public SocketConnection( BoltServerAddress address, SecurityPlan securityPlan, L
...
+        this.socket.start();
     }
 
     @Override
     public void init( String clientName, Map<String,Value> authToken )
     {
+        Collector.InitCollector initCollector = new Collector.InitCollector();
         queueMessage( new InitMessage( clientName, authToken ), initCollector );
         sync();
+        this.serverInfo = new InternalServerInfo( socket.address(), initCollector.server() );
```
lutovich: `InternalServerInfo` constructor's second arg is called `String version` but here we pass `initCollector.server()` for it. This is a bit confusing. Maybe `InitCollector#server()` can be renamed to `#serverVersion()`?

driver/src/main/java/org/neo4j/driver/internal/summary/InternalServerInfo.java
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+package org.neo4j.driver.internal.summary;
+
+import org.neo4j.driver.internal.net.BoltServerAddress;
+import org.neo4j.driver.v1.summary.ServerInfo;
+
+public class InternalServerInfo implements ServerInfo
+{
+    private BoltServerAddress address;
```
lutovich: Fields can be final

driver/src/main/java/org/neo4j/driver/v1/summary/ServerInfo.java
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Returns a string telling the address of the server the query was executed.
+     * @return The address of the server the query was executed.
+     */
+    String address();
+
+    /**
+     * Returns a string telling which version of the server the query was executed.
+     * @return The server version of <code>null</code> if not available.
```
lutovich, line 37: Is it possible for the `version` to not be available while `address` is available?

driver/src/test/java/org/neo4j/driver/v1/integration/BookmarkIT.java
```diff
@@ -46,8 +45,9 @@
     @Before
     public void assumeBookmarkSupport()
     {
-        System.out.println( session.server() );
-        assumeTrue( ServerVersion.version( session.server() ).greaterThanOrEqual( v3_1_0 ) );
+        String version = session.run( "RETURN 1" ).consume().server().version();
+        System.out.println( version );
```
lutovich: This println can probably be removed.

lutovich (commented):
@zhenlineo made couple small comments.

driver/src/main/java/org/neo4j/driver/v1/summary/ServerInfo.java
zhenlineo: It is possible. The version is not supported before server 3.1. I will add a line in the doc

PR #275, opened 2016-11-18
Added StatementResult#summary 

zhenlineo:
to obtain the summary in the end of a result stream by pulling all unconsumed records into memory


technige:
:+1:

PR #278, opened 2016-11-23
Docs examples clean-ups

petraselmer:


zhenlineo:
@petraselmer You got an interesting error in CI.

PR #282, opened 2016-11-25
Removed ping

lutovich:
Previously ping query (`RETURN 1`) was executed on each `Session#close()`. This was intended to verify the underlying connection and decide if it should be returned to the pool or not.

Running such ping is not quite necessary because we already execute `RESET` and `SYNC` before that. Combination of `RESET` and `SYNC` guarantees to send at least one message and thus validates the connection. Also query `RETURN 1` made database start transaction and possibly do planning, this work seem unnecessary.

This commit removes ping and adds tests to verify that we actually send messages to socket during pooled connection validation.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/net/SocketConnection.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/net/SocketConnection.java
@@ -50,8 +50,8 @@ public class SocketConnection implements Connection
 {
     private final Queue<Message> pendingMessages = new LinkedList<>();
     private final SocketResponseHandler responseHandler;
-    private AtomicBoolean isInterrupted = new AtomicBoolean( false );
-    private AtomicBoolean isAckFailureMuted = new AtomicBoolean( false );
+    private final AtomicBoolean isInterrupted = new AtomicBoolean( false );
+    private final AtomicBoolean isAckFailureMuted = new AtomicBoolean( false );
     private InternalServerInfo serverInfo;
 
     private final SocketClient socket;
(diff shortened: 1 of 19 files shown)


driver/src/main/java/org/neo4j/driver/internal/net/SocketConnection.java
```diff
@@ -66,10 +66,19 @@ public SocketConnection( BoltServerAddress address, SecurityPlan securityPlan, L
...
+     * <b>Note:</b> this constructor should be used <b>only</b> for testing.
+     *
+     * @param socket the socket to use for network interactions.
+     * @param serverInfo the info about server this connection points to.
+     * @param logger the logger.
+     */
+    public SocketConnection( SocketClient socket, InternalServerInfo serverInfo, Logger logger )
     {
         this.socket = socket;
+        this.serverInfo = serverInfo;
```
zhenlineo, line 81: What this is for? Is there any tests that require this parameter?

zhenlineo:
Besides the strange serverInfo without a clear use in tests, this change looks good

driver/src/main/java/org/neo4j/driver/internal/net/SocketConnection.java
lutovich: This parameter is needed for `PooledConnectionValidatorTest#sendsSingleResetMessageForValidConnection` test. Field `serverInfo` is usually initialized during `init` and this makes it easier to test just the validation part.

PR #283, opened 2016-11-29
Better handling of closed driver

lutovich:
It was previously possible to use closed driver instance, obtain sessions from it and run queries. This was seen to cause problems in multi-threaded scenarios when some worker threads try to use driver, which was concurrently closed by other worker.
 This commit makes driver track if it is closed or not and throw exceptions if on every interaction when closed. Thrown exception is `IllegalStateException`. Actuall `#close()` method can be called multiple times and is a no-op for already closed driver.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/BaseDriver.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/BaseDriver.java
@@ -16,19 +16,26 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.neo4j.driver.internal;
 
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
 import org.neo4j.driver.internal.security.SecurityPlan;
+import org.neo4j.driver.v1.AccessMode;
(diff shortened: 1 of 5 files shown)


lutovich:
PR introduces a `ReadWriteLock` in driver to disallow `#close()` and `#session()` run concurrently. To ensure that it does not affect performance this JMH benchmark was used:
```
@Benchmark
public Session session() {
    Session session = driver.session();
    session.close();
    return session;
}
```

Results with 1 thread:
```
Benchmark                  Mode  Cnt   Score   Error   Units
sessionOpenCloseCurrent   thrpt  150  21.926 ± 0.508  ops/ms
sessionOpenCloseWithLock  thrpt  150  21.750 ± 0.537  ops/ms
```

Results with 8 threads:
```
Benchmark                  Mode  Cnt   Score   Error   Units
sessionOpenCloseCurrent   thrpt  150  60.869 ± 3.324  ops/ms
sessionOpenCloseWithLock  thrpt  150  61.202 ± 3.614  ops/ms
```

So it looks like lock does not reduce performance, at least for the used configuration.

PR #285, opened 2016-11-30
Removed finalize from NetworkSession

lutovich:
It contained logic for closing opened underlying connection and logging info about the leak. Closing of resources in finalize is not reliable and overridden finalize can hurt when many sessions are garbage collected.

Commit also introduces a system property `org.neo4j.driver.logSessionLeaks` to enable logging of unclosed/leaked sessions and log stacktrace of where they were instantiated. This is meant to help with session leak investigations. So finalization overhead will be present only when problem is present and is being investigated.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/LeakLoggingNetworkSession.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/LeakLoggingNetworkSession.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) 2002-2016 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.neo4j.driver.internal;
+
+import org.neo4j.driver.internal.spi.Connection;
+import org.neo4j.driver.v1.Logger;
+
+import static java.lang.System.lineSeparator;
+
+class LeakLoggingNetworkSession extends NetworkSession
+{
+    private final Logger log;
+    private final String stackTrace;
+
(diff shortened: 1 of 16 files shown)


driver/src/main/java/org/neo4j/driver/internal/LeakLoggingNetworkSession.java
```diff
@@ -0,0 +1,65 @@
...
+    {
+        logLeakIfNeeded();
+        super.finalize();
+    }
+
+    private void logLeakIfNeeded()
+    {
+        if ( isOpen() )
+        {
+            log.error( "Neo4j Session object leaked, please ensure that your application" +
```
zhenlineo, line 49: error? or info? or warn?
These three are all printed out by default in command line logging.

driver/src/main/java/org/neo4j/driver/internal/LeakLoggingNetworkSessionFactory.java
```diff
@@ -0,0 +1,49 @@
...
+    private static final String LOGGER_NAME = "sessionLeak";
+
+    private final Logger logger;
+
+    public LeakLoggingNetworkSessionFactory( Logging logging )
+    {
+        this.logger = logging.getLog( LOGGER_NAME );
+    }
+
+    public static boolean leakLoggingEnabled()
```
zhenlineo: wondering if we should enable it by a system variable or by a config option?
Such as
```Driver driver = Graphdatabase.Driver(uri, auth, Config.build().withSessionLeakLoggingEnabled().toConfig())```

lutovich: My main motivation for using system property is that users would not need to recompile their apps for leak investigation. So it would be possible to just restart it with `-Dorg.neo4j.driver.logSessionLeaks=true`. Do you think this makes sense?

driver/src/main/java/org/neo4j/driver/internal/LeakLoggingNetworkSession.java
lutovich: Leak is basically an error, that is why I use `log.error` here.

driver/src/main/java/org/neo4j/driver/internal/LeakLoggingNetworkSessionFactory.java
zhenlineo: aha, that's good thinking. I was just thinking that they need to restart anyway, but yeah, if they do not need to change the code, that's even better

zhenlineo:
You need to update your file header

zhenlineo:
@lutovich Sorry, need to rebase. I am general +1 on this, while I am wondering if we want to add this special parameter as a system property or we should consider to add in Config and build a xml config reader

PR #286, opened 2016-12-01
lock free driver

zhenlineo:
This PR ensures that after a driver is closed, no new session could be started.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/BaseDriver.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/BaseDriver.java
@@ -18,7 +18,7 @@
  */
 package org.neo4j.driver.internal;
 
-import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.concurrent.atomic.AtomicBoolean;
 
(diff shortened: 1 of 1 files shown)


driver/src/main/java/org/neo4j/driver/internal/BaseDriver.java
```diff
@@ -67,33 +58,26 @@ public final Session session()
...
-        }
-        finally
-        {
-            closedLock.readLock().unlock();
+            // the driver is already closed and we either 1. obtain this session from the old session pool
+            // or 2. we obtain this session from a new session pool
+            // For 1. this closeResources will take no effect as everything is already closed.
+            // For 2. this closeResources will close the new connection pool just created to ensure no resource leak.
+            closeResources();
+            throw new IllegalStateException( "This driver instance has already been closed" );
```
lutovich: Maybe pull creation of this exception in a method so it is not duplicated here and in `assertOpen`

lutovich (approved):
@zhenlineo PR looks good. Could we maybe have some tests with mocks that check we close resources in `#session()` when driver is closed?

PR #291, opened 2016-12-08
Better handling of concurrent driver close

lutovich:


driver/src/main/java/org/neo4j/driver/internal/net/pooling/BlockingPooledConnectionQueue.java
```diff
@@ -64,10 +69,10 @@ public boolean offer( PooledConnection pooledConnection )
             pooledConnection.dispose();
         }
         if (isTerminating.get()) {
-            PooledConnection poll = queue.poll();
-            if (poll != null)
+            PooledConnection connection = queue.poll();
```
fickludd, line 72: Very sensible.

driver/src/main/java/org/neo4j/driver/v1/GraphDatabase.java
```diff
@@ -151,97 +122,9 @@ public static Driver driver( String uri, AuthToken authToken, Config config )
...
-            default:
-                throw new ClientException(
-                        "Unknown TLS authentication strategy: " + config.trustStrategy().strategy().name() );
-            }
-        }
-        else
-        {
-            return insecure();
-        }
+        return new DriverFactory().newInstance( uri, authToken, config.routingSettings(), config );
```
fickludd, line 128: I don't know, how about a static factory instance or a static newInstance method? Feels strange to create a stateless Factory object every time, but maybe you have a reason...

driver/src/test/java/org/neo4j/driver/internal/net/SocketConnectorTest.java
```diff
@@ -0,0 +1,163 @@
...
+
+        try
+        {
+            connector.connect( LOCAL_DEFAULT );
+            fail( "Exception expected" );
+        }
+        catch ( Exception e )
+        {
+            assertThat( e, instanceOf( ClientException.class ) );
+        }
```
fickludd, line 92: One can write a quite useful assert helper for checking exceptions, see `org.neo4j.test.assertion.Assert.assertException`. If you're doing this often.

zhenlineo: I am generally against to use `exception.expect` as it is too hard to tell which line the exception is really thrown. When using it, from the line the exception is declared, any line afterwards could be the place where the exception is thrown.

driver/src/main/java/org/neo4j/driver/v1/GraphDatabase.java
zhenlineo: Good point. Will get a new PR to fix this if possible

lutovich: `DriverFactory` class has protected methods so classes that extend it can "inject" different behaviour. Thus it is not possible to make `#newInstance()` static. Allocation of `DriverFactory` might also be eliminated by the JVM if `GraphDatabase#driver()` is called often.

driver/src/test/java/org/neo4j/driver/internal/net/SocketConnectorTest.java
lutovich: Unfortunately we do not have dependencies on artifacts from `neo4j/neo4j`.

fickludd: Yeah, I was thinking to simply copy paste that one assert method into some test-util class if you find it useful. Just FYI though, do as you like =).

PR #292, opened 2016-12-08
Better internal error handling

zhenlineo:
Introduced a few internal errors for tracking errors internally as checked errors.
Before the internal errors go outside (Driver, Session, Transaction, Result), the errors will be casted to a public error.

Each method is marked with the specific errors that it is going to throw for now. This could be put into one general error once we agree on the internal/external error mappings.

Just made it compile, still some work need to do to fix many tests

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/cluster/LoadBalancer.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/cluster/LoadBalancer.java
@@ -21,12 +21,17 @@ package org.neo4j.driver.internal.cluster;
 import java.util.HashSet;
 
 import org.neo4j.driver.internal.RoutingErrorHandler;
+import org.neo4j.driver.internal.exceptions.BoltProtocolException;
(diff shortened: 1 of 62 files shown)


driver/src/main/java/org/neo4j/driver/internal/cluster/LoadBalancer.java
```diff
@@ -145,27 +156,28 @@ private synchronized void ensureRouting() throws ServiceUnavailableException
                 writers.update( cluster.writers(), removed );
                 routers.update( cluster.routers(), removed );
                 // purge connections to removed addresses
+                // TODO an error should be caught here to ensure all sets are closed even if there is some error
                 for ( BoltServerAddress address : removed )
```
zhenlineo: This TODO needs to be addressed

lutovich:
We've decided to not convert all exceptions to checked exceptions internally. Closing this PR for now.

PR #294, opened 2016-12-13
Make socket connection timeout configurable

lutovich:
Previously socket connect did not have a timeout and was bounded only by the timeout configured on the OS level. Such timeout could be rather long and result in repeated stalls trying to refresh routing table against a failed router.


zhenlineo:
lower the default timeout to 5s and add the default value into javadoc

PR #297, opened 2016-12-27
Configurable test-on-borrow for pooled connections

lutovich:
Connections are pooled for both direct and routing driver. Each acquired session contains a socket connection which was acquired from the pool. It is possible for an idle pooled connection to become invalid while just resting in the pool. This could happen when connection is terminated after some timeout by a load balancer or some other network facility.

This PR introduces a configurable connection liveness check timeout. So freshly acquired connection, that has been idle in the pool for more than configurable timeout, will be validated by sending a RESET message. If it appears to be broken acquisition will retry until a valid connection is found.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/net/pooling/SocketConnectionPool.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/net/pooling/SocketConnectionPool.java
@@ -19,11 +19,13 @@
 package org.neo4j.driver.internal.net.pooling;
 
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.neo4j.driver.internal.net.BoltServerAddress;
 import org.neo4j.driver.internal.spi.Connection;
 import org.neo4j.driver.internal.spi.ConnectionPool;
+import org.neo4j.driver.internal.spi.ConnectionValidator;
 import org.neo4j.driver.internal.spi.Connector;
 import org.neo4j.driver.internal.util.Clock;
 import org.neo4j.driver.internal.util.Supplier;
@@ -46,13 +48,13 @@ public class SocketConnectionPool implements ConnectionPool
     /**
      * Pools, organized by server address.
      */
-    private final ConcurrentHashMap<BoltServerAddress,BlockingPooledConnectionQueue> pools =
-            new ConcurrentHashMap<>();
+    private final ConcurrentMap<BoltServerAddress,BlockingPooledConnectionQueue> pools = new ConcurrentHashMap<>();
 
     private final AtomicBoolean closed = new AtomicBoolean();
 
     private final PoolSettings poolSettings;
     private final Connector connector;
+    private final ConnectionValidator<PooledConnection> connectionValidator;
     private final Clock clock;
     private final Logging logging;
 
@@ -60,6 +62,7 @@ public class SocketConnectionPool implements ConnectionPool
     {
         this.poolSettings = poolSettings;
         this.connector = connector;
+        this.connectionValidator = new PooledConnectionValidator( this );
         this.clock = clock;
         this.logging = logging;
     }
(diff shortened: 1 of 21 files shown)


driver/src/main/java/org/neo4j/driver/v1/Config.java
```diff
@@ -116,17 +123,14 @@ public int maxIdleConnectionPoolSize()
...
-     * used again, to ensure they are still live.
+     * Pooled sessions that have been idle in the pool for longer than this timeout
+     * will be tested before they are used again, to ensure they are still live.
      *
      * @return idle time in milliseconds
-     * @deprecated pooled sessions are automatically checked for validity before being returned to the pool. This
-     * method will always return <code>-1</code> and will be possibly removed in future.
      */
-    @Deprecated
     public long idleTimeBeforeConnectionTest()
```
zhenlineo, line 131: There is no need to keep this old name. Pls pick a good name that better describes this config instead.

driver/src/main/java/org/neo4j/driver/v1/Config.java
```diff
@@ -256,30 +261,53 @@ public ConfigBuilder withMaxIdleSessions( int size )
...
-         * @param timeout minimum idle time in milliseconds
+         * @param value the minimum idle time in milliseconds
+         * @param unit the unit in which the duration is given
          * @return this builder
-         * @deprecated pooled sessions are automatically checked for validity before being returned to the pool.
-         * This setting will be ignored and possibly removed in future.
          */
-        @Deprecated
-        public ConfigBuilder withSessionLivenessCheckTimeout( long timeout )
+        public ConfigBuilder withSessionLivenessCheckTimeout( long value, TimeUnit unit )
```
zhenlineo: Same here, the name should be consistent with the field/method name in `Config`

driver/src/main/java/org/neo4j/driver/internal/net/pooling/SocketConnectionPool.java
```diff
@@ -106,34 +124,57 @@ private BlockingPooledConnectionQueue pool( BoltServerAddress address )
...
+    private PooledConnection acquireConnection( final BoltServerAddress address,
+            final BlockingPooledConnectionQueue connectionQueue )
     {
-        BlockingPooledConnectionQueue connections = pools.remove( address );
-        if ( connections != null )
+        Supplier<PooledConnection> supplier = newPooledConnectionSupplier( address, connectionQueue );
+
+        int acquisitionAttempt = 1;
+        PooledConnection connection = connectionQueue.acquire( supplier );
+        while ( !canBeAcquired( connection, acquisitionAttempt ) )
```
zhenlineo: Suggesting to test queue.isEmpty() instead. And the while loop would be changed to something like
```
boolean done = false;
while(!done)
{
  if(connectionQueue.isEmpty())
  {
    done = true; // let `queue.acquire` for queue size + 1 times 
  }
  PooledConnection conn = connectionQueue.acquire(supplier);
  if(isHealthyConnection(conn))
  {
    return conn;
  }
}
// `isHealthyConnection` will not throw exception, so this is a final protection of NPE
throw new ClientException("Failed to establish connection with the server or some other better error message"); 


private boolean isHealthyConneciton(PooledConnection conn)
{
  if ( poolSettings.idleTimeBeforeConnectionTestConfigured() && hasBeenIdleForTooLong(conn) )
  {
    return connectionValidator.isConnected( connection );
  }
  return true;
}
```

driver/src/main/java/org/neo4j/driver/internal/net/pooling/SocketConnectionPool.java
```diff
@@ -106,34 +124,57 @@ private BlockingPooledConnectionQueue pool( BoltServerAddress address )
...
+                PooledConnectionReleaseConsumer releaseConsumer =
+                        new PooledConnectionReleaseConsumer( connectionQueue, connectionValidator );
+                return new PooledConnection( connector.connect( address ), releaseConsumer, clock );
+            }
+        };
     }
 
-    @Override
-    public void close()
+    private boolean canBeAcquired( PooledConnection connection, int acquisitionAttempt )
```
zhenlineo: This is not needed if we have `isHealthyConneciton` (see the previous comment)

driver/src/main/java/org/neo4j/driver/internal/net/pooling/SocketConnectionPool.java
```diff
@@ -68,26 +71,41 @@ public SocketConnectionPool( PoolSettings poolSettings, Connector connector, Clo
     public Connection acquire( final BoltServerAddress address )
     {
         assertNotClosed();
+        BlockingPooledConnectionQueue connectionQueue = pool( address );
+        PooledConnection connection = acquireConnection( address, connectionQueue );
```
zhenlineo, line 75: Not related to this PR but wonder if it would be better API if we let the pooled connection to hold a reference to connecitonValidator and let the pooledConnection to provide methods to test isConnected and isReusable.

driver/src/main/java/org/neo4j/driver/v1/Config.java
```diff
@@ -256,30 +261,53 @@ public ConfigBuilder withMaxIdleSessions( int size )
...
         {
+            long idleTimeBeforeConnectionTestMillis = unit.toMillis( value );
+            if ( idleTimeBeforeConnectionTestMillis <= 0 )
+            {
+                throw new IllegalArgumentException( String.format(
```
zhenlineo: Why not let the user to set to -1 (default value)?

zhenlineo (changes requested):

driver/src/main/java/org/neo4j/driver/internal/net/pooling/SocketConnectionPool.java
lutovich: It could also simplify things around these suppliers and consumers. Let's think about it.

driver/src/main/java/org/neo4j/driver/internal/net/pooling/SocketConnectionPool.java
lutovich: Changed code to track when connection was created and then return it. So it'll retry until new connection is created and then return it to the user.

lutovich:
@zhenlineo comments addressed

driver/src/main/java/org/neo4j/driver/v1/Config.java
```diff
@@ -256,30 +261,47 @@ public ConfigBuilder withMaxIdleSessions( int size )
...
+         */
+        @Deprecated
+        public ConfigBuilder withSessionLivenessCheckTimeout( long timeout )
+        {
+            withConnectionLivenessCheckTimeout( timeout, TimeUnit.MILLISECONDS );
+            return this;
+        }
+
+        /**
+         * Pooled sessions that have been idle in the pool for longer than this timeout
```
zhenlineo: Pls use `Pooled connections` instead as `pooled sessions` is a wrong terminology used before.

driver/src/main/java/org/neo4j/driver/v1/Config.java
```diff
@@ -116,17 +123,14 @@ public int maxIdleConnectionPoolSize()
     }
 
     /**
-     * Pooled connections that have been unused for longer than this timeout will be tested before they are
-     * used again, to ensure they are still live.
+     * Pooled sessions that have been idle in the pool for longer than this timeout
```
zhenlineo: pooled "connections"

zhenlineo (commented):
LGTM besides a minor doc change suggestion.

zhenlineo (approved):

PR #299, opened 2017-01-02
Abstract PooledConnection interface to abstract pooling logic out of Connection

zhenlineo:
Session will hold a PooledConnection rather than Connect after this change.
This PR is a start step to introduce internal checked errors.


lutovich:
We've decided to not convert all exceptions to checked exceptions internally. Closing this PR for now.

Issue #302, opened 2017-01-03
(Bolt) Use of ServiceUnavailableException in TLS-enabled channel


nandosola:
Hi,

I'm trying to debug a potential issue in [neo4j-ogm](https://github.com/neo4j/neo4j-ogm) that might be related to Bolt session handling after a server restart (see also: spring-projects/spring-data-neo4j#373). Before going any further, I have a doubt concerning the Java Driver:

- For plaintext connections, the first operation after the restart results in a [`ServiceUnavailableException`](https://github.com/neo4j/neo4j-java-driver/blob/1.1/driver/src/main/java/org/neo4j/driver/internal/net/BufferingChunkedInput.java#L420) with the following message:

  > "Connection terminated while receiving data. This can happen due to network instabilities, or due to restarts of the database."

- For TLS-enabled connections,  the first operation after the restart results in a [`ClientException`](https://github.com/neo4j/neo4j-java-driver/blob/1.1/driver/src/main/java/org/neo4j/driver/internal/security/TLSSocketChannel.java#L166) with exactly the same message. 

My question is, why are there two different exceptions instead of just using `ServiceUnavailableException` for both?

----
Just FYI, `neo4j-ogm` doesn't handle `ServiceUnavailableException` yet. It just wraps  `ClientExceptions` coming from server disconnections or network glitches in `CypherException`, so the library user has to catch them and reconnect manually, invoking a new query.

zhenlineo:
We will have a look at this. It should be a ServiceUnavailable for TLS connection too by reading directly.

zhenlineo:
The bug should have been fixed and released in 1.1.2 driver. I am closing this issue but pls let me know if the problem still exists.

PR #304, opened 2017-01-09
Tls connections throw ServiceUnavaliable after restart of server

zhenlineo:
Fixed the bug where tls connections are failed with ClientException, while  plaintext connections are failed with ServiceUnav when server restart.
#302

### DIFF
--- a/driver/src/test/java/org/neo4j/driver/v1/integration/ServerKilledIT.java
+++ b/driver/src/test/java/org/neo4j/driver/v1/integration/ServerKilledIT.java
@@ -20,7 +20,12 @@ package org.neo4j.driver.v1.integration;
 
 import org.junit.Rule;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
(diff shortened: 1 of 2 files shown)


driver/src/test/java/org/neo4j/driver/v1/integration/ServerKilledIT.java
```diff
@@ -43,19 +48,33 @@
...
  * pool properly if the server dies, or all connections are lost for some other reason.
  */
+@RunWith(Parameterized.class)
 public class ServerKilledIT
 {
     @Rule
     public TestNeo4j neo4j = new TestNeo4j();
 
+    @Parameters(name = "{0} connections")
+    public static Collection<Object[]> data() {
```
lutovich, line 58: Can we make this method return a collection of `Config.EncryptionLevel`? I think it'll make code simpler - no need for ignored `testName` constructor argument and flat list of `Config.EncryptionLevel` can be returned.

lutovich (commented):
@zhenlineo looks good. Pointed out one possible simplification.

PR #305, opened 2017-01-16
Remove server from routingTable if failed to establish connection

zhenlineo:
* Fixed the bug where if the client failed to establish connection with the server, the server is not removed from the routing table
* Refactoring on routing table

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/cluster/ClusterRoutingTable.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/cluster/ClusterRoutingTable.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.neo4j.driver.internal.cluster;
+
+import java.util.HashSet;
+
+import org.neo4j.driver.internal.net.BoltServerAddress;
+import org.neo4j.driver.internal.util.Clock;
+
+import static java.lang.String.format;
+import static java.util.Arrays.asList;
+
+public class ClusterRoutingTable implements RoutingTable
+{
+    private static final int MIN_ROUTERS = 1;
+
+    private final Clock clock;
+    private long expirationTimeout;
+    private final RoundRobinAddressSet readers, writers, routers;
+
+    public ClusterRoutingTable( Clock clock, BoltServerAddress... routingAddresses )
+    {
+        this( clock );
+        routers.update( new HashSet<>( asList( routingAddresses ) ), new HashSet<BoltServerAddress>() );
+    }
+
+    private ClusterRoutingTable( Clock clock )
+    {
+        this.clock = clock;
+        this.expirationTimeout = clock.millis() - 1;
+
+        this.readers = new RoundRobinAddressSet();
+        this.writers = new RoundRobinAddressSet();
+        this.routers = new RoundRobinAddressSet();
+    }
+
+    @Override
+    public boolean stale()
+    {
+        return expirationTimeout < clock.millis() || // the expiration timeout has been reached
+               routers.size() <= MIN_ROUTERS || // we need to discover more routing servers
+               readers.size() == 0 || // we need to discover more read servers
+               writers.size() == 0; // we need to discover more write servers
+    }
+
+    @Override
+    public synchronized HashSet<BoltServerAddress> update( ClusterComposition cluster )
+    {
+        expirationTimeout = cluster.expirationTimestamp;
+        HashSet<BoltServerAddress> removed = new HashSet<>();
+        readers.update( cluster.readers(), removed );
+        writers.update( cluster.writers(), removed );
+        routers.update( cluster.routers(), removed );
+        return removed;
+    }
+
+    @Override
+    public synchronized void forget( BoltServerAddress address )
+    {
+        // Don't remove it from the set of routers, since that might mean we lose our ability to re-discover,
+        // just remove it from the set of readers and writers, so that we don't use it for actual work without
+        // performing discovery first.
+        readers.remove( address );
+        writers.remove( address );
+    }
+
+    @Override
+    public RoundRobinAddressSet readers()
+    {
+        return readers;
+    }
+
+    @Override
+    public RoundRobinAddressSet writers()
+    {
+        return writers;
+    }
+
+    @Override
+    public BoltServerAddress nextRouter()
+    {
+        return routers.next();
+    }
+
+    @Override
+    public int routerSize()
+    {
+        return routers.size();
+    }
+
+    @Override
+    public void removeWriter( BoltServerAddress toRemove )
+    {
+        writers.remove( toRemove );
+    }
+
+    @Override
+    public void removeRouter( BoltServerAddress toRemove )
+    {
+        routers.remove( toRemove );
+    }
+
+    @Override
+    public String toString()
+    {
+        return format( "Ttl %s, currentTime %s, routers %s, writers %s, readers %s",
+                expirationTimeout, clock.millis(), routers, writers, readers );
+    }
+}
--- a/driver/src/main/java/org/neo4j/driver/internal/cluster/ClusterRoutingTable.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/cluster/ClusterRoutingTable.java
@@ -20,6 +20,7 @@
 package org.neo4j.driver.internal.cluster;
 
 import java.util.HashSet;
+import java.util.Set;
 
 import org.neo4j.driver.internal.net.BoltServerAddress;
 import org.neo4j.driver.internal.util.Clock;
@@ -33,7 +34,9 @@ public class ClusterRoutingTable implements RoutingTable
 
     private final Clock clock;
     private long expirationTimeout;
-    private final RoundRobinAddressSet readers, writers, routers;
+    private final RoundRobinAddressSet readers;
+    private final RoundRobinAddressSet writers;
+    private final RoundRobinAddressSet routers;
 
     public ClusterRoutingTable( Clock clock, BoltServerAddress... routingAddresses )
     {
@@ -52,7 +55,7 @@ public class ClusterRoutingTable implements RoutingTable
     }
 
     @Override
-    public boolean stale()
+    public boolean isStale()
     {
         return expirationTimeout < clock.millis() || // the expiration timeout has been reached
                routers.size() <= MIN_ROUTERS || // we need to discover more routing servers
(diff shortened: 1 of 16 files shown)


driver/src/main/java/org/neo4j/driver/internal/cluster/ClusterRoutingTable.java
```diff
@@ -0,0 +1,126 @@
...
+import static java.lang.String.format;
+import static java.util.Arrays.asList;
+
+public class ClusterRoutingTable implements RoutingTable
+{
+    private static final int MIN_ROUTERS = 1;
+
+    private final Clock clock;
+    private long expirationTimeout;
+    private final RoundRobinAddressSet readers, writers, routers;
```
lutovich: Could you please split this line into separate declarations?

driver/src/main/java/org/neo4j/driver/internal/cluster/ClusterRoutingTable.java
```diff
@@ -0,0 +1,126 @@
...
+    public boolean stale()
+    {
+        return expirationTimeout < clock.millis() || // the expiration timeout has been reached
+               routers.size() <= MIN_ROUTERS || // we need to discover more routing servers
+               readers.size() == 0 || // we need to discover more read servers
+               writers.size() == 0; // we need to discover more write servers
+    }
+
+    @Override
+    public synchronized HashSet<BoltServerAddress> update( ClusterComposition cluster )
```
lutovich: It is better to have `Set<BoltServerAddress>` as the return type

driver/src/main/java/org/neo4j/driver/internal/cluster/ClusterRoutingTable.java
```diff
@@ -0,0 +1,126 @@
...
+        this.clock = clock;
+        this.expirationTimeout = clock.millis() - 1;
+
+        this.readers = new RoundRobinAddressSet();
+        this.writers = new RoundRobinAddressSet();
+        this.routers = new RoundRobinAddressSet();
+    }
+
+    @Override
+    public boolean stale()
```
lutovich: Maybe `isStale()`?

driver/src/main/java/org/neo4j/driver/internal/cluster/Rediscovery.java
```diff
@@ -0,0 +1,95 @@
...
+
+import static java.lang.String.format;
+
+public class Rediscovery
+{
+    private static final String NO_ROUTERS_AVAILABLE = "Could not perform discovery. No routing servers available.";
+
+    // Given a old routing table and connection pool, use the connection composition provider to obtain a new
+    // cluster composition, which would be used to update the routing table and connection pool
+    public static ClusterComposition lookupRoutingTable(
```
lutovich: Maybe make `LoadBalancer` create an instance of `Rediscovery` with `RoutingSettings`, `Clock`, etc. Otherwise this method has too many parameters. It would also make testing of `LoadBalancer` easier with custom `Rediscovery`.

driver/src/main/java/org/neo4j/driver/internal/cluster/Rediscovery.java
```diff
@@ -0,0 +1,95 @@
...
+                {
+                    log.error( format( "Failed to connect to routing server '%s'.", address ), e );
+                    continue;
+                }
+                if ( cluster == null || !cluster.isValid() )
+                {
+                    log.info(
+                            "Server <%s> unable to perform routing capability, dropping from list of routers.",
+                            address );
+                    routingTable.removeRouter( address );
```
lutovich, line 87: Do we still want to remove a router here?

driver/src/main/java/org/neo4j/driver/internal/cluster/RoutingTable.java
```diff
@@ -0,0 +1,44 @@
...
+
+    HashSet<BoltServerAddress> update( ClusterComposition cluster );
+
+    void forget( BoltServerAddress address );
+
+    RoundRobinAddressSet readers();
+
+    RoundRobinAddressSet writers();
+
+    BoltServerAddress nextRouter();
```
lutovich, line 37: Could we maybe replace `nextRouter()` and `routerSize()` with `RoundRobinAddressSet routers()`? Then callers would call `routingTable.routers().next()` and `routingTable.routers().size()`.

driver/src/main/java/org/neo4j/driver/internal/net/ChannelFactory.java
```diff
@@ -69,7 +65,7 @@ private static void connect( SocketChannel soChannel, BoltServerAddress address,
         }
         catch ( SocketTimeoutException e )
         {
-            throw new IOException( "Timeout " + timeoutMillis + "ms expired", e );
+            throw new ConnectException( "Timeout " + timeoutMillis + "ms expired" + e.getMessage() );
```
lutovich: I think it is better to keep `e` as an argument, not just `e.getMessage()`

driver/src/main/java/org/neo4j/driver/internal/net/SocketClient.java
```diff
@@ -120,7 +119,10 @@ public void start()
         try
         {
             logger.debug( "~~ [CONNECT] %s", address );
-            setChannel( ChannelFactory.create( address, securityPlan, timeoutMillis, logger ) );
+            if( channel == null )
```
lutovich, line 122: Could we instead assert that channel is `null` as the first thing in this method and throw `IllegalStateException` if it is not null. This will disallow multiple `#start()` calls.

driver/src/main/java/org/neo4j/driver/internal/security/TLSSocketChannel.java
```diff
@@ -79,7 +81,21 @@ public TLSSocketChannel( ByteChannel channel, Logger logger, SSLEngine sslEngine
...
+            runHandshake();
+        }
+        catch ( SSLHandshakeException e )
+        {
+            Throwable error = e;
+            if( error.getCause() != null ) // get the real exception
+            {
+                error = e.getCause();
+            }
+            throw new ClientException( "Failed to establish secured connection with the server: " + error.getMessage(),
```
lutovich: It is probably better to move this `try-catch` block inside `runHandshake()` and just re-throw `ClientException` with whatever `IOException` is thrown as the cause.

driver/src/main/java/org/neo4j/driver/internal/security/TLSSocketChannel.java
```diff
@@ -164,6 +180,14 @@ private HandshakeStatus unwrap( ByteBuffer buffer ) throws IOException
          */
         if ( channel.read( cipherIn ) < 0 )
         {
+            try
+            {
+                channel.close();
```
lutovich: Please add a test for this fix

driver/src/main/java/org/neo4j/driver/internal/security/TLSSocketChannel.java
```diff
@@ -289,14 +313,19 @@ private HandshakeStatus wrap( ByteBuffer buffer ) throws IOException
...
                 {
-                    throw new ClientException(
-                            String.format(
-                                    "Failed to enlarge network buffer from %s to %s. This is either because the " +
-                                    "new size is however less than the old size, or because the application " +
-                                    "buffer size %s is so big that the application data still cannot fit into the " +
-                                    "new network buffer.", curNetSize, netSize, buffer.capacity() ) );
+                    try
+                    {
+                        channel.close();
```
lutovich: Please add a test for this fix.

driver/src/test/java/org/neo4j/driver/internal/net/SocketClientTest.java
```diff
@@ -22,32 +22,69 @@
...
 
     @Rule
     public ExpectedException exception = ExpectedException.none();
 
+    @Parameters(name = "{0} connections")
+    public static Collection<Object[]> data()
+    {
+        return Arrays.asList(new Object[][] {
+                { "insecure", SecurityPlan.insecure() },
+                { "encrypted", createEncryptedSecurityPlan() }
```
lutovich: Can we just use `SecurityPlan.forAllCertificates()` here and declare `#data()` to throw `Exception`?

driver/src/main/java/org/neo4j/driver/internal/net/SocketClient.java
```diff
@@ -133,11 +135,7 @@ public void start()
         }
         catch ( IOException e )
         {
-            throw new ClientException( "Unable to process request: " + e.getMessage(), e );
-        }
-        catch ( GeneralSecurityException e )
-        {
-            throw new ClientException( "Unable to establish ssl connection with server: " + e.getMessage(), e );
+            throw new ServiceUnavailableException( "Unable to process request: " + e.getMessage(), e );
```
lutovich, line 138: Could a test be added to assert that router is removed when this exception is thrown?

lutovich (changes requested):
@zhenlineo review completed. Made some comments.

driver/src/main/java/org/neo4j/driver/internal/cluster/Rediscovery.java
zhenlineo: I am not trying to change existing logic in this PR.

driver/src/main/java/org/neo4j/driver/internal/cluster/RoutingTable.java
zhenlineo: It was a on purpose design to make it easier to mock/override to return next router as demands. See the class named `ListBasedRoutingTable`

driver/src/main/java/org/neo4j/driver/internal/net/SocketClient.java
zhenlineo: The check here is to allow injection of a mocked channel into this class for testing. This null check is for `do not create a new channel if I already give you a fake channel`.

driver/src/main/java/org/neo4j/driver/internal/net/ChannelFactory.java
zhenlineo: ConnectException has no constructor with `e` as aurgument

lutovich: Maybe we can keep `IOException`? Javadoc for `ConnectException` says "Typically, the connection was refused remotely (e.g., no process is listening on the remote address/port)" which is a bit different in our case.

driver/src/main/java/org/neo4j/driver/internal/security/TLSSocketChannel.java
```diff
@@ -67,10 +68,26 @@ public static TLSSocketChannel create( BoltServerAddress address, SecurityPlan s
...
+    {
+        TLSSocketChannel tlsChannel = new TLSSocketChannel( channel, logger, sslEngine );
+        try
+        {
+            tlsChannel.runHandshake();
+        }
+        catch ( SSLHandshakeException e )
+        {
+            throw new ClientException( "Failed to establish secured connection with the server: " + e.getMessage(),
+                    e.getCause() );
```
lutovich: No reason to get the cause.

driver/src/main/java/org/neo4j/driver/internal/security/TLSSocketChannel.java
```diff
@@ -288,16 +345,7 @@ private HandshakeStatus wrap( ByteBuffer buffer ) throws IOException
...
-                if (written == 0)
-                {
-                    throw new ClientException(
-                            String.format(
-                                    "Failed to enlarge network buffer from %s to %s. This is either because the " +
-                                    "new size is however less than the old size, or because the application " +
-                                    "buffer size %s is so big that the application data still cannot fit into the " +
-                                    "new network buffer.", curNetSize, netSize, buffer.capacity() ) );
-                }
+                channelWrite( cipherOut );
```
lutovich: Let's keep `written == 0` check

lutovich (approved):

PR #312, opened 2017-01-25
Throw ServiceUnavailableException when socket write fails

lutovich:
Previously `ClientException` was thrown which was incorrect and resulted in connections not being properly disposed.

Also made small cleanup of unused/test-only methods and classes.


zhenlineo:
Could you add test for encrypted connections too?

lutovich:
@zhenlineo changed test in `CausalClusteringIT` uses encrypted channel with basic auth. Also rebased this PR.

PR #315, opened 2017-01-31
Throw AuthenticationException if failed to conn due to auth errors

zhenlineo:
This includes the error that the client provide a wrong credentials and
the server provides a wrong certificate.


driver/src/main/java/org/neo4j/driver/internal/security/TLSSocketChannel.java
```diff
@@ -81,7 +82,7 @@ public static TLSSocketChannel create( ByteChannel channel, Logger logger, SSLEn
         }
         catch ( SSLHandshakeException e )
         {
-            throw new ClientException( "Failed to establish secured connection with the server: " + e.getMessage(), e );
+            throw new UnauthorizedException( "Failed to establish secured connection with the server: " + e.getMessage(), e );
```
technige: I'm not sure that we should reuse `UnauthorizedException` for TLS handshake failure as well as for actual login failure. The Python driver uses `ServiceUnavailable` here but that doesn't seem quite right either.

Perhaps we should have something separate such as `SecurityLayerUnavailable`?

zhenlineo: I would prefer to have fewer public errors. So how about change both of them to `SecurityException`/`AuthorizationException`?

lutovich: +1 for `AuthorizationException`

lutovich (commented):
@zhenlineo changes look good, name of `UnauthorizedException` is probably the only thing that could be improved.

driver/src/main/java/org/neo4j/driver/internal/security/TLSSocketChannel.java
technige: I think the name is fine as-is for login failure but why would fewer errors be useful here? Login failure _could_ trigger an application code path that requires the user to enter new login details whereas an SSL handshake failure should not.

lutovich (approved):

zhenlineo:
Will merge this one latter after the big routing specification change, as it should be easier to rebase this PR

PR #317, opened 2017-02-02
Hydration

technige:
Extracts hydration (turning structures into native objects) functionality and tidies up some PackStream internals.


zhenlineo:
@technige Can we delete this branch?

PR #320, opened 2017-02-12
Rename hashcode method to hashCode to override Object.hashCode()

szarnyasg:
The `hashCode()` method in the [`InternalRecord`](https://github.com/neo4j/neo4j-java-driver/blob/262224400350e0f9db41be7db5d75be8ed880f7a/driver/src/main/java/org/neo4j/driver/internal/InternalRecord.java#L170) class is spelt with a lowercase `c`, hence it is not executed when running `hashCode()`. Instead, the `Object.hashCode()` method is executed, which does not work well for `InternalRecord` instances and gives different hash values for tuples with the same key-value pairs.

This pull request fixes that by renaming the method appropriately.

Additionally, the [`Record`](https://github.com/neo4j/neo4j-java-driver/blob/262224400350e0f9db41be7db5d75be8ed880f7a/driver/src/main/java/org/neo4j/driver/v1/Record.java#L134) interface tries to enforce the implementation of `equals()` and `hashCode()`. However, it seems it is not possible to enforce the implementation of the these methods by using an interface - if they are not overridden in the class, the implementation of the `Object` method will be used.

As a minimal example, this interface and its implementation compile without errors.

```java
public interface MyInterface {

    @Override
    int hashCode();

}

public class MyImplementation implements MyInterface {

}
```

One could enforce their implementation is to use an abstract class. The following code does not compile.

```java
public abstract class MyAbstractClass {

    @Override
    public abstract int hashCode();

}


public class MyConcreteClass extends MyAbstractClass {

}
```

However, this solution cannot be applied to `Record` as the `InternalRecord` class already implements another class (`InternalMapAccessorWithDefaultValue`).

In this pull request, I also deleted the `@Override` definitions for `equals` and `hashCode` in the `Record` interface as they basically have no effect.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/InternalRecord.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/InternalRecord.java
@@ -41,7 +41,7 @@ public class InternalRecord extends InternalMapAccessorWithDefaultValue implemen
 {
     private final List<String> keys;
     private final Value[] values;
-    private int hashcode = 0;
+    private int hashCode = 0;
 
     public InternalRecord( List<String> keys, Value[] values )
     {
@@ -132,6 +132,7 @@ public class InternalRecord extends InternalMapAccessorWithDefaultValue implemen
         return format( "Record<%s>", formatPairs( InternalValue.Format.VALUE_ONLY, asMap( ofValue() ) ) );
     }
 
+    @Override
     public boolean equals( Object other )
     {
         if ( this == other )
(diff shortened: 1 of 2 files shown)


lutovich:
Hi @szarnyasg.

Thanks a lot for the PR! Changes look good, this is clearly an oversight from our side. Could you please also add `@Override` annotation to `#equals()` and `#hashCode()` methods in `InternalRecord`?

To be able to merge your changes we will need a signed CLA from you. See http://neo4j.com/developer/cla/ for further information. Additional information on contributing code is found at http://neo4j.com/developer/contributing-code/.

szarnyasg:
@lutovich I emailed a CLA to `cla@neotechnology.com` last year with the following timestamp: Sat, Sep 24, 2016 at 4:18 PM, but did not receive a response. Do I have to print the PDF, undersign on paper and scan it? This was not entirely clear for me from the [CLA guide](https://neo4j.com/developer/cla/).

I'll add the changes in a minute.

szarnyasg:
Update: I added the override annotations.

lutovich:
@szarnyasg I'm sorry but I can't see your CLA in the inbox. Could you please resend it? You need to only attach that PDF to the letter, no need to print and scan it.

lutovich:
@szarnyasg found your CLA. Tried to search for your name but it is spelled differently in the CLA inbox and on GitHub :)
Will start a new PR build and merge when it turns green. Thanks!

szarnyasg:
Thanks - the wonders of accented characters :-). I'll pay attention to this in the future.

PR #322, opened 2017-02-20
Move error handling logic to the connection layer

lutovich:
It was previously spread between `RoutingSession`, `RoutingTransaction` and `RoutingStatementResult` and they all did the same thing when connection errors happened.

This PR introduces `RoutingPooledConnection` that handles errors for routing driver by removing entries from the routing table and purging connection pool for failed address.


driver/src/main/java/org/neo4j/driver/internal/cluster/RoutingPooledConnection.java
```diff
@@ -0,0 +1,289 @@
...
+
+    private RuntimeException handledClientException( ClientException e )
+    {
+        if ( isFailureToWrite( e ) )
+        {
+            // The server is unaware of the session mode, so we have to implement this logic in the driver.
+            // In the future, we might be able to move this logic to the server.
+            switch ( accessMode )
+            {
+            case READ:
```
zhenlineo, line 272: We actually do not need this `READ/WRITE` mode as the error itself already self explained:

`Neo.ClientError.Cluster.NotALeader` -> server at xx no longer accepts writes
`Neo.ClientError.General.ForbiddenOnReadOnlyDatabase` -> Write queries cannot be performed in `READ` access mode

So we could totally remove access mode from this class

driver/src/main/java/org/neo4j/driver/internal/spi/PooledConnection.java
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.neo4j.driver.internal.spi;
+
+public interface PooledConnection extends Connection
+{
+    long lastUsedTimestamp();
```
zhenlineo: We should also move 
`onError`
`hasUnrecoverableErrors` 
from `Connection` to `PooledConnection`

driver/src/main/java/org/neo4j/driver/internal/cluster/RoutingPooledConnection.java
```diff
@@ -0,0 +1,289 @@
...
+            default:
+                throw new IllegalArgumentException( accessMode + " not supported." );
+            }
+        }
+        return e;
+    }
+
+    private static boolean isFailureToWrite( ClientException e )
+    {
+        return e.code().equals( "Neo.ClientError.Cluster.NotALeader" ) ||
```
zhenlineo: Warning: you might get `ClientException` without code field at this stage :)

zhenlineo (changes requested):
+1

driver/src/main/java/org/neo4j/driver/internal/cluster/RoutingPooledConnection.java
lutovich: It is possible to get `NotALeader` when trying to write in a read session with connection towards follower and `ForbiddenOnReadOnlyDatabase` when connection is towards read replica. So it looks like we still need this switch on the access mode.

lutovich:
@zhenlineo comments should be addressed now

zhenlineo (approved):

PR #324, opened 2017-02-23
Decouple session from connection

lutovich:


driver/src/main/java/org/neo4j/driver/internal/InternalDriver.java
```diff
@@ -61,42 +64,57 @@ public final Session session()
...
+        Session session = sessionFactory.newInstance( mode );
+        if ( closed.get() )
         {
             // the driver is already closed and we either 1. obtain this session from the old session pool
             // or 2. we obtain this session from a new session pool
             // For 1. this closeResources will take no effect as everything is already closed.
             // For 2. this closeResources will close the new connection pool just created to ensure no resource leak.
             closeResources();
-            throw driverCloseException();
+            throw new DriverClosedException();
```
zhenlineo: Maybe not a new exception type in driver for now

driver/src/main/java/org/neo4j/driver/internal/InternalStatementResult.java
```diff
@@ -319,6 +321,7 @@ private boolean tryFetchNext()
         {
             if ( done )
             {
+                connectionHandler.resultBuffered();
```
zhenlineo: if you do `connectionHandler.resultBuffered()` at the place where `done` is set to be true, then you probably could also address failure/ignore cases too.

driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
```diff
@@ -132,21 +118,23 @@ public synchronized void reset()
...
         if ( currentTransaction != null )
         {
             currentTransaction.markToClose();
             lastBookmark = currentTransaction.bookmark();
             currentTransaction = null;
         }
-        connection.resetAsync();
+        if ( currentConnection != null )
+        {
+            currentConnection.resetAsync();
```
zhenlineo, line 124: I would suggest to remove reset. Make this method deprecated and then give an empty impl inside.

driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
```diff
@@ -241,25 +203,42 @@ public TypeSystem typeSystem()
...
+        closeCurrentConnection( true );
+    }
+
+    @Override
+    public synchronized void transactionClosed( ExplicitTransaction tx )
+    {
+        if ( currentTransaction != null && currentTransaction == tx )
+        {
+            lastBookmark = currentTransaction.bookmark();
+            currentTransaction = null;
```
zhenlineo: connection also need to be closed

driver/src/main/java/org/neo4j/driver/internal/net/pooling/PooledSocketConnection.java
```diff
@@ -58,7 +59,7 @@
     private final Consumer<PooledConnection> release;
 
     private boolean unrecoverableErrorsOccurred = false;
-    private Runnable onError = null;
+    private ConnectionHandler connectionHandler;
```
zhenlineo: suggesting rename something else such as `ConnectionTerminationHandler`?

driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
lutovich: Can we maybe remove it in a separate PR? This one is quite big already :)

driver/src/test/java/org/neo4j/driver/internal/InternalStatementResultTest.java
```diff
@@ -386,23 +386,89 @@ public void shouldNotPeekIntoTheFutureWhenResultIsEmpty()
...
+        verify( resourcesHandler ).onResultConsumed();
+    }
+
+    @Test
+    public void shouldNotifyResourcesHandlerWhenSummary()
+    {
+        SessionResourcesHandler resourcesHandler = mock( SessionResourcesHandler.class );
+        StatementResult result = createResult( 10, resourcesHandler );
+
+        assertNotNull( result.summary() );
```
zhenlineo, line 433: maybe summary several times to ensure only called once?
Similar to other methods and also some combination of all the methods? such as result.single(), result.summary(), then result.consume()?

lutovich: Added a separate test for this here: https://github.com/neo4j/neo4j-java-driver/pull/324/commits/0ddac3d5036b7017f80f6592ba4815b0b00bcd87

PR #325, opened 2017-02-27
API updates

lutovich:
This PR adds following new API methods:

 * `Driver#session(String bookmark)` to allow specifying initial bookmark for the session which will be used by the first transaction. This method is mainly added to bring symmetry with regards to `Session#lastBookmark()` and to emphasize that bookmark is a session concept.
 * `Driver#session(AccessMode mode, String bookmark)` to allow specifying both initial bookmark and default `AccessMode` per session.
 * `Session#readTransaction(Function<Transaction, T> work)` to allow declarative read transactions with exponential backoff retry policy.
 * `Session#writeTransaction(Function<Transaction, T> work)` to allow declarative write transactions with exponential backoff retry policy.
 * `ConfigBuilder#withMaxTransactionRetryTime(int value, TimeUnit unit)` to configure the retry policy for `#readTransaction()` and `#writeTransaction()`.

It also deprecates:
 * `Session#beginTransaction(String bookmark)` in favour or `Driver#session(bookmark)`


driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
```diff
@@ -155,28 +164,41 @@ public void close()
...
-        return beginTransaction( null );
+        return beginTransaction( mode );
     }
 
     @Override
     public synchronized Transaction beginTransaction( String bookmark )
     {
-        ensureSessionIsOpen();
-        ensureNoOpenTransactionBeforeOpeningTransaction();
+        setLastBookmark( bookmark );
```
zhenlineo: Why we disallow a user to set null bookmark? (It does not make many sense, but maybe a user wants to break CC from this tx?)

driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
```diff
@@ -225,6 +247,47 @@ public synchronized void onConnectionError( boolean recoverable )
...
+            {
+                return work.apply( tx );
+            }
+            catch ( Throwable newError )
+            {
+                decision = retryLogic.apply( newError, decision );
+
+                if ( decision.shouldRetry() )
+                {
+                    errors = recordError( newError, errors );
```
zhenlineo, line 264: Can we just directly chain the errors? why we have to first remember them (in `recordError`) and then build the error chain (in `addSuppressed`)?

zhenlineo:
Looks good, minor questions. The retry logic impl is a bit scaring

driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
lutovich: Good point, I just tried to use same `#setLastBookmark()` method everywhere and it prohibits nulls. Will fix.

driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
lutovich: Printed stacktrace of a chain of suppressed methods looks quite strange, every next suppressed trace is shifted to the right. With this approach all previous errors will be on the same level :)

lutovich:
Thanks @zhenlineo! I agree about retry logic, will try to simplify it in subsequent PRs. There is no point in having separate `RetryLogic` and `RetryDecision`...

Issue #205, opened 2016-07-26
question about InternalSession.isOpen method


inserpio:
Hi All, I'm Lorenzo from the [Neo4j JDBC driver](https://github.com/neo4j-contrib/neo4j-jdbc) team.

I'm contacting you because we have this open [issue 69](https://github.com/neo4j-contrib/neo4j-jdbc/issues/69) that looks like to be somehow related to your [InternalSession.isOpen](https://github.com/neo4j/neo4j-java-driver/blob/1.1/driver/src/main/java/org/neo4j/driver/internal/InternalSession.java#L101) method.
In fact, the guy that opened this issue is also proposing to slightly change this your method to solve his problem (please see: [issue 69 comment](https://github.com/neo4j-contrib/neo4j-jdbc/issues/69#issuecomment-234210383)).

Design choice for the JDBC Driver's BOLT implementation was to wrap as much as possible this your java driver (carefully monitoring the overhead we're introducing not to lower performances): that's why we'd like to keep calling your methods instead of rewriting the same logic of you.

So, what's your opinion about this our issue? Do you think this is something we'd solve from our side or the proposed patch sounds reasonable for you? Or whatelse?

Cheers,
Lorenzo

zhenlineo:
We actually made the `connection.isOpen` and `session.isOpen` differently on purpose. `connection.isOpen` test if the socket connection with a server has been closed, whereas `session.isOpen` test if the session has free the connection to so-called connection pool.

In our bolt driver design, a connection refers to a real socket connection between a driver and a server,  while a session is a logical idea of a "conversation" with the server. We expect a user to run a cypher statement and then "close" this "conversation" (session). When the user closes this session, we do not close the connection but return it in the connection pool so that we could reuse it latter. Therefore we expect a connection will not be closed before a session.

By reading the original issue, I was wondering why the connection is closed before the session. In our driver code, the only way that the connection is closed before the session is that we received a fatal bolt protocol error message. So it might because a bug in our code that result in the user having the problems.

Could you ask the user if he has received some exception before he hits this `connection already closed` error?

Thanks in advance.
Zhen

inserpio:
Thanks for the reply @zhenlineo and for the useful clarification about the driver design.
I'll ask some more info to that guy and ping you back.
Cheers,
Lorenzo

troyplay:
[...]

PR #327, opened 2017-03-01
Deprecate Session#reset()

lutovich:
Related to https://github.com/neo4j/neo4j-java-driver/issues/205.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/v1/Session.java
+++ b/driver/src/main/java/org/neo4j/driver/v1/Session.java
@@ -116,7 +116,16 @@ public interface Session extends Resource, StatementRunner
      * any statement that is currently executing and ignores any subsequently queued statements. Following
      * the reset, the current transaction will have been rolled back and any outstanding failures will
      * have been acknowledged.
+     *
+     * @deprecated This method should not be used and violates the expected usage pattern of {@link Session} objects.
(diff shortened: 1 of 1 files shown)


driver/src/main/java/org/neo4j/driver/v1/Session.java
```diff
@@ -116,7 +116,16 @@
...
      * the reset, the current transaction will have been rolled back and any outstanding failures will
      * have been acknowledged.
+     *
+     * @deprecated This method should not be used and violates the expected usage pattern of {@link Session} objects.
+     * They are expected to be not thread-safe and should not be shared between thread. However this method is only
+     * useful when {@link Session} object is passed to another monitoring thread that calls it when appropriate.
+     * It is not useful when {@link Session} is used in a single thread because in this case {@link #close()}
+     * can be used. Since version 3.1, Neo4j database allows users to specify maximum transaction execution time and
+     * contains procedures to list and terminate running queries. These functions should be used instead of calling
+     * this method. <b>Will be removed in the next major release.</b>
```
technige: Can you remove "Will be removed in the next major release". We'll deprecate to indicate this will not be ported to other drivers and should be avoided whenever possible but will not be replaced until the async API is introduced.

PR #332, opened 2017-03-10
Resolve DNS to IP & IPv6 address support

zhenlineo:


driver/src/main/java/org/neo4j/driver/internal/cluster/ClusterRoutingTable.java
```diff
@@ -78,9 +78,7 @@ public boolean isStale()
     @Override
     public synchronized void forget( BoltServerAddress address )
     {
-        // Don't remove it from the set of routers, since that might mean we lose our ability to re-discover,
-        // just remove it from the set of readers and writers, so that we don't use it for actual work without
-        // performing discovery first.
+        routers.remove( address );
```
lutovich, line 81: This `#forget()` method is also called on connection errors. Do we really want to drop routers here? I'd vote for not changing this and keeping `#removeRouter()`.

driver/src/main/java/org/neo4j/driver/internal/cluster/DnsResolver.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.HashSet;
+import java.util.Set;
+
+import org.neo4j.driver.internal.net.BoltServerAddress;
+import org.neo4j.driver.v1.exceptions.ServiceUnavailableException;
+
+public interface DnsResolver
```
lutovich: Could we maybe have `HostNameResolver` interface and then `DnsResolver` as it's implementation?

driver/src/main/java/org/neo4j/driver/internal/cluster/DnsResolver.java
```diff
@@ -0,0 +1,58 @@
...
+{
+    DnsResolver DEFAULT = new DnsResolver()
+    {
+        @Override
+        public Set<BoltServerAddress> resolve( BoltServerAddress initialRouter )
+        {
+            try
+            {
+                InetAddress[] ipAddresses = InetAddress.getAllByName( initialRouter.host() );
+                Set<BoltServerAddress> addresses = new HashSet<>( ipAddresses.length );
```
lutovich: Specified like this initial capacity might still cause `HashSet` to resize because of default load factor or 0.75. Maybe not bother with initial size at all? :)

driver/src/main/java/org/neo4j/driver/internal/cluster/DnsResolver.java
```diff
@@ -0,0 +1,58 @@
...
+                {
+                    addresses.add( new BoltServerAddress( ipAddresses[i].getHostAddress(), initialRouter.port() ) );
+                }
+
+                return addresses;
+
+            }
+            catch ( UnknownHostException e )
+            {
+                throw new ServiceUnavailableException(
```
lutovich: What will happen if seed uri is already an IP address? Will `UnknownHostException` be thrown?
What do you think about just logging failure here and returning `Collections.singleton(initialRouter)`?

driver/src/main/java/org/neo4j/driver/internal/cluster/Rediscovery.java
```diff
@@ -76,8 +81,8 @@ public ClusterComposition lookupClusterComposition( ConnectionPool connections,
     private ClusterComposition lookupClusterCompositionOnKnownRouters( ConnectionPool connections,
             RoutingTable routingTable )
     {
-        boolean triedInitialRouter = false;
         int size = routingTable.routerSize();
+        Set<BoltServerAddress> triedServers = new HashSet<>( size );
```
lutovich: `HashSet`'s initial capacity might be tricky here as well.

driver/src/test/java/org/neo4j/driver/internal/cluster/RediscoveryTest.java
```diff
@@ -454,10 +455,22 @@ private static ClusterComposition rediscover( BoltServerAddress initialRouter, C
         Clock mockedClock = mock( Clock.class );
         Logger mockedLogger = mock( Logger.class );
 
-        Rediscovery rediscovery = new Rediscovery( initialRouter, settings, mockedClock, mockedLogger, provider );
+        Rediscovery rediscovery = new Rediscovery( initialRouter, settings, mockedClock, mockedLogger, provider,
+                directMapProvider );
         return rediscovery.lookupClusterComposition( connections, routingTable );
     }
 
+    private static DnsResolver directMapProvider = new DnsResolver()
```
lutovich: Maybe introduce a `private static class` for this or move it to the top of the class? It is just a bit strange to see field in the middle of the class.

driver/src/test/java/org/neo4j/driver/internal/cluster/RediscoveryTest.java
```diff
@@ -454,10 +455,22 @@ private static ClusterComposition rediscover( BoltServerAddress initialRouter, C
...
+    {
+        @Override
+        public Set<BoltServerAddress> resolve( BoltServerAddress initialRouter )
+        {
+            Set<BoltServerAddress> directMap = new HashSet<>( 1 );
```
lutovich: Could probably use `Collections.singleton( initialRouter )` here

lutovich (changes requested):
@zhenlineo review round completed. Do you think it is possible to add a boltkit test with some fake resolver to verify this feature end to end?

driver/src/main/java/org/neo4j/driver/internal/cluster/ClusterRoutingTable.java
zhenlineo: What's the point to keep in routing table? they were there to protect from running out of routers. Now we no longer have the problem with dns

driver/src/main/java/org/neo4j/driver/internal/cluster/DnsResolver.java
zhenlineo: IP to IP is not a problem, the test include using `127.0.0.1`

zhenlineo:
@lutovich All boltkit tests that creates a real driver already go via DnsResolver

driver/src/test/java/org/neo4j/driver/internal/cluster/RediscoveryTest.java
zhenlineo: No, I cannot return any singlention set as it does not support `removeAll` method

lutovich (approved):

PR #333, opened 2017-03-13
Deprecation cleanup and beginTx behaviour fix

technige:


driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
```diff
@@ -174,10 +175,14 @@ public synchronized Transaction beginTransaction()
         return beginTransaction( mode );
     }
 
+    @Deprecated
     @Override
     public synchronized Transaction beginTransaction( String bookmark )
     {
-        lastBookmark = bookmark;
+        if ( bookmark != null )
```
zhenlineo, line 182: How about move the check of null into `setBookmark` method and invoke the method from here? So that all the places where we want to set bookmark will just call a single `setBookmark` method.

driver/src/test/java/org/neo4j/driver/internal/NetworkSessionTest.java
```diff
@@ -604,25 +606,36 @@ public void setLastBookmark()
     public void possibleToOverwriteBookmarkWithNull()
     {
         NetworkSession session = newSession( mock( ConnectionProvider.class ), WRITE );
-        session.setLastBookmark( "TheBookmark" );
+        session.setBookmark( "TheBookmark" );
```
zhenlineo: The test should be rewrite to `impossibleToOverwriteBookmarkWithNull`

examples/src/main/java/org/neo4j/docs/driver/Examples.java
```diff
@@ -264,6 +264,7 @@ public static Driver requireEncryption() throws Exception
         return driver;
     }
 
+    // TODO: remove this example now that TOFU is deprecated
```
lutovich, line 267: Can we execute this TODO now?

lutovich:
@technige changes look good, there are couple small comments to address.

zhenlineo:
Also pls re-open the PR targeting at 1.2 instead of 1.3 branch.

driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
technige: `setBookmark` will still have to support `null` as we don't have a separate `clearBookmark` method. But using the setter inside the `if` block is a good idea.

driver/src/test/java/org/neo4j/driver/internal/NetworkSessionTest.java
technige: Actually `setBookmark` is an internal method only so the test can be removed.

examples/src/main/java/org/neo4j/docs/driver/Examples.java
technige: Not as part of this PR as it'll break the manual.

technige:
Closing and reopening against 1.2

PR #334, opened 2017-03-13
Add support of passing routing parameters via driver config

zhenlineo:
Added `withRoutingParameters` in driver Config for passing in getServersV2 parameters

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/cluster/ServerVersion.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/cluster/ServerVersion.java
@@ -16,7 +16,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.neo4j.driver.v1.util;
+package org.neo4j.driver.internal.cluster;
 
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -44,6 +44,7 @@ public class ServerVersion
     public static final ServerVersion v3_2_0 = new ServerVersion(3, 2, 0);
     public static final ServerVersion v3_1_0 = new ServerVersion(3, 1, 0);
     public static final ServerVersion v3_0_0 = new ServerVersion(3, 0, 0);
+    public static final ServerVersion v3_2_0 = new ServerVersion(3, 2, 0);
 
     public static ServerVersion version( Driver driver )
     {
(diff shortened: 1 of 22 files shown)


driver/src/main/java/org/neo4j/driver/internal/cluster/ServerVersion.java
```diff
@@ -16,7 +16,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.neo4j.driver.v1.util;
+package org.neo4j.driver.internal.cluster;
```
lutovich: Maybe `org.neo4j.driver.internal.util`?

driver/src/test/java/org/neo4j/driver/internal/RoutingDriverTest.java
```diff
@@ -148,8 +149,8 @@ public void shouldFailIfNoRouting()
         // Then
         catch ( ServiceUnavailableException e )
         {
-            assertThat( e.getMessage(),
-                    containsString( "Failed to call 'dbms.cluster.routing.getServers' procedure on server" ) );
+            assertThat( e.getMessage(), containsString( "Failed to run " +
+                    "'Statement{text='CALL dbms.cluster.routing.getServers', parameters={}}' on server." ) );
```
lutovich, line 153: Imho user-facing message with `Statement{` does not look pretty. Could we maybe concatenate a prettier message in `GetServersProcedureClusterCompositionProvider` instead of using default `Statement#toString()`?

driver/src/test/java/org/neo4j/driver/internal/cluster/GetServersProcedureRunnerTest.java
```diff
@@ -0,0 +1,112 @@
...
+        // Given
+        GetServersProcedureRunner runner = new TestGetServersProcedureRunner( value( (Object)null ) );
+        Connection mock = mock( Connection.class );
+        when( mock.server() ).thenReturn(
+                new InternalServerInfo( new BoltServerAddress( "123:45" ), "Neo4j/3.2.1" ) );
+        // When
+        runner.run( mock );
+
+        // Then
+        assertThat( runner.procedureCalled().toString(), equalTo(
```
lutovich: `Statement` as a good `#equals()` method defined. I think it should be used for assertions instead of `#toString()`.

lutovich (commented):
@zhenlineo review round completed. Made couple comments.

driver/src/test/java/org/neo4j/driver/internal/RoutingDriverTest.java
zhenlineo: Let's improve it in another PR, I will merge this PR once it is green.

PR #336, opened 2017-03-16
Move to boltkit

zhenlineo:


driver/src/test/java/org/neo4j/driver/v1/util/Neo4jRunner.java
```diff
@@ -42,15 +46,16 @@
 
     private static final boolean debug = Boolean.getBoolean( "neo4j.runner.debug" );
 
-    public static final String NEORUN_START_ARGS = System.getProperty( "neorun.start.args" );
+    private static final String DEFAULT_NEOCTRL_ARGS = "-e 3.1.2";
+    private static final String NEOCTRL_ARGS = System.getProperty( "neoctrl.args" ) == null?
```
lutovich: Method `System.getProperty(String key, String defaultValue)` can be used here

lutovich (commented):
@zhenlineo changes look good, made one minor comment. Also class `ClusterRule` uses boltkit and currently has Neo4j version hardcoded. It should be updated to use some neoctrl args system property.

PR #341, opened 2017-03-23
Reparent AuthEx

technige:
Reparent `AuthenticationException` under `ClientException` since this originates on the server, not locally. `SecurityException` remains in use for local TLS failures.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/v1/exceptions/AuthenticationException.java
+++ b/driver/src/main/java/org/neo4j/driver/v1/exceptions/AuthenticationException.java
@@ -25,7 +25,7 @@ package org.neo4j.driver.v1.exceptions;
  *
  * @since 1.1
  */
-public class AuthenticationException extends SecurityException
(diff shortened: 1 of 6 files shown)


driver/src/main/java/org/neo4j/driver/v1/exceptions/AuthenticationException.java
```diff
@@ -25,7 +25,7 @@
  *
  * @since 1.1
  */
-public class AuthenticationException extends SecurityException
+public class AuthenticationException extends ClientException
```
zhenlineo, line 28: I am concerned about changing this from SecurityException to ClientException directly, as it might break any users who already moved to 1.2.0/1.2.1 users.

Can we make it both ClientException and SecurityException for Java and any other language that is sensitive to Exception types?

zhenlineo:
I am killing this PR to settle down the plan that Auth exception is a Security exception today.

PR #342, opened 2017-03-24
Deprecated old routing retry settings

lutovich:
This PR deprecates `ConfigBuilder#withRoutingFailureLimit()` and `ConfigBuilder#withRoutingRetryDelay()` methods because they are now obsolete. They were initially introduced to postpone failure when no routers are available. Currently there exists a better way of doing this with transaction functions (`Session#readTransaction()` and `Session@writeTransaction()`). Driver also memorizes the seed URI and falls back to it during rediscovery when all other routers are dead.


zhenlineo:
Should this be merged to 1.2 or 1.3 branch?

lutovich:
@zhenlineo changed the base branch to 1.2

PR #344, opened 2017-03-27
Simplify retry logic and log when retry happens

lutovich:


driver/src/main/java/org/neo4j/driver/internal/net/pooling/BlockingPooledConnectionQueue.java
```diff
@@ -162,6 +162,6 @@ private void disposeSafely( PooledConnection connection )
 
     private static Logger createLogger( BoltServerAddress address, Logging logging )
     {
-        return logging.getLog( "connectionQueue[" + address + "]" );
+        return logging.getLog( "ConnectionQueue[" + address + "]" );
```
technige, line 165: Any reason this doesn't show the full class name?

technige (approved):
Looks good to me

driver/src/main/java/org/neo4j/driver/internal/net/pooling/BlockingPooledConnectionQueue.java
lutovich: Sorry, missed your comment. No real reason. I just aligned all logger names in this PR to start with a capital letter.

PR #345, opened 2017-03-27
Simplifies config for (not) using encryption.

technige:


driver/src/main/java/org/neo4j/driver/v1/Config.java
```diff
@@ -156,9 +156,18 @@ public int connectionTimeoutMillis()
...
     public EncryptionLevel encryptionLevel()
     {
-        return encryptionLevel;
+        return encrypted ? EncryptionLevel.REQUIRED : EncryptionLevel.NONE;
+    }
+
+    /**
+     * @return indicator for encrypted communication.
+     */
+    public boolean encrypted()
```
lutovich, line 168: What do you think about `isEncrypted` name for this method?

driver/src/main/java/org/neo4j/driver/v1/Config.java
```diff
@@ -328,9 +337,30 @@ public ConfigBuilder withConnectionLivenessCheckTimeout( long value, TimeUnit un
          * @param level the TLS level to use
          * @return this builder
          */
+        @Deprecated
```
lutovich, line 340: `EncryptionLevel` enum should probably also be deprecated

lutovich (commented):
@technige PR looks good, couple small questions

zhenlineo:
Interesting we already have `public boolean logLeakedSessions()`. 
So maybe we should keep `public boolean encrypted()` instead? CC @lutovich

lutovich:
@zhenlineo yeah, let's then keep `#encrypted()`

PR #346, opened 2017-03-29
Routing context in URI

lutovich:
Routing context is a map of string key-value pairs. It is send to the database when calling `dbms.cluster.routing.getRoutingTable` procedure as it's single argument. This is only supported by 3.2+ servers.

Previously routing context was defined on the config. It is problematic because routing context in `bolt+routing` specific and thus makes no sense for single instance database.

This PR makes routing context part of the `bolt+routing` URI. It lives in the query part of the URI as regular query parameters like:
`bolt+routing://localhost:7687/?policy=my_policy&region=china`.

Clustering-specific configuration method is removed.


zhenlineo:
LG, merging

PR #352, opened 2017-04-10
Unmute TCK tests for 3.2

MishaDemianenko:
Cypher profiler exception is fixed now and it is possible now to unleash TCK tests for 3.2


MishaDemianenko:
@zhenlineo can you please take a loo, thx!

PR #355, opened 2017-04-11
Read in absence of viable writer

lutovich:
Previously driver did not allow reads and writes when received routing table did not contains both routers, readers and writers. This was inconsistent with Causal Cluster which allows reads when leader is absent. Leader might be unavailable for a long time (when there is a DC failure, etc.) so it makes sense to allow clients to perform read activity.

This commit makes driver accept routing table with no writers and allow clients to perform read operations when writers are not available. It might be problematic when there is a cluster partition and one partition contains majority. For this case special care must be taken so driver does not get stuck talking only to the smaller partition which only knows about itself. This is done on best effort basis - driver tries to contact seed router if previously accepted routing table did not contain writes.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/cluster/ClusterRoutingTable.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/cluster/ClusterRoutingTable.java
@@ -25,6 +25,7 @@ import java.util.Set;
 
 import org.neo4j.driver.internal.net.BoltServerAddress;
 import org.neo4j.driver.internal.util.Clock;
+import org.neo4j.driver.v1.AccessMode;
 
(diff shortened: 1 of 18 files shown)


driver/src/main/java/org/neo4j/driver/internal/cluster/ClusterRoutingTable.java
```diff
@@ -115,7 +116,7 @@ public void removeWriter( BoltServerAddress toRemove )
 
 
     @Override
-    public String toString()
+    public synchronized String toString()
```
zhenlineo, line 119: Let's revisit this synchronized code in another PR.

PR #358, opened 2017-04-21
Added empty principal to fix kerberos AuthToken

eebus:


zhenlineo:
LG. Will merge after it is greed.

zhenlineo (approved):

Issue #359, opened 2017-04-21
memory leak in logging usage


if6was9:
There is a severe memory leak here:

https://github.com/neo4j/neo4j-java-driver/blob/1.3/driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java#L65

Our application ends up in GC hell after running for a few days.  Here is a chart from new relic that shows this:

![image](https://cloud.githubusercontent.com/assets/463742/25280281/428ad2f2-265e-11e7-999a-76f53d7f0ccb.png)

I'll submit a PR shortly.

if6was9:
https://github.com/neo4j/neo4j-java-driver/pull/360

lutovich:
Hello @if6was9,

Java driver 1.2.2 was released yesterday and is now available in maven central. Could you please try it out to confirm that this issue is resolved?

Thanks.

lutovich:
Hello @if6was9,

I'll close this issue assuming that fix in 1.2.2 and 1.3.0 works for you. Attached PR adds a logger count assertion to the stress test. Stress test was showing the same logger leak reported here.

Thanks for reporting this problem!

if6was9:
I believe the issue is resolved.  Thank you.

PR #362, opened 2017-04-25
Fix possible logger leak

lutovich:
Various components like session and connection include their identifiers in logs for diagnostic purposes. Most did this by creating loggers with specialized names. This is a problematic approach because it forces logging framework to create many distinct logger objects. JUL keeps such loggers as weakly referenced values in a map. This caused excessive GC activity and application failures.

This PR introduces a logger able to prepend diagnostics to messages and delegate rest of the work. It makes identifiers part of messages. Session, connection and other objects that require identification in logs now use this logger.

Example log message:
 `2017-04-25 02:36:47,319 [ConnectionQueue] [127.0.0.1] Error disposing connection`

Related to https://github.com/neo4j/neo4j-java-driver/pull/360
Fixes https://github.com/neo4j/neo4j-java-driver/issues/359

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
@@ -18,11 +18,10 @@
  */
 package org.neo4j.driver.internal;
 
-import java.util.ArrayList;
-import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


lutovich:
@if6was9 could you please take a look? I think this change allows us to keep session ids for diagnostics and should fix logger leak.

if6was9:
I think this will fix the GC issue.

PR #368, opened 2017-05-12
Byte array support

technige:
This PR adds client-side support for byte arrays using the PackStream BYTES type. Since this is only available in the 3.2 server and above, the code also detects when an attempt is made to write BYTES data to a connection to an earlier server version; in this case, a `ClientError` is raised with a useful error message.


zhenlineo:
Removed the back reference from packstream to connection in #370

PR #372, opened 2017-05-17
Construct routing driver from first available address

technige:


driver/src/main/java/org/neo4j/driver/v1/GraphDatabase.java
```diff
@@ -131,4 +133,44 @@ public static Driver driver( URI uri, AuthToken authToken, Config config )
...
+    /**
+     * Try to create a bolt+routing driver from the first available address.
+     * This is wrapper for the {@link #driver} method that finds the first
+     * server to respond positively.
+     *
+     * @param addresses a list of server addresses for Neo4j instances
+     * @param authToken authentication to use, see {@link AuthTokens}
+     * @return a new driver instance
+     */
+    public static Driver routingDriverFromFirstAvailableAddress( List<String> addresses, AuthToken authToken )
```
lutovich: Maybe it is ok to not add this method because we do not add another overload with just list of addresses?

driver/src/test/java/org/neo4j/driver/v1/integration/CausalClusteringIT.java
```diff
@@ -592,6 +616,24 @@ public Logger getLog( String name )
         return GraphDatabase.driver( boltUri, clusterRule.getDefaultAuthToken(), config );
     }
 
+    private Driver discoverDriver( List<String> addresses )
+    {
+        Logging devNullLogging = new Logging()
```
lutovich: `DevNullLogging.DEV_NULL_LOGGING` constant can be used instead

driver/src/main/java/org/neo4j/driver/v1/GraphDatabase.java
```diff
@@ -131,4 +133,44 @@ public static Driver driver( URI uri, AuthToken authToken, Config config )
 
         return new DriverFactory().newInstance( uri, authToken, routingSettings, retrySettings, config );
     }
+
+    /**
+     * Try to create a bolt+routing driver from the first available address.
+     * This is wrapper for the {@link #driver} method that finds the first
+     * server to respond positively.
+     *
+     * @param addresses a list of server addresses for Neo4j instances
```
lutovich: Could we mention that address is `host` or `host:port` without scheme?

lutovich (approved):
@technige changes look good. Should this go in 1.3 or 1.4?

technige:
@lutovich This will actually be for 1.4. Good point.

driver/src/main/java/org/neo4j/driver/v1/GraphDatabase.java
```diff
@@ -131,4 +133,44 @@ public static Driver driver( URI uri, AuthToken authToken, Config config )
...
+     * @param authToken authentication to use, see {@link AuthTokens}
+     * @param config user defined configuration
+     * @return a new driver instance
+     */
+    public static Driver routingDriverFromFirstAvailableAddress( List<String> addresses, AuthToken authToken, Config config )
```
zhenlineo: Can we do something like
```
public static Driver routingDriver( List<URI> uris, AuthToken authToken, Config config )
{
    for( URI uri: uris )
    {
        assertRoutingScheme( uri );
        try
        {
            return driver( uri, authToken, config );
        }
        catch( ServiceUnavailableException e )
        {
            // try the next one
        }
    }
    throw new ServiceUnavailableException( "Failed to discover an available server" );
}

private static void assertRoutingScheme( uri )
{
    if( !uri.getScheme().equals("bolt+routing") )
    {
        throw new IllegalArgumentException( "Unsupported scheme " + uri.getScheme() + " for routing driver. Supported scheme is 'bolt+routing'. " );
    }
}
```

driver/src/main/java/org/neo4j/driver/v1/GraphDatabase.java
zhenlineo: I would also vote for a single (long) method to start with.

zhenlineo (commented):
I would suggest to keep use URI (with scheme) as it might be confusing that sometimes we require scheme and sometimes we do not.

PR #373, opened 2017-05-18
Fail early on auth errors in direct driver

lutovich:
This PR makes direct driver test connectivity on creation. It's done by opening and closing a connection. Previously driver did not try to connect during creation and auth errors (incorrect credentials, etc.) popped up only when first query was executed or first transaction was started. Such delayed errors were confusing. Also routing driver behaved differently because it fetches routing table at creation and auth errors come out of constructor.


lutovich:
Hi @inserpio, this PR will make driver created with `bolt://` uri fail early when credentials are wrong.

inserpio:
Hi @lutovich, thank you very much for taking care about that. I will change the JDBC driver accordingly to benefit of that.
Cheers, Lorenzo

PR #375, opened 2017-05-18
Fix `Transaction#isOpen()` after errors

lutovich:
`ExplicitTransaction` has a FAILED state which is reached when fatal connection error happens or `Session#reset()` is called. In both cases corresponding transaction on server will be rolled back by the database. This PR makes sure FAILED state is changed to ROLLED_BACK when transaction is closed. It's needed to make sure `Transaction#isOpen()` does not report status incorrectly after closing of a FAILED transaction.


zhenlineo:
LG, pls merge and forward merge

PR #376, opened 2017-05-22
Multiple bookmarks

lutovich:
Previously it was possible to only supply a single bookmark when creating a new session. However there is a use-case to be able to supply multiple bookmarks when multiple worker threads execute write queries and then reader should be able to observe all those writes. To achieve this driver now exposed session creation methods that take an iterable of bookmarks.

It was chosen to add additional `Driver#session()` overloads with `Iterable<String>` instead of extending single-bookmark methods with varargs because:
 * this maintains backwards compatibility for pre-existing methods
 * addition of varargs would make API look strange because call or `Driver#session()` and call of `Driver#session(String... bookmarks)` without params is essentially the same thing
 * invocation of vararg methods with single `null` argument can be confusing
 * varargs are more suitable for manual params listing

Driver will now send:
```
{
    bookmark: "max",
    bookmarks: ["one", "two", "max"]
}
```
instead of simple:
```
{
    bookmark: "max"
}
```
this is done to maintain backwards compatibility with databases that only support single bookmark. It forces driver to parse and compare bookmarks which violates the fact that bookmarks are opaque. This is done only to maintain backwards compatibility and should not be copied. Code doing this will eventually be removed.

Related Bolt server PR: https://github.com/neo4j/neo4j/pull/9404

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/Bookmark.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/Bookmark.java
@@ -1,0 +1,154 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.neo4j.driver.internal;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+
+import org.neo4j.driver.v1.Value;
+
+import static java.util.Collections.emptyMap;
+import static java.util.Collections.singleton;
+import static org.neo4j.driver.v1.Values.value;
+
+public final class Bookmark
+{
+    private static final String BOOKMARK_KEY = "bookmark";
+    private static final String BOOKMARKS_KEY = "bookmarks";
+    private static final String BOOKMARK_PREFIX = "neo4j:bookmark:v1:tx";
+
+    private static final long UNKNOWN_BOOKMARK_VALUE = -1;
+
+    private static final Bookmark EMPTY = new Bookmark( Collections.<String>emptySet() );
+
+    private final Iterable<String> values;
+    private final String maxValue;
+
+    private Bookmark( Iterable<String> values )
+    {
+        this.values = values;
+        this.maxValue = maxBookmark( values );
+    }
+
+    public static Bookmark empty()
+    {
+        return EMPTY;
+    }
+
+    public static Bookmark from( String value )
+    {
+        if ( value == null )
+        {
+            return empty();
+        }
+        return from( singleton( value ) );
+    }
+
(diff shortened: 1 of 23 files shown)


driver/src/main/java/org/neo4j/driver/internal/Bookmark.java
```diff
@@ -0,0 +1,172 @@
...
+
+        Iterator<String> iterator = bookmarks.iterator();
+
+        String maxBookmark = iterator.next();
+        long maxValue = bookmarkValue( maxBookmark );
+
+        while ( iterator.hasNext() )
+        {
+            String bookmark = iterator.next();
+            long value = bookmarkValue( bookmark );
```
zhenlineo, line 127: value might be null for the following case:
bookmarks = new String[]{aBookmark, null, null};

driver/src/main/java/org/neo4j/driver/internal/InternalDriver.java
```diff
@@ -74,6 +75,23 @@ public final Session session( String bookmark )
     @Override
     public final Session session( AccessMode mode, String bookmark )
     {
+        return newSession( mode, Bookmark.from( bookmark ) );
+    }
+
+    @Override
+    public Session session( Collection<String> bookmarks )
```
zhenlineo: Five methods can be merged in to two
```
public Session session( String... bookmark )
public Session session( AccessMode mode, String... bookmark )
```

driver/src/main/java/org/neo4j/driver/internal/Bookmark.java
```diff
@@ -0,0 +1,172 @@
...
+    {
+        return values.isEmpty();
+    }
+
+    public String asString()
+    {
+        return maxValue;
+    }
+
+    public Map<String,Value> asParameters()
```
zhenlineo: maybe `asBeginTransactionParameters()`?

driver/src/main/java/org/neo4j/driver/internal/Bookmark.java
```diff
@@ -0,0 +1,172 @@
...
+        }
+        return new Bookmark( values );
+    }
+
+    public boolean isEmpty()
+    {
+        return values.isEmpty();
+    }
+
+    public String asString()
```
zhenlineo: This method name is confusing in a way that I would question why only return one bookmark rather than all of them back (like toString method).
Maybe change to names like `asSingleString()`, or `maxBookmarkAsString`?

driver/src/main/java/org/neo4j/driver/internal/Bookmark.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+        return new Bookmark( values );
+    }
+
+    public boolean isEmpty()
+    {
+        return maxValue == null;
+    }
+
+    public String asString()
```
zhenlineo: This method name is confusing in a way that I would question why only return one bookmark rather than all of them back (like toString method).
Maybe change to names like `asSingleString()`, or `maxBookmarkAsString`, or `singleBookmarkAsString`?

driver/src/test/java/org/neo4j/driver/internal/BookmarkTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+    @Test
+    public void bookmarkFromNullString()
+    {
+        Bookmark bookmark = Bookmark.from( (String) null );
+        assertTrue( bookmark.isEmpty() );
+    }
+
+    @Test
+    public void bookmarkFromIterable()
```
zhenlineo, line 98: What about a test for 
```
Bookmark bookmark = Bookmark.from( asList(
               null, "neo4j:bookmark:v1:tx10", "neo4j:bookmark:v1:tx12" ) );
```

lutovich: Added such test here: https://github.com/neo4j/neo4j-java-driver/pull/376/files#diff-0aefeb7eff1a4b6a3cdea429082587eeR144

zhenlineo:
LGTM

PR #381, opened 2017-06-20
char values should be encoded as strings

ali-ince:
Added overloads of pack and value methods with a char parameter to stop calls of this to be invoked on int overloads. Added corresponding tests.


lutovich (approved):
Changes look good. Maybe we could also add `Values#value(char... input)`?

PR #382, opened 2017-06-21
Do not cache resolved inet address

lutovich:
Internally driver always uses `BoltServerAddress` to represent host and port. These addresses are constructed from the initial driver URI and from routing procedure output in routing driver. Bolt address is only resolved to an `InetAddress` when new connection is established.

Resolved `InetSocketAddress` was cached in `BoltServerAddress` which caused problems when corresponding DNS record changed. Driver would then still try to connect to a cached stale `InetSocketAddress`.

This PR makes resolution happen on every connection.


zhenlineo:
shall we fix it in 1.3/1.2?

lutovich:
@zhenlineo I thought about fixing this only in 1.4 and doing 1.4.1 release to avoid touching previous branches.

PR #384, opened 2017-07-04
Decouple load balancing logic from address set

lutovich:
Round-robin load balancing was previously coupled with the address set implementation and made it hard to use different load balancing algorithm.

This PR turns `RoundRobinAddressSet` into a simple concurrent set of addresses and moves load balancing logic into a separate component that takes list of available addresses and returns one that should be used next. This allows easier implementation of new load balancing algorithms. Rediscovery procedure will now not use load balancing and query routers in a known order (which is randomized by the database).


ali-ince (approved):
Changes look good. Very minor and not related with this change, it seems that the synchronized keywords in ClusterRoutingTable#update and ClusterRoutingTable#forget can be removed as those are called through LoadBalancer#refreshRouting's logic.

PR #386, opened 2017-07-06
Trace logging around load balancing and connection pool

lutovich:
Based on https://github.com/neo4j/neo4j-java-driver/pull/385


ali-ince (approved):
Shall we also change the error log message in the BlockingPooledConnectionQueue#disposeSafely method to include it has happened during queue termination? We may even re-consider its log level.

lutovich:
@ali-ince log message improved and changed level from error to warn.

PR #387, opened 2017-07-07
Improve disposal of broken connections

lutovich:
Connection pool can discard broken connections during acquisition (when connection liveness check timeout is configured) and when connections are returned to the pool. In both cases connections should be disposed and removed from the set of active connections. This is especially important with least connected load balancing strategy which examines amount of active connections for each address.

This PR makes sure broken connections are disposed through the connections queue to make sure active set is always updated.

Based on https://github.com/neo4j/neo4j-java-driver/pull/386, only last commit is new.


driver/src/main/java/org/neo4j/driver/internal/DriverFactory.java
```diff
@@ -146,7 +149,21 @@ protected InternalDriver createDriver( Config config, SecurityPlan securityPlan,
     protected LoadBalancer createLoadBalancer( BoltServerAddress address, ConnectionPool connectionPool, Config config,
             RoutingSettings routingSettings )
     {
-        return new LoadBalancer( address, routingSettings, connectionPool, createClock(), config.logging() );
+        return new LoadBalancer( address, routingSettings, connectionPool, createClock(), config.logging(),
+                createLoadBalancingStrategy( config, connectionPool ) );
+    }
+
+    private LoadBalancingStrategy createLoadBalancingStrategy( Config config, ConnectionPool connectionPool )
```
zhenlineo, line 156: Method can be static

driver/src/main/java/org/neo4j/driver/v1/Logger.java
```diff
@@ -25,29 +25,41 @@
...
      */
     void warn( String message, Object... params );
 
     /**
+     * Logs warnings that happened during using the driver
+     *
+     * @param message the warning message
+     * @param cause the cause of the warning
+     */
+    void warn( String message, Throwable cause );
```
zhenlineo, line 56: This is a API change that will probably break user's code. Just saying...

zhenlineo (commented):
Looks good.

PR #389, opened 2017-07-12
fix for failed to enlarge network buffer error

ali-ince:
changed TLSSocketChannel#wrap to agressively do channelWrite upon BUFFER_OVERFLOW. Previously it was failing fast when nothing gets written to the underlying channel which can actually happen on the socket level.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/security/TLSSocketChannel.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/security/TLSSocketChannel.java
@@ -345,18 +345,22 @@ public class TLSSocketChannel implements ByteChannel
             }
             else
             {
+                logger.debug( "Network output buffer doesn't need enlarging, flushing data to the channel instead to open up space on the buffer." );
                 // flush as much data as possible
                 cipherOut.flip();
-                if ( channelWrite( cipherOut ) == 0 )
-                {
-                    throw new ClientException( format(
-                                    "Failed to enlarge network buffer from %s to %s. This is either because the " +
-                                    "new size is however less than the old size, or because the application " +
-                                    "buffer size %s is so big that the application data still cannot fit into the " +
-                                    "new network buffer.", curNetSize, netSize, buffer.capacity() ) );
+                while ( cipherOut.hasRemaining() ) {
+                    int written = channelWrite( cipherOut );
+
+                    if (written > 0) {
+                        break;
+                    }
+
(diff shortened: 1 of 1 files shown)


zhenlineo:
@ali-ince I retarget this at 1.4 now based on the discussion we had this morning.

driver/src/main/java/org/neo4j/driver/internal/security/TLSSocketChannel.java
```diff
@@ -345,18 +345,22 @@ private HandshakeStatus wrap( ByteBuffer buffer ) throws IOException, ClientExce
...
                 // flush as much data as possible
                 cipherOut.flip();
-                if ( channelWrite( cipherOut ) == 0 )
-                {
-                    throw new ClientException( format(
-                                    "Failed to enlarge network buffer from %s to %s. This is either because the " +
-                                    "new size is however less than the old size, or because the application " +
-                                    "buffer size %s is so big that the application data still cannot fit into the " +
-                                    "new network buffer.", curNetSize, netSize, buffer.capacity() ) );
+                while ( cipherOut.hasRemaining() ) {
```
lutovich, line 351: Would it make sense to limit write attempts here? Could execution get stuck here when unable to write?

lutovich (approved):
@ali-ince change looks good. One small question.

driver/src/main/java/org/neo4j/driver/internal/security/TLSSocketChannel.java
ali-ince: Yeah, I also thought on adding a limit but;

If we end up in a position where we could not write anything on a socket, it most probably points a resource starvation on the computer either cpu or network buffers related. In either case the problem lies somewhere else and if we throw some exception here - where theoretically it is not an error - we would break the runtime and eventually become a point of failure.

We also loop on the same condition at [line](https://github.com/neo4j/neo4j-java-driver/pull/389/files/a78e441dc69486fc5b7649f1d11e396b5ace3d9f#diff-4bff1912d58fd3cac510c55ab49334e1R329) on the same method, so I thought this would also not be a problem.

Does it make sense?

zhenlineo: I found this argument is fair.

lutovich: Makes sense, and if socket is completely broken exception will be thrown by `#channelWrite ()` method.

PR #398, opened 2017-08-10
Added ability to enforce max connection lifetime

lutovich:
Driver keeps idle socket connections in a connection pool. These connections can get invalidated while resting idle in the pool. They can be killed by network equipment like load balancer, proxy or firewall. It is also safer to refresh connections once in a while so that database can renew corresponding resources.

This PR adds `maxConnectionLifetime` setting and makes driver close too old connections. Checking and closing happens in an application thread that tries to acquire a connection. Feature can be used on it's own or in combination with `connectionLivenessCheckTimeout` to guarantee validity of acquired connections.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/net/pooling/PoolSettings.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/net/pooling/PoolSettings.java
@@ -21,17 +21,21 @@ package org.neo4j.driver.internal.net.pooling;
 public class PoolSettings
 {
     public static final int NO_IDLE_CONNECTION_TEST = -1;
+    public static final int INFINITE_CONNECTION_LIFETIME = -1;
 
     public static final int DEFAULT_MAX_IDLE_CONNECTION_POOL_SIZE = 10;
     public static final int DEFAULT_IDLE_TIME_BEFORE_CONNECTION_TEST = NO_IDLE_CONNECTION_TEST;
+    public static final int DEFAULT_MAX_CONNECTION_LIFETIME = INFINITE_CONNECTION_LIFETIME;
 
     private final int maxIdleConnectionPoolSize;
     private final long idleTimeBeforeConnectionTest;
+    private final long maxConnectionLifetime;
 
-    public PoolSettings( int maxIdleConnectionPoolSize, long idleTimeBeforeConnectionTest )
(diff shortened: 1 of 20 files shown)


driver/src/main/java/org/neo4j/driver/internal/net/pooling/PooledSocketConnection.java
```diff
@@ -281,6 +284,12 @@ public void setResourcesHandler( SessionResourcesHandler resourcesHandler )
     }
 
     @Override
+    public long creationTimestamp()
```
zhenlineo, line 287: Do we need to expose `creationTimestamp` and `lastUsedTimestamp` if we have a `isValid` method?

```
when obtaining a conn from the pool
do
{
    conn = pool.dequeue()
} while(!conn.isValid()) // where we calculate timestamp
```

driver/src/main/java/org/neo4j/driver/internal/net/pooling/PoolSettings.java
```diff
@@ -53,4 +57,19 @@ public boolean idleTimeBeforeConnectionTestConfigured()
...
+    {
+        if ( !maxConnectionLifetimeConfigured() )
+        {
+            throw new IllegalStateException(
+                    "Max connection lifetime is not configured: " + maxConnectionLifetime );
+        }
+        return maxConnectionLifetime;
+    }
+
+    public boolean maxConnectionLifetimeConfigured()
```
zhenlineo, line 71: It might be okay to assume -1 or negative indicate `infinite` or `forever`, rather than not configured.

ali-ince (approved):
Changes look fine for me.

driver/src/main/java/org/neo4j/driver/internal/net/pooling/PooledSocketConnection.java
lutovich: I think addition of `PooledConnection#isValid()` will make connection too smart. It'll basically have to be aware of pool settings, do ping and stuff. I guess we could move this into existing `ConnectionValidator` interface. Maybe as a separate PR?

driver/src/main/java/org/neo4j/driver/internal/net/pooling/PoolSettings.java
lutovich: Yeah, that's basically what it means in the code. Pool treats negative and zero as if connections have infinite lifetime. Do you think this method should reflect this in it's name?

PR #399, opened 2017-08-18
removed exceptions for pool settings.

praveenag:


zhenlineo:
LG

lutovich (changes requested):
@praveenag those exceptions are only internal. They prevent our code from using `idleTimeBeforeConnection` and `maxConnectionLifetime` when they are not configured by the user. Callers are expected to use `PoolSettings` like this:

```java
if ( poolSettings.idleTimeBeforeConnectionTestConfigured() )
{
    long timeout = poolSettings.idleTimeBeforeConnectionTest();
    testConnectionIfNeeded(connection, timeout);
}
```

This is similar to `Iterator#next()` which might throw `NoSuchElementException` when not guarded by `Iterator#hasNext()`.

Change in `ConnectionHandlingIT` is good.

zhenlineo:
Suggest to rename the method in poolSettings:
idleTimeBeforeConnectionTestConfigured -> idleTimeBeforeConnectionTestEnabled
as it is not a test for whether a user has configured this config option or not, but whether we should do a test or not given the timeout in the config.

Also suggest to change the exception message from `Idle time before connection test is not configured: ` to `Idle time before connection test is not enabled ... Contact the driver author if this error message is seen.`

praveenag:
@lutovich How about now?
@zhenlineo I have removed the exceptions but renamed methods.

lutovich (approved):

zhenlineo:
LG

PR #400, opened 2017-08-25
Log failures in `GraphDatabase#routingDriver()`

lutovich:
This method creates routing driver using the first address that responds positively from the given list of addresses. It previously ignored failures which made it harder to diagnose connectivity problems. Such failures will now be logged using configured logging.


ali-ince (approved):
Changes look good.

Issue #102, opened 2015-12-19
Async/Future based .run()


theduke:
I know this is still a WIP,

but are you planning on adding an async version of .run() which returns a Future?

jakewins:
Yes - although we've discussed adding a Reactive Streams based API, rather than using a future, so it'd be similar to how the javascript driver handles async results. You could build a Future-based API on top of that easily tho.

The underlying protocol, Bolt, is asynchronous, so this is just a matter of exposing that in this driver.

Out of curiosity - what's your use case?

theduke:
I'm using Neo4j with Akka (http://akka.io/), where you are supposed to block as little as possible.

JacobASeverson:
+1 :-)

lpfeup:
I would also be interested in this. Also using Neo4j with Akka.

jn73:
Any update on this? I'm working on a akka http service and this would help a lot.

PhilAndrew:
+1 async is really a requirement for performance.  holding onto many threads with one thread per request to the database is analogous to why we now use non blocking web server's.  

For people in the Scala world we really never ever do blocking IO, its considered a horrible sin, so its difficult to use this within the Scala world. Java people also try not to do blocking IO.

jellybams:
+1 for use with akka-http + akka-streams

dazraf:
This would be an invaluable API. With toolkits like Akka, Vert.x and others, we are finding that we can get significantly greater performance with asynchronous APIs. _Please Note:_ the API should not be Future based because, like most database APIs, the key is asynchronous processing of result _streams_ not result _sets_. Please consider implementing this API using something like RxJava Observables; or equivalent. Thanks.

PhilAndrew:
Agree Akka Streams would be good as it works well in Java and Scala world.

technige:
This is now on our roadmap.

vicaba:
@technige where can i track the async api development?

lutovich:
Hello everyone!

We've made some progress on the async API and here is an initial PR: https://github.com/neo4j/neo4j-java-driver/pull/402.
Reviews, opinions and feedback are extremely welcome!

It will take us a bit more to release draft of async API in an alpha release. You are, of course, very welcome to build and try it out before that :)

lutovich:
Hello again,

[...]

PR #402, opened 2017-09-15
Draft of async API

lutovich:
PR adds initial version of async API for direct (only `bolt://` scheme, not `bolt+routing://`) driver.

Main API contact points:
 * `Response` interface that is a `Future` one can subscribe to with `ResponseListener`
 * `StatementResultCursor` which is an async counterpart of `StatementResult`; it is basically an async iterator with `#hasNext()` being `Response<Boolean> fetchAsync()` and `#next()` being `Record current()`
 * various overloads or `Session#runAsync()` to run queries asynchronously in auto-commit mode
 * `Session#beginTransactionAsync()` to start a transaction asynchronously
 * various overloads or `Transaction#runAsync()` to run queries asynchronously in the transaction
 * `Transaction#commitAsync()` and `Transaction#rollbackAsync()` to finish transaction with commit or rollback
 * `Session#closeAsync()` to close session asynchronously and release all resources

Implementation notes:
 * Netty is used internally for async IO; it is now specified as a dependency with scope `compile` but will eventually be shaded into the driver artifact
 * All internal futures & promises are built on top of corresponding Netty classes; all callbacks and future transformations execute in Netty event loop threads
 * Channels are pooled in a `FixedChannelPool` which adds two new config settings: `maxConnectionPoolSize` and `connectionAcquisitionTimeoutMillis`
 * Javadocs are missing right now
 * Packages will need some reshuffling
 * Simple back-pressure is available using channel auto-read functionality but it's turned off by default
 * Async API lives in parallel to sync API and uses separate network connections, pools and everything; this is temporary and eventually sync will be implemented on top of async

Not part of this PR:
 * Transaction functions with retries
 * Routing driver
 * Netty logging does not use driver logging
 * Protocol level back-pressure
 * Good test coverage

Related to https://github.com/neo4j/neo4j-java-driver/issues/102


zhenlineo:
LG, let's merge!

BTW, Should we merge this into a separate branch? or should we merge into main branch?

driver/src/main/java/org/neo4j/driver/internal/ConnectionSettings.java
```diff
@@ -29,7 +29,7 @@
  */
 public class ConnectionSettings
 {
-    private static final String DEFAULT_USER_AGENT = format( "neo4j-java/%s", driverVersion() );
+    public static final String DEFAULT_USER_AGENT = format( "neo4j-java/%s", driverVersion() );
```
ali-ince: Very small thing but is this being used elsewhere? Shall we set it as private again?

driver/src/main/java/org/neo4j/driver/internal/ExplicitTransaction.java
```diff
@@ -57,28 +70,68 @@
...
+        this.asyncConnection = asyncConnection;
+        this.session = session;
+        this.resourcesHandler = SessionResourcesHandler.NO_OP;
+    }
+
+    public void begin( Bookmark initialBookmark )
+    {
+        Map<String,Value> parameters = initialBookmark.asBeginTransactionParameters();
+
+        connection.run( "BEGIN", parameters, NoOpResponseHandler.INSTANCE );
```
ali-ince: Do these "BEGIN", "COMMIT" and "ROLLBACK" strings  deserve their constants?

ali-ince (approved):
Looks really good! Only a couple of comments (even can be ignored :)).

lutovich:
We decided to keep async API in a separate branch for now. It will get merged into 1.5 a bit later.
I'll close this PR now. Please track progress in https://github.com/neo4j/neo4j-java-driver/tree/1.5-async.

PR #404, opened 2017-09-20
Async transaction functions

lutovich:
Added support for transaction functions that return futures. Cleaned up future helper methods.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
@@ -18,7 +18,9 @@
  */
 package org.neo4j.driver.internal;
 
-import io.netty.util.concurrent.GlobalEventExecutor;
+import io.netty.util.concurrent.EventExecutorGroup;
+import io.netty.util.concurrent.Future;
+import io.netty.util.concurrent.FutureListener;
 
 import java.util.Map;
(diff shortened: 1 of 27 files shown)


driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
```diff
@@ -240,7 +245,7 @@ public void close()
     {
         if ( asyncConnectionFuture != null )
         {
-            return asyncConnectionFuture.thenCombine( new Function<AsyncConnection,InternalFuture<Void>>()
+            return asyncConnectionFuture.thenCompose( new Function<AsyncConnection,InternalFuture<Void>>()
```
ali-ince, line 248: Can both `asyncConnectionFuture` and  `currentAsyncTransactionFuture` be not null?

ali-ince (approved):
Looks good to me.

driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
lutovich: They can, yeah. I see your point, basically if we close session it will first force-release connection (with RESET message) and then try to close active transaction. Later, however, is not really needed after RESET. I think `else if ( currentAsyncTransactionFuture != null )` branch here is not really required. Will think a bit more about this and probably remove it in a separate PR.

PR #405, opened 2017-09-21
Improvements for StatementResultCursor API

lutovich:
Based on https://github.com/neo4j/neo4j-java-driver/pull/404.

Main changes:
 * replaced `#fetchAsync()` + `#current()` with single `#nextAsync()` to simplify handling of futures
 * added `#forEachAsync()`
 * added `#listAsync()`

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/DriverFactory.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/DriverFactory.java
@@ -19,6 +19,7 @@
 package org.neo4j.driver.internal;
 
 import io.netty.bootstrap.Bootstrap;
+import io.netty.util.concurrent.EventExecutorGroup;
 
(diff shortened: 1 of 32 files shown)


ali-ince (approved):
LGTM!

zhenlineo:
A few commits:
1. How do we expect a user to use `cursor.nextAsync()`? It seems to me that a user should never use `cursor.nextAsync()`, instead they should probably use `#singleAsync()` (missing), `#forEachAsync`, or `#listAsync`.
2. If we replaced `#fetchAsync() + #current()` with single `#nextAsync()`, then maybe we should go over with .NET and JS API to see if we should uniform or differ there?
3. I thought `#forEachAsync` would return `summary` :P
4. Shall we also have `public Response<List<Record>> listAsync(final Consumer<Record> action)`?

PR #406, opened 2017-09-22
Java 8, CompletionStage in async and stress test improvements

lutovich:
This PR bumps required java version to 8 so that we can use `CompletionStage` and `CompletableFuture` in async API.

Stress test has been split into a CC and single instance test. It now also tests async API.


driver/src/main/java/org/neo4j/driver/internal/handlers/PullAllResponseHandler.java
```diff
@@ -102,56 +104,56 @@ public synchronized void onRecord( Value[] fields )
...
+            completeRecordFuture( record );
         }
         else
         {
             queueRecord( record );
         }
     }
 
-    public synchronized InternalFuture<Record> nextAsync()
+    public synchronized CompletionStage<Record> nextAsync()
```
ali-ince, line 117: I think we can improve concurrent performance among `#onRecord()` and `#nextAsync()` methods by avoiding method level  `synchronized` access. We will have to keep an eye on `recordFuture` and queuing records (we can maybe use `ConcurrentLinkedQueue` instead of `Queue`).

PR #408, opened 2017-09-27
Perform writes fully in event loop

lutovich:
`NettyConnection` is the abstraction over netty channel that allows us to write RUN, PULL_ALL and other messages. Currently write consists of two steps: (1) enqueue a response handler (2) perform actual write. Step (1) is executed in the caller thread and step (2) is always executed in the event loop thread (guaranteed by Netty).

Step (1) involves modification of a non-synchronized `LinkedList` which is a problem. Later event loop thread will try to pop handlers from the same list meaning we access it from different threads.

This PR make both (1) and (2) execute in the event loop to avoid synchronization issues. It basically makes sure `InboundMessageDispatcher` is only accessed by the event loop.

Also removed `ConcurrentSet` implementation in favour of the one existing in Netty.


ali-ince (approved):
LGTM!

PR #409, opened 2017-09-28
Initial support for async in routing driver

lutovich:


driver/src/main/java/org/neo4j/driver/internal/cluster/loadbalancing/LoadBalancingStrategy.java
```diff
@@ -33,11 +33,15 @@
      */
     BoltServerAddress selectReader( BoltServerAddress[] knownReaders );
 
+    BoltServerAddress selectReaderAsync( BoltServerAddress[] knownReaders );
```
ali-ince, line 36: and here

driver/src/main/java/org/neo4j/driver/internal/cluster/loadbalancing/LoadBalancer.java
```diff
@@ -193,6 +286,21 @@ private BoltServerAddress selectAddress( AccessMode mode, AddressSet servers )
         }
     }
 
+    private BoltServerAddress selectAddressAsync( AccessMode mode, AddressSet servers )
```
ali-ince, line 289: Do we need this sync/async distinction at this level? Do you have plans to implement selection differently in the future?

driver/src/main/java/org/neo4j/driver/internal/cluster/loadbalancing/RoundRobinLoadBalancingStrategy.java
```diff
@@ -47,11 +47,23 @@ public BoltServerAddress selectReader( BoltServerAddress[] knownReaders )
     }
 
     @Override
+    public BoltServerAddress selectReaderAsync( BoltServerAddress[] knownReaders )
```
ali-ince, line 50: and here

driver/src/main/java/org/neo4j/driver/internal/cluster/loadbalancing/LoadBalancer.java
```diff
@@ -44,56 +48,83 @@
...
     {
         this.connections = connections;
+        this.asyncConnectionPool = asyncConnectionPool;
         this.routingTable = routingTable;
         this.rediscovery = rediscovery;
         this.loadBalancingStrategy = loadBalancingStrategy;
         this.log = log;
 
-        refreshRoutingTable();
+        if ( connections != null )
```
ali-ince, line 95: does this depend only on `connections`? Or should `asyncConnectionPool` has a say in the decision?

ali-ince (approved):
LGTM! I added a couple of comments. Besides, having both `ConnectionPool` and `AsyncConnectionPool` in the code base started to be confusing for me, maybe it is time for a clean-up if those old pieces are not needed anymore.

lutovich:
@ali-ince thanks for the review! I planned to get rid of those `*Async` suffixes and pool duality in the next step when doing blocking on top of async.

PR #410, opened 2017-09-28
More helper methods in StatementResultCursor

lutovich:
This PR adds following methods to `StatementResultCursor`:

 * `CompletionStage<Record> singleAsync()` to safely retrieve single record from the cursor and assert that cursor contains exactly one record
 * `CompletionStage<List<T>> listAsync(Function<Record, T>)` to transform all incoming records and store them in list
 * `CompletionStage<ResultSummary> consumeAsync()` to drop/skip all incoming records and get result summary

They are added for convenience and to be consistent with existing blocking API in `StatementResult`.

Also `#forEachAsync(Consumer<Record>)` now returns future of `ResultSummary`.


zhenlineo (approved):

ali-ince (approved):
LGTM!

PR #411, opened 2017-10-02
Async routing improvements

lutovich:
PR makes async routing respect `Config#withRoutingRetryDelay()` setting and retry rediscovery configured amount of times with this delay. Retries are scheduled in Netty even loop. Also it makes load balancer "loop" until a working connection is acquired instead of failing on first failed acquisition attempt.

Based on https://github.com/neo4j/neo4j-java-driver/pull/409


ali-ince (approved):
LGTM.

PR #413, opened 2017-10-03
Make sure network channel is closed on handshake or INIT failure

lutovich:
Channel connection executed by `AsyncConnectorImpl` is a three step process:
 1) connect netty channel
 2) perform bolt handshake
 3) send INIT message

When either of these steps fail we should close (possibly open) network channel from step (1) and not track this channel in the pool or in the set of active channels.

This PR makes sure channels are closed and pool does not account for connections that failed during step (2) or (3).


zhenlineo:
LG

ali-ince (approved):
LGTM!

PR #424, opened 2017-10-26
Relax connection termination policy in routing driver

lutovich:
Previously routing driver terminated all connections towards a particular address when one of active connections had a network error. Connections were also terminated when new routing table did not contain some address that was present in the previous routing table. Such behaviour might be problematic because it results in terminated queries. Network errors might have been temporary but always resulted in termination of active queries.

This commit makes driver keep existing connections and terminate only idle ones on network error. Address will still be excluded from the routing table and subsequent queries will not be pointed on it. Corresponding connection pool is transferred into a "passive" state where it does not serve new connections and disposes returned connections immediately. Pool is transferred back to "active" state during next rediscovery only if corresponding address is present in the new routing table.

Old behaviour still remains and can be activated using `-DpurgeOnError=true` system property. It exists as a fallback only for exceptional cases and will most likely be silently removed in future.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/cluster/ClusterRoutingTable.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/cluster/ClusterRoutingTable.java
@@ -78,16 +78,17 @@ public class ClusterRoutingTable implements RoutingTable
     public synchronized RoutingTableChange update( ClusterComposition cluster )
     {
         expirationTimeout = cluster.expirationTimestamp();
-        Set<BoltServerAddress> pre = servers();
+        Set<BoltServerAddress> previousServers = servers();
+
         readers.update( cluster.readers() );
         writers.update( cluster.writers() );
         routers.update( cluster.routers() );
-        Set<BoltServerAddress> cur = servers();
+        Set<BoltServerAddress> currentServers = servers();
 
-        Set<BoltServerAddress> added = new HashSet<>( cur );
-        Set<BoltServerAddress> removed = new HashSet<>( pre );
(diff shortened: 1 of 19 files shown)


driver/src/main/java/org/neo4j/driver/internal/cluster/ClusterRoutingTable.java
```diff
@@ -66,14 +67,16 @@ public boolean isStaleFor( AccessMode mode )
     }
 
     @Override
-    public synchronized Set<BoltServerAddress> update( ClusterComposition cluster )
+    public synchronized RoutingTableChange update( ClusterComposition cluster )
     {
         expirationTimeout = cluster.expirationTimestamp();
+        // todo: what if server is added as reader and removed as writer? we should not treat it as removed
```
zhenlineo: To answer your todo, there is no need to know if the status of an address in `RoutingTable`, we only care about the whole distinct addresses in the `RoutingTable`. The `ConnectionPools` only need to know if an address is `active` (can reuse connections) or `passive` (can be terminated once no inUse connections in the pool)

See more explanation in https://github.com/lutovich/neo4j-java-driver/pull/1

zhenlineo (commented):
I got you this PR https://github.com/lutovich/neo4j-java-driver/pull/1 to simplify some logic

driver/src/main/java/org/neo4j/driver/internal/cluster/ClusterRoutingTable.java
lutovich: Looks good!

PR #426, opened 2017-11-13
Use AccessMode correctly in NetworkSession

lutovich:
`NetworkSession` acquires connections from the connection pool for the given access mode. Previously it tried to reuse existing connection when available. However, it did not pay attention to access mode when reusing connections. This could result in READ connection being used for WRITE operation and vice versa.

This PR fixes the problem by making session buffer existing result, if any. Buffering means that in-use connection will be released back to the pool. It also allows session propagate unconsumed failures from previous query executions.


zhenlineo:
LG

PR #428, opened 2017-11-13
Improve default config values for connection pool

lutovich:
PR changes default values from "unlimited" to lower and more realistic values. Modified settings are:

  * `maxConnectionLifetime` - changed from unlimited to 1 hour because it seems
    reasonable to periodically "refresh" connections and make sure they are
    not affected by network infrastructure.

  * `maxConnectionPoolSize` - changed from `Integer.MAX_VALUE` to 100 because
    unlimited connection pool is dangerous and might result in excessive use
    of resources and OOMs in the database, if connections there are
    not limited.

  * `connectionAcquisitionTimeout` - changed from unlimited to 60 seconds
    to not cause infinite blocking.


zhenlineo:
+1

PR #429, opened 2017-11-15
Fixed couple netty buffer leaks

lutovich:
Multiple leaks fixed in tests, one leak fixed in production code in `OutboundMessageHandler`.


driver/src/main/java/org/neo4j/driver/internal/async/outbound/OutboundMessageHandler.java
```diff
@@ -65,14 +65,14 @@ protected void encode( ChannelHandlerContext ctx, Message msg, List<Object> out
...
         }
         catch ( Throwable error )
         {
-            throw new EncoderException( "Failed to write outbound message: " + msg, error );
-        }
-        finally
-        {
             output.stop();
+            // release buffer because it will not get added to the out list and no other handler is going to handle it
+            messageBuf.release();
```
zhenlineo, line 74: 👍

zhenlineo (approved):

PR #430, opened 2017-11-15
Fixed handling of empty byte arrays

lutovich:
Previously decoding of empty byte arrays caused an error because pack stream layer forced pack input to fill in an empty buffer. Later tried to fetch next chunk in such case and still fill the given buffer with nothing. This caused it to fail later on message boundary, when asserting that the whole chunk has been consumed.

This PR fixes the problem by making pack input do nothing when asked to fill in an empty buffer. It also adds a branch in pack stream to always return a single empty byte array and not even attempt to read.


zhenlineo (approved):
LG, suggesting to also add tests for other drivers.

ali-ince (approved):
LGTM!

PR #431, opened 2017-11-20
Improve logging in channel pipeline

lutovich:
Made all handlers in the channel pipeline prefix their log messages with channel info. Added tests and fixed issue where code would hang if channel gets disconnected during bolt handshake. Improved layout of java packages.


driver/src/main/java/org/neo4j/driver/internal/async/inbound/InboundMessageDispatcher.java
```diff
@@ -120,18 +121,20 @@ public void handleSuccessMessage( Map<String,Value> meta )
     @Override
     public void handleRecordMessage( Value[] fields )
     {
+        ResponseHandler handler = handlers.peek();
         if ( log.isDebugEnabled() )
         {
-            log.debug( "Received RECORD message with metadata %s", Arrays.toString( fields ) );
+            log.debug( "Received RECORD message with metadata %s for handler %s", Arrays.toString( fields ), handler );
```
lutovich: Print all incoming messages as in `LoggingResponseHandler` in 1.4

driver/src/main/java/org/neo4j/driver/internal/async/inbound/MessageDecoder.java
```diff
@@ -24,18 +24,59 @@
...
         if ( msg instanceof ByteBuf )
         {
+            ByteBuf chunkBuf = (ByteBuf) msg;
+
             // on every read check if input buffer is empty or not
             // if it is empty then it's a message boundary and full message is in the buffer
-            readMessageBoundary = ((ByteBuf) msg).readableBytes() == 0;
+            readMessageBoundary = chunkBuf.readableBytes() == 0;
+
+            if ( log.isTraceEnabled() )
```
lutovich: Consider logging chunk header and message boundary so that we can reconstruct messages from hex when debugging.

driver/src/main/java/org/neo4j/driver/internal/cluster/Rediscovery.java
```diff
@@ -232,9 +232,7 @@ else if ( composition != null )
             }
             else
             {
-                ClusterComposition cluster = response.clusterComposition();
-                logger.info( "Got cluster composition %s", cluster );
-                return cluster;
+                return response.clusterComposition();
```
lutovich, line 235: Consider logging records received from routing procedure on debug level

lutovich:
@zhenlineo comments should now be fixed

driver/src/main/java/org/neo4j/driver/internal/async/outbound/OutboundMessageHandler.java
```diff
@@ -77,7 +92,7 @@ protected void encode( ChannelHandlerContext ctx, Message msg, List<Object> out
 
         if ( log.isTraceEnabled() )
         {
-            log.trace( "Message %s encoded as\n%s\n", msg, prettyHexDump( messageBuf ) );
+            log.trace( "C: %s encoded as\n%s", msg, prettyHexDump( messageBuf ) );
```
zhenlineo: Shall we change this to "C:\n%s" so that
* it is the same as `log.trace( "S:\n%s", prettyHexDump( msg ) );`
* to avoid printing message twice as it already printed out in debug.

driver/src/main/java/org/neo4j/driver/internal/cluster/Rediscovery.java
zhenlineo: If debug level is enabled, then the routing table will be printed out in the record message coming from getRoutingTable procedure. Therefore there might be no need to log again in debug level.

Instead we might want a logger at `INFO` level when there is a new routing table if we only have INFO level on.

driver/src/main/java/org/neo4j/driver/internal/cluster/RoutingProcedureClusterCompositionProvider.java
```diff
@@ -73,7 +73,11 @@ private ClusterCompositionResponse processRoutingResponse( RoutingProcedureRespo
 
         List<Record> records = response.records();
 
-        log.info( "Got getServers response: %s", records );
+        if ( log.isDebugEnabled() )
+        {
+            log.debug( "Received response from %s procedure: %s", invokedProcedureString( response ), records );
```
zhenlineo: sugget to change to INFO

PR #432, opened 2017-11-20
Deprecated 'idleConnectionPoolSize' config option

lutovich:
Newly added 'maxConnectionPoolSize' option should be used instead.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/v1/Config.java
+++ b/driver/src/main/java/org/neo4j/driver/v1/Config.java
@@ -335,7 +335,8 @@ public class Config
          *
          * @param size the max number of sessions to keep open
          * @return this builder
-         * @deprecated please use {@link #withMaxConnectionPoolSize(int)} instead.
+         * @deprecated please use a combination of {@link #withMaxConnectionPoolSize(int)} and
+         * {@link #withConnectionAcquisitionTimeout(long, TimeUnit)} instead.
          */
         @Deprecated
         public ConfigBuilder withMaxSessions( int size )
@@ -352,7 +353,8 @@ public class Config
          *
          * @param size the max number of idle sessions to keep open
          * @return this builder
-         * @deprecated please use {@link #withMaxConnectionPoolSize(int)} instead.
+         * @deprecated please use a combination of {@link #withMaxConnectionPoolSize(int)} and
+         * {@link #withConnectionAcquisitionTimeout(long, TimeUnit)} instead.
          */
         @Deprecated
         public ConfigBuilder withMaxIdleSessions( int size )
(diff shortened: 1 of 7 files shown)


driver/src/main/java/org/neo4j/driver/v1/Config.java
```diff
@@ -117,21 +129,26 @@ public boolean logLeakedSessions()
...
     {
-        return maxIdleConnectionPoolSize;
+        return maxConnectionPoolSize;
     }
 
     /**
      * Max number of idle connections per URL for this driver.
+     *
      * @return the max number of connections
+     * @deprecated please use {@link #maxConnectionPoolSize()} instead.
```
zhenlineo, line 146: I would suggest changing to `Please consider to use #maxConnectionPoolSize` and `#MaxConnectionLifetime` to replace this method instead.

Same applies for comments of all `MaxIdle` methods

driver/src/main/java/org/neo4j/driver/v1/Config.java
```diff
@@ -311,47 +327,54 @@ public ConfigBuilder withLeakedSessionsLogging()
...
          * @return this builder
-         * @deprecated please use {@link #withMaxIdleConnections(int)} instead.
+         * @deprecated please use {@link #withMaxConnectionPoolSize(int)} instead.
          */
         @Deprecated
         public ConfigBuilder withMaxIdleSessions( int size )
         {
-            this.maxIdleConnectionPoolSize = size;
-            return this;
+            return withMaxConnectionPoolSize( size );
```
zhenlineo: I was about to suggest to give a warning here, but then realized we might do not have a logger till this moment...

I feel we should not delegate maxIdle to max, as this config changes meaning already. 
I am wondering if it makes sense to just ignore the value we set to this method. What do you think of the following changes?

```
MaxPoolSize -> MaxConnectionPoolSize
MaxIdleConnections -> ignored
MaxIdleSessions -> ignored
```

zhenlineo (commented):
Wonder if it makes sense to not mix maxIdle and max size, as they are kind of different in meanings?

driver/src/main/java/org/neo4j/driver/v1/Config.java
lutovich: Improved javadoc with your suggestion for `withMaxSessions`, `withMaxIdleSessions`, and `withMaxIdleConnections`.

driver/src/main/java/org/neo4j/driver/v1/Config.java
ali-ince: Since we do not have any idle connection checking in place on the async implementation, ignoring `MaxIdleConnections` and `MaxIdleSessions` config values makes more sense for me.

lutovich: Thanks for pointing this out. I agree, will change

lutovich:
@zhenlineo, @ali-ince suggested config change implemented

PR #433, opened 2017-11-20
Improve connection acquisition timeout error

lutovich:
Previously driver would propagate `TimeoutException` directly from netty's `FixedChannelPool`. This PR changes it to `ClientException`. Also added couple integration tests.

### DIFF
--- a/driver/src/test/java/org/neo4j/driver/v1/integration/CausalClusteringIT.java
+++ b/driver/src/test/java/org/neo4j/driver/v1/integration/CausalClusteringIT.java
@@ -35,7 +35,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
 import org.neo4j.driver.internal.cluster.RoutingSettings;
-import org.neo4j.driver.internal.logging.DevNullLogger;
 import org.neo4j.driver.internal.retry.RetrySettings;
 import org.neo4j.driver.internal.util.ChannelTrackingDriverFactory;
 import org.neo4j.driver.internal.util.FakeClock;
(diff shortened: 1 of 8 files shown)


driver/src/test/java/org/neo4j/driver/v1/util/cc/ClusterRule.java
```diff
@@ -52,11 +53,28 @@ public AuthToken getDefaultAuthToken()
...
+            }
+        }
+    }
+
     @Override
     protected void before() throws Throwable
     {
         assumeTrue( "BoltKit cluster support unavailable", boltKitAvailable() );
 
+        stopSingleInstanceDatabase();
```
zhenlineo, line 76: Do we still have problems to have both single instance and shared cluster running at the same time? This change might make test slower by stopping and restarting server additionally in our tests, right?

driver/src/test/java/org/neo4j/driver/v1/integration/CausalClusteringIT.java
```diff
@@ -85,6 +86,12 @@
     @Rule
     public final ClusterRule clusterRule = new ClusterRule();
 
+    @AfterClass
+    public static void stopSharedCluster()
+    {
+        ClusterRule.stopSharedCluster();
```
zhenlineo, line 92: Why we want to immediately shop the cluster after this test class? Should we actually want to keep the cluster running until the whole tests finish?

If this is the only place that requires a cluster, then maybe we should change the ClusterRule to be a class role?

zhenlineo (commented):
The changes looks good. But have a few questions regarding when to start and stop servers for tests.

driver/src/test/java/org/neo4j/driver/v1/integration/CausalClusteringIT.java
lutovich: Cluster takes a lot of memory and might influence other tests. There exists one more test that uses this rule. I can make it a `@ClassRule` and add cleaning of db in some `@After` method instead.

driver/src/test/java/org/neo4j/driver/v1/util/cc/ClusterRule.java
lutovich: Running both cluster and single instance db leaves only a small amount of free memory on the machine, this makes PR build occasionally hit OOM. I also thought it would make builds significantly slower but it did not, last 3 builds for this change have very similar time as other builds.

zhenlineo (approved):

PR #434, opened 2017-11-21
Prohibit blocking operations in IO threads

lutovich:


driver/src/main/java/org/neo4j/driver/internal/async/EventLoopGroupFactory.java
```diff
@@ -0,0 +1,150 @@
...
+
+import org.neo4j.driver.v1.Session;
+
+/**
+ * Manages creation of Netty {@link EventLoopGroup}s, which are basically {@link Executor}s that perform IO operations.
+ */
+public final class EventLoopGroupFactory
+{
+    private static final String THREAD_NAME_PREFIX = "Neo4jDriverIO";
+    private static final int THREAD_PRIORITY = Thread.MAX_PRIORITY;
```
ali-ince, line 41: I think we should stick with `Thread.NORM_PRIORITY`. Maybe we can depend on a system property to override the default, if such a requirement arises.

driver/src/test/java/org/neo4j/driver/v1/integration/SessionAsyncIT.java
```diff
@@ -1028,6 +1029,65 @@ public void shouldPropagateFailureFromFirstIllegalQuery()
         assertEquals( 0, countNodesByLabel( "Node3" ) );
     }
 
+    @Test
```
ali-ince, line 1032: I think we shall also add some more task chaining tests that mixes up sync / async usage with successful completion i.e. making use of `thenApplyAsync` to make chained task to be run in another executor? It may be good to  address this kind of usage in the docs.

ali-ince (commented):
Changes look good to me along with a couple of comments.

driver/src/main/java/org/neo4j/driver/internal/async/EventLoopGroupFactory.java
lutovich: `MAX_PRIORITY` is what netty uses inside. Here I'm basically duplicating default behaviour just to create threads of our own class. Such thread can then be easily detected with simple `instanceof`.

lutovich:
@ali-ince added unit test to check that created event loop threads are the same as default ones created by netty and couple tests with blocking operations in `#thenXXXAsync`

PR #436, opened 2017-11-27
Relax connection termination policy in routing driver

lutovich:
Previously routing driver terminated all connections towards a particular address when one of active connections had a network error. Connections were also terminated when new routing table did not contain some address that was present in the previous routing table. Such behaviour might be problematic because it results in terminated queries. Network errors might have been temporary but always resulted in termination of active queries.

This PR makes driver keep connections towards machine that had failure but remove it from the routing table. This is done to prevent subsequent queries from using it until rediscovery. After rediscovery address can either appear again in the routing procedure response or not. If it is returned then driver will re-add it to routing table and old pool with all connections will be reused. If address is missing from routing table and pool has no active connections then driver will close the pool.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/async/pool/ConnectionPoolImpl.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/async/pool/ConnectionPoolImpl.java
@@ -25,6 +25,7 @@ import io.netty.channel.pool.ChannelPool;
 import io.netty.util.concurrent.Future;
 
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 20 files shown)


driver/src/main/java/org/neo4j/driver/internal/async/pool/ConnectionPoolImpl.java
```diff
@@ -86,27 +93,30 @@ public ConnectionPoolImpl( ChannelConnector connector, Bootstrap bootstrap, Pool
...
-        // purge idle connections in the pool and pool itself
-        ChannelPool pool = pools.remove( address );
-        if ( pool != null )
+        for ( BoltServerAddress address : pools.keySet() )
         {
-            pool.close();
+            if ( !addressesToRetain.contains( address ) )
+            {
+                int activeChannels = activeChannelTracker.activeChannelCount( address );
+                if ( activeChannels == 0 )
```
zhenlineo, line 103: Will we also purge all `IDLE` connection if the activeChannel is not 0?
If so then maybe also a test: `shouldCloseIdleConnInPoolWhenRetaining`?

FYI: 

 Driver | remove from RT  | close idle  | disallow return to pool 
 ------ | --------------- | ---------- | -------------------------------- 
 JS       | * o | * o | * o 
 python      | * o     |   * o | X X 
 .NET | * o      |    * o | * o 
 Java 1.4 | * o | * o | * o 
 Java 1.5 | * o | X X | X X 

[*]connection error
[o]not in new routing table

lutovich: No, it will not purge idle connections in that case. Whole pool will be closed when `activeChannels == 0` and address is not in the routing table.

PR #437, opened 2017-11-28
Redirect netty logging to driver logging

zhenlineo:
An example of output on DEBUG (FINE) level using our `ConsoleLogging`:
```
2017-11-28 05:53:19,431 [0xb2d66fa6] S: SUCCESS {}
2017-11-28 05:53:19,431 Channel [id: 0xb2d66fa6, L:/127.0.0.1:55381 - R:localhost/127.0.0.1:7687] released back to the pool
2017-11-28 05:53:19,433 Channel [id: 0xb2d66fa6, L:/127.0.0.1:55381 - R:localhost/127.0.0.1:7687] acquired from the pool
2017-11-28 05:53:19,434 [0xb2d66fa6] C: RUN "MATCH (n) WITH n LIMIT 10000 DETACH DELETE n RETURN count(n)" {}
2017-11-28 05:53:19,434 [0xb2d66fa6] C: PULL_ALL
2017-11-28 05:53:19,438 [0xb2d66fa6] S: SUCCESS {fields=["count(n)"], result_available_after=1}
2017-11-28 05:53:19,438 [0xb2d66fa6] S: RECORD [0]
2017-11-28 05:53:19,438 [0xb2d66fa6] S: SUCCESS {result_consumed_after=0, type="rw"}
2017-11-28 05:53:19,438 [0xb2d66fa6] C: RESET
2017-11-28 05:53:19,440 [0xb2d66fa6] S: SUCCESS {}
2017-11-28 05:53:19,440 Channel [id: 0xb2d66fa6, L:/127.0.0.1:55381 - R:localhost/127.0.0.1:7687] released back to the pool
2017-11-28 05:53:19,468 [id: 0x796463d7, L:/127.0.0.1:55382 - R:localhost/127.0.0.1:7687] HANDSHAKEN: TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA
2017-11-28 05:53:19,508 Closing driver instance org.neo4j.driver.internal.InternalDriver@3108bc
2017-11-28 05:53:19,509 Closing connection pool towards localhost:7687
2017-11-28 05:53:21,633 Freed 19 thread-local buffer(s) from thread: Neo4jDriverIO-4-2
```

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/InternalDriver.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/InternalDriver.java
@@ -112,7 +112,7 @@ public class InternalDriver implements Driver
     {
         if ( closed.compareAndSet( false, true ) )
         {
-            log.info( "Driver instance is closing" );
+            log.info( "Closing driver instance %s", this );
             return sessionFactory.close();
         }
         return completedFuture( null );
--- a/driver/src/main/java/org/neo4j/driver/internal/InternalDriver.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/InternalDriver.java
@@ -44,6 +44,7 @@ public class InternalDriver implements Driver
         this.securityPlan = securityPlan;
         this.sessionFactory = sessionFactory;
         this.log = logging.getLog( Driver.class.getSimpleName() );
+        log.info( "Driver instance %s created", this );
     }
 
     @Override
--- a/driver/src/main/java/org/neo4j/driver/internal/async/ChannelConnectedListener.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/async/ChannelConnectedListener.java
@@ -25,12 +25,14 @@ import io.netty.channel.ChannelPipeline;
 import io.netty.channel.ChannelPromise;
 
 import org.neo4j.driver.internal.BoltServerAddress;
+import org.neo4j.driver.internal.logging.ChannelActivityLogger;
 import org.neo4j.driver.v1.Logger;
 import org.neo4j.driver.v1.Logging;
 import org.neo4j.driver.v1.exceptions.ServiceUnavailableException;
 
 import static java.lang.String.format;
-import static org.neo4j.driver.internal.async.ProtocolUtil.handshake;
+import static org.neo4j.driver.internal.async.BoltProtocolV1Util.handshakeBuf;
+import static org.neo4j.driver.internal.async.BoltProtocolV1Util.handshakeString;
 
 public class ChannelConnectedListener implements ChannelFutureListener
 {
@@ -38,7 +40,6 @@ public class ChannelConnectedListener implements ChannelFutureListener
     private final ChannelPipelineBuilder pipelineBuilder;
     private final ChannelPromise handshakeCompletedPromise;
     private final Logging logging;
-    private final Logger log;
 
     public ChannelConnectedListener( BoltServerAddress address, ChannelPipelineBuilder pipelineBuilder,
             ChannelPromise handshakeCompletedPromise, Logging logging )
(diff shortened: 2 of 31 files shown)


driver/src/main/java/org/neo4j/driver/internal/InternalDriver.java
```diff
@@ -112,7 +112,7 @@ public void close()
     {
         if ( closed.compareAndSet( false, true ) )
         {
-            log.info( "Driver instance is closing" );
+            log.info( "Closing driver instance %s", this );
```
lutovich, line 116: Maybe worth adding info log where driver is created? Otherwise printing of `this` does not seem very valuable.

driver/src/main/java/org/neo4j/driver/internal/async/ChannelConnectedListener.java
```diff
@@ -47,21 +48,22 @@ public ChannelConnectedListener( BoltServerAddress address, ChannelPipelineBuild
...
     @Override
     public void operationComplete( ChannelFuture future )
     {
         Channel channel = future.channel();
+        Logger log = new ChannelActivityLogger( channel, logging, getClass() );
 
         if ( future.isSuccess() )
         {
-            log.trace( "Channel %s connected, running bolt handshake", channel );
+            log.trace( "Channel %s connected, initiating bolt handshake", channel );
```
lutovich: `ChannelActivityLogger` will also prefix messages with channel id, so it will appear twice in the message. Maybe better keep using regular logger?

driver/src/main/java/org/neo4j/driver/internal/logging/ConsoleLogging.java
```diff
@@ -94,10 +94,10 @@ public String format( LogRecord record )
             StringBuilder builder = new StringBuilder( 1000 );
             builder.append( dateFormat.format( new Date( record.getMillis() ) ) );
             builder.append(" ");
-            builder.append("[").append(record.getLoggerName()).append("] ");
-            // builder.append( "[" ).append( record.getSourceClassName() ).append( "." );
-            // builder.append( record.getSourceMethodName() ).append( "] - " );
-            // builder.append( "[" ).append( record.getLevel() ).append( "] - " );
+//            builder.append("[").append(record.getLoggerName()).append("] ");
```
lutovich, line 97: Should this line be uncommented? Can we remove next 3 commented lines?

driver/src/main/java/org/neo4j/driver/internal/logging/DelegateLogger.java
```diff
@@ -0,0 +1,425 @@
...
+
+import io.netty.util.internal.logging.InternalLogLevel;
+import io.netty.util.internal.logging.InternalLogger;
+
+import org.neo4j.driver.v1.Logger;
+
+import static java.lang.String.format;
+
+
+public class DelegateLogger implements InternalLogger
```
lutovich: How about calling this smth like `NettyLogger`?
I think we could also extend `io.netty.util.internal.logging.AbstractInternalLogger` to reduce amount of code.

driver/src/main/java/org/neo4j/driver/internal/logging/DelegateLogger.java
```diff
@@ -0,0 +1,425 @@
...
+    private Logger log;
+    public DelegateLogger( Logger log )
+    {
+        this.log = log;
+    }
+
+    @Override
+    public String name()
+    {
+        return null;
```
lutovich: This class could take name as another constructor argument and return it here.

driver/src/main/java/org/neo4j/driver/internal/logging/DelegateLogging.java
```diff
@@ -0,0 +1,43 @@
...
+
+import io.netty.util.internal.logging.InternalLogger;
+import io.netty.util.internal.logging.InternalLoggerFactory;
+
+import org.neo4j.driver.v1.Logging;
+
+/**
+ * This is the logging factory to delegate netty's logging to our logging system
+ */
+public class DelegateLogging extends InternalLoggerFactory
```
lutovich: How about calling this smth like `NettyLogging`?

driver/src/main/java/org/neo4j/driver/internal/logging/DelegateLogger.java
```diff
@@ -0,0 +1,425 @@
...
+    @Override
+    public void warn( String format, Object arg )
+    {
+        log.warn( toDriverLoggerFormat( format ), arg );
+    }
+
+    @Override
+    public void warn( String format, Object... arguments )
+    {
+        log.warn( toDriverLoggerFormat( format ), arguments );
```
lutovich: This method should also treat last argument in a special way if it's `Throwable`. Similar to how `#error()` works

driver/src/main/java/org/neo4j/driver/internal/logging/DelegateLogger.java
```diff
@@ -0,0 +1,425 @@
...
+            break;
+        case ERROR:
+            error( t );
+            break;
+        }
+    }
+
+    private String toDriverLoggerFormat( String format )
+    {
+        return format.replaceAll( "\\{\\}", "%s" );
```
lutovich: It might be better for performance to precompile pattern with `Pattern.compile("\\{\\}")` into a `static final` constant and then use it like `PATTERN.matcher(format).replaceAll("%s")`.

lutovich (commented):
@zhenlineo review completed. Made some comments & suggestions.

driver/src/main/java/org/neo4j/driver/internal/InternalDriver.java
zhenlineo: Added another logging in constructor

driver/src/main/java/org/neo4j/driver/internal/logging/ConsoleLogging.java
zhenlineo: This ConsoleLogging class is just for internal use purpose. I would prefer keep these commented there in case if we ever need to print more info to help debug.

PR #438, opened 2017-11-29
Backpressure for reading records

lutovich:
Previously driver would read records from network even if client did not consume them. This happened because event loop threads were always in auto-read mode despite the amount of currently buffered records. This could lead to memory problems when handling large results that are slowly consumed.

This PR adds simple network-level backpressure for reading records. Auto-read mode is turned off when amount of buffered records reaches high watermark. It is turned on again when bunch of records are consumed and buffer size reaches low watermark. Also added unit tests for PULL_ALL and RUN handlers.


ali-ince (approved):
LGTM!

PR #439, opened 2017-11-30
Fixed auto-read problem with result summary

lutovich:
All result stream should be buffered/consumed when result summary is requested or when session is closed. This means underlying channel should read everything that's available. It did not do so when summary was requested after number of buffered records exceeded high watermark.

This PR makes fetching of summary or failure enable auto-read on the underlying channel. So SUCCESS and FAILURE messages will be read without waiting for all records to be consumed.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/handlers/PullAllResponseHandler.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/handlers/PullAllResponseHandler.java
@@ -170,6 +170,10 @@ public abstract class PullAllResponseHandler implements ResponseHandler
         {
             if ( summaryFuture == null )
             {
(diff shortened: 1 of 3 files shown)


ali-ince (approved):
Changes look ok for me. One thing I found myself thinking is whether we should by-pass queuing recods in `PullAllResponseHandler` when `StatementResultCursor#consumeAsync` is called. WDYT?

lutovich:
Yeah, I think it's possible. Currently `PullAllResponseHandler` exposes a fairly small set of methods and `#consume()` builds on them. We could move it to `PullAllResponseHandler` and skip queueing. Similar optimization can be made for `#listAsync()` - we could just return a copy of the records queue.

PR #440, opened 2017-11-30
Add example for passing bookmarks

ali-ince:


examples/src/main/java/org/neo4j/docs/driver/PassBookmarkExample.java
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.neo4j.docs.driver;
+
+// tag::=pass-bookmarks-import[]
```
lutovich: Is `=` needed in `tag::=pass-bookmarks-import[]` and `end::=pass-bookmarks-import[]`?

lutovich (approved):
@ali-ince lgtm, one small comment. Love these company names!

examples/src/main/java/org/neo4j/docs/driver/PassBookmarkExample.java
ali-ince: Company names are owned by @technige and I'm not sure how I ended up with those `=` signs :)

PR #441, opened 2017-12-01
Handle thread interruption by closing the channel

lutovich:
Blocking API like `Session#run()` is now implemented on top of async API, i.e. `Session#runAsync()`. Blocking calls simply invoke async ones and uninterruptibly wait for them to complete. Previously, with blocking IO, thread interruptions resulted in `ClosedByInterruptException` which terminated query/transaction in a harsh way. Right now interrupts will do nothing. Handling them is not strictly required because `Thread#interrupt()` is not a nice thing to do. However, interrupts are still used in some environments as a last resort thing.

This commit makes driver react on thread interrupts by closing the underlying Netty channel. It's only done in places where channel is actually available. Closing of channel results in exception and the waiting thread fails. It will also make database reset state for this connection, terminate transaction and query.

Such approach seems very similar to the previous one except it will result in `ServiceUnavailableException` instead of `ClosedByInterruptException`.

### DIFF
--- a/driver/src/test/java/org/neo4j/driver/v1/GraphDatabaseTest.java
+++ b/driver/src/test/java/org/neo4j/driver/v1/GraphDatabaseTest.java
@@ -171,7 +171,7 @@ public class GraphDatabaseTest
                 GraphDatabase.driver( "bolt://localhost:" + serverSocket.getLocalPort() );
                 fail( "Exception expected" );
             }
-            catch ( Exception ignore )
+            catch ( ServiceUnavailableException ignore )
             {
                 // expected
             }
--- a/driver/src/test/java/org/neo4j/driver/v1/GraphDatabaseTest.java
+++ b/driver/src/test/java/org/neo4j/driver/v1/GraphDatabaseTest.java
@@ -21,11 +21,13 @@ package org.neo4j.driver.v1;
 import org.junit.Test;
 
 import java.io.File;
+import java.net.ServerSocket;
 import java.net.URI;
 import java.util.List;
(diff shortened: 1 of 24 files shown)


ali-ince (approved):
Changes look good to me!

lutovich:
Added a commit with some more tests. @ali-ince could you please take a look?

driver/src/test/java/org/neo4j/driver/v1/GraphDatabaseTest.java
```diff
@@ -156,4 +158,28 @@ public void shouldLogWhenUnableToCreateRoutingDriver() throws Exception
...
+            // setup other thread to interrupt current thread when it blocks
+            TestUtil.interruptWhenInWaitingState( Thread.currentThread() );
+            try
+            {
+                GraphDatabase.driver( "bolt://localhost:" + serverSocket.getLocalPort() );
+                fail( "Exception expected" );
+            }
+            catch ( Exception ignore )
+            {
+                // expected
```
ali-ince: @lutovich Would it be better to expect a `ServiceUnavailableException` here instead of a `IllegalStateException`?

lutovich: Yeah, that's true. What do you think about always wrapping exceptions from connectivity verification in `ServiceUnavailableException`? (unless it's already a `ServiceUnavailableException`)

lutovich:
@ali-ince exceptions should now be fixed

PR #442, opened 2017-12-08
Use connect timeout in Bolt and TLS handshake

lutovich:
Previously configured connection timeout has only been applied to actual `Socket#connect()` call. This is not the only thing driver does to establish a connection. It also executes TLS handshake (if configured to use encryption) and Bolt handshake to negotiate protocol version with the database. Later two steps perform blocking reads without any timeout.

This could be a problem when database does not respond in time. Also driver might be simply connecting to something that is not the database and never responds.

This commit makes driver use configured value of connect timeout as blocking read timeout for TLS and Bolt handshakes. So both will not hang forever when other side does not respond. Default value of connection timeout is 5 seconds. With this commit driver will wait up to 5 seconds for TLS and Bolt handshake.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/net/ChannelFactory.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/net/ChannelFactory.java
@@ -22,9 +22,7 @@ import java.io.IOException;
 import java.net.ConnectException;
 import java.net.Socket;
 import java.net.SocketTimeoutException;
-import java.net.StandardSocketOptions;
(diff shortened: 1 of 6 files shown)


ali-ince (approved):
Changes look good to me! What do you think of adding/changing socket options (`keepalive`, `buffersizes`, `reuseaddress`, `solinger`, `tcpnodelay`) based on system properties?

lutovich:
@ali-ince makes sense. I'll create a card to do it as a separate PR.

PR #447, opened 2017-12-14
Use connect timeout in Bolt and TLS handshake

lutovich:
Previously configured connection timeout has only been used to limit amount of time it takes to establish a TCP connection. This is not the only thing driver does to establish a logical connection. It also executes TLS handshake (if configured to use encryption) and Bolt handshake to negotiate protocol version with the database. Later two steps perform reads without any timeout.

This could be a problem when database does not respond in time. Also driver might be simply connecting to something that is not the database and never responds.

This commit makes driver use configured value of connect timeout as read timeout for TLS and Bolt handshakes. So both will not hang forever when other side does not respond. Default value of connection timeout is 5 seconds. With this commit driver will wait up to 5 seconds for TLS and Bolt handshakes. Timeout is enforced by `ConnectTimeoutHandler` which is added to the channel pipeline when connection is established and removed from it when Bolt handshake completes.


ali-ince (approved):
:+1:

PR #451, opened 2017-12-19
Couple tweaks to generate less garbage

lutovich:
PR makes the following changes:
 * static final `CompletableFuture`s completed with `null` and `false`
 * improved initial sizing of `HashMap`s so that they do not resize
 * `BoltServerAddress#equals()` now checks port first and then host
 * `ActiveChannelTracker` only stores channel counts, not actual channels
 * improved `#listAsync()` to transform records buffered in the queue

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
@@ -46,7 +46,6 @@ import org.neo4j.driver.v1.Values;
 import org.neo4j.driver.v1.exceptions.ClientException;
 import org.neo4j.driver.v1.types.TypeSystem;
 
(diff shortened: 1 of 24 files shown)


driver/src/main/java/org/neo4j/driver/internal/NetworkSession.java
```diff
@@ -276,7 +277,7 @@ public TypeSystem typeSystem()
     {
         if ( connectionStage == null )
         {
-            return completedFuture( false );
+            return completedWithFalse();
```
ali-ince: Since `connectionStage` is initialized with `completedWithNull`, this `connectionStage == null` check can be removed or checked against `completedWithNull`.

ali-ince (approved):
:+1: with one small comment.

PR #461, opened 2018-01-19
Expose point type with Bolt V2

lutovich:
Introduced Bolt V1 as an extension of Bolt V1 with point type support. Added interfaces to expose point with single multi-dimensional coordinate. These interfaces are now experimental and will most likely change before the next stable release.


ali-ince:
Changes looks good to me!

PR #463, opened 2018-01-22
Fixed issue with connection pool deactivation

lutovich:
Previously load balancer moved connection pool to deactivated state when corresponding cluster member had a network error. This was made to disallow any new connections towards that member. Member has also been removed from the routing table, so that callers never try to acquire connection from a deactivated pool. Deactivated pools were re-activated during rediscovery.

However, there was a case when deactivated pool towards the seed router would remain deactivated forever without a chance to be re-activated. It happened when connections to all cores failed and driver had to perform rediscovery using seed router. In this case all connections pools were deactivated, and rediscovery was not able to complete because it failed trying to obtain connection from a deactivated pool.

This PR fixes the problem by removing pool activation/deactivation. Instead driver will simply make instance non-routable by removing it's address from the routing table. Corresponding connection pool will not be changed. Later rediscovery will cleanup pools for non-routable addresses that have no active connections.

Same connection error handling logic is in 1.5. Only tests should be merged forward.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/security/TLSSocketChannel.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/security/TLSSocketChannel.java
@@ -372,6 +372,8 @@ public class TLSSocketChannel implements ByteChannel
                 cipherOut.compact();
             }
             break;
+        case CLOSED:
+            throw new IOException( "TLS socket channel is closed" );
         default:
             throw new ClientException( "Got unexpected status " + status );
         }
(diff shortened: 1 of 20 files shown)


driver/src/main/java/org/neo4j/driver/internal/security/TLSSocketChannel.java
```diff
@@ -372,6 +372,8 @@ private HandshakeStatus wrap( ByteBuffer buffer ) throws IOException, ClientExce
                 cipherOut.compact();
             }
             break;
+        case CLOSED:
```
zhenlineo, line 375: Just wonder why this is needed? Is it a fix to something?

zhenlineo (approved):

driver/src/main/java/org/neo4j/driver/internal/security/TLSSocketChannel.java
lutovich: Added stress test closes all connections to simulate them breaking. Without this branch closing might result in `ClientException`, which is not retriable. It feels better to have it as `IOException` so that it bubbles up to users as `ServiceUnavailableException` or `SessionExpiredException`. Does this sound sensible?

ali-ince (approved):
👍

PR #467, opened 2018-01-25
Do not limit TLS and Bolt handshakes by default

lutovich:
Both can now timeout if server does not respond in 5 seconds. This is stricter than just connect timeout of 5 seconds and might result in more transient exceptions when database is under high load.

This PR disables timeout for TLS and Bolt handshakes, which makes behaviour exactly like in the previous driver version. Feature can still be enabled with "timeoutTlsAndBoltHandshakes" boolean system property. This feels like a safe thing to do in a maintenance branch.


lutovich:
Decided to simply do a revert instead.

PR #470, opened 2018-02-13
Draft of driver metrics

zhenlineo:
Added Basics of Driver metrics and Connection pool metrics.
The driver metrics are enabled with system property `driver.metrics.enabled=true`

PR #471, opened 2018-02-28
Metrics Enhancement

zhenlineo:
Based on #470 
Added connection `acquiring`, `acquired` count and `timedOutToAcquire` count on connection pool metrics
Change the histogram only record value generated by successful requests.
Fixed a bug where the pool status was showing the big pool rather than each small pool status.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/async/pool/NettyChannelTracker.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/async/pool/NettyChannelTracker.java
@@ -26,7 +26,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.neo4j.driver.internal.BoltServerAddress;
-import org.neo4j.driver.internal.metrics.ListenerEvent;
+import org.neo4j.driver.internal.metrics.ListenerEvent.ConnectionListenerEvent;
 import org.neo4j.driver.internal.metrics.MetricsListener;
 import org.neo4j.driver.v1.Logger;
 import org.neo4j.driver.v1.Logging;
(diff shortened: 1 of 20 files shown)


driver/src/main/java/org/neo4j/driver/internal/async/pool/NettyChannelTracker.java
```diff
@@ -65,26 +65,27 @@ public void channelAcquired( Channel channel )
     @Override
     public void channelCreated( Channel channel )
     {
-        log.debug( "Channel %s created", channel );
-        incrementInUse( channel );
-        metricsListener.afterCreated( serverAddress( channel ) );
+        throw new IllegalStateException( "A fatal error happened in this driver as this method should never be called. " +
```
lutovich: Maybe just `throw new UnsupportedOperationException()`? Current message reminds me of `ThisShouldNotHappenError` :)

driver/src/main/java/org/neo4j/driver/internal/metrics/ListenerEvent.java
```diff
@@ -18,10 +18,18 @@
...
 package org.neo4j.driver.internal.metrics;
 
-
 public interface ListenerEvent
 {
     void start();
+
     long elapsed();
+
+    interface PoolListenerEvent extends ListenerEvent
```
lutovich: Why are these empty interfaces needed? Can't `ListenerEvent` be used everywhere?

driver/src/main/java/org/neo4j/driver/internal/metrics/spi/PoolStatus.java
```diff
@@ -20,5 +20,18 @@
 
 public enum PoolStatus
 {
-    Open, Closed
+    Open( 0 ),
+    Closed( 1 );
+
+    private final int value;
```
lutovich: Can this field be removed? It's really up to the users how to interpret the returned status

driver/src/main/java/org/neo4j/driver/internal/metrics/spi/PoolStatus.java
```diff
@@ -20,5 +20,18 @@
 
 public enum PoolStatus
 {
-    Open, Closed
+    Open( 0 ),
```
lutovich: I'd vote for renaming enum values to `OPEN` and `CLOSED` which follows the recommended java code style

lutovich (changes requested):
@zhenlineo changes look good. Made couple comments.

driver/src/main/java/org/neo4j/driver/internal/metrics/ListenerEvent.java
zhenlineo: For 
`MetricsListener#afterAcquiredOrCreated( BoltServerAddress serverAddress, PoolListenerEvent acquireEvent );`
and
`MetricsListener#void afterAcquiredOrCreated( BoltServerAddress serverAddress, ConnectionListenerEvent acquireEvent );`

It is also more clear API to use.

lutovich: `ListenerEvent.PoolListenerEvent` and `ListenerEvent.ConnectionListenerEvent` are the same as `ListenerEvent`. I do not really see the reason for having them right now. They only force us to type more :)

driver/src/main/java/org/neo4j/driver/internal/metrics/spi/PoolStatus.java
zhenlineo: But I kind of do not want to have a if-else just for this conversion when publishing the result.

```metrics.add( new Metric( path( serverName, "poolStatus" ), metric.poolStatus().value(), date ) );```

vs.

```
PoolStatus staus = metric.poolStatus()
int statusNum = -1; 
if( status == PoolStatus.OPEN)
{
    statusNum = 0;
}
else if( status == PoolStatus.CLOSED)
{
   statusNum = 1;
}
metrics.add( new Metric( path( serverName, "poolStatus"), statusNum, date ) );
```

I feel for a metrics collector, it does not matter what value you assigned to each PoolStatus.
It is more important to record all possible PoolStatus. I mean it would be good to know the real value, but it will not be a big issue if a user does not know as long as they see they are different. So I thought it would be helpful to have a simple `value` to use directly.

lutovich (approved):

PR #473, opened 2018-03-05
Fixed a bug found on idle and inUse counters

zhenlineo:
Before we always decrement idle counters when a channel is closed. However, this is wrong when a channel is closed before it is returen to the pool.
So change the code to make sure that no matter the channel is closed inside or outside the pool, the idle and inUse counters should be correct.

### DIFF
--- a/driver/src/test/java/org/neo4j/driver/internal/async/pool/NettyChannelTrackerTest.java
+++ b/driver/src/test/java/org/neo4j/driver/internal/async/pool/NettyChannelTrackerTest.java
@@ -120,7 +120,7 @@ public class NettyChannelTrackerTest
         assertEquals( 0, tracker.idleChannelCount( address ) );
 
         // When closed before session.close
-        channel.close();
+        channel.close().sync();
 
         // Then
         assertEquals( 1, tracker.inUseChannelCount( address ) );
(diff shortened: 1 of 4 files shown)


driver/src/test/java/org/neo4j/driver/internal/async/pool/NettyChannelTrackerTest.java
```diff
@@ -110,6 +110,48 @@ public void shouldDecrementCountForAddress()
...
+    public void shouldDecreaseIdleWhenClosedOutsidePool() throws Throwable
+    {
+        // Given
+        Channel channel = newChannel();
+        tracker.channelCreated( channel, null );
+        assertEquals( 1, tracker.inUseChannelCount( address ) );
+        assertEquals( 0, tracker.idleChannelCount( address ) );
+
+        // When closed before session.close
+        channel.close();
```
lutovich: Maybe `channel.close().sync()`?
`Channel#close()` is async and returns a future.

driver/src/test/java/org/neo4j/driver/internal/async/pool/NettyChannelTrackerTest.java
```diff
@@ -110,6 +110,48 @@ public void shouldDecrementCountForAddress()
...
+        tracker.channelCreated( channel, null );
+        assertEquals( 1, tracker.inUseChannelCount( address ) );
+        assertEquals( 0, tracker.idleChannelCount( address ) );
+
+        tracker.channelReleased( channel );
+        assertEquals( 0, tracker.inUseChannelCount( address ) );
+        assertEquals( 1, tracker.idleChannelCount( address ) );
+
+        // When closed before acquire
+        channel.close();
```
lutovich: `channel.close().sync();`

lutovich (approved):

PR #474, opened 2018-03-07
Temporal types

lutovich:
PR makes driver support new temporal Cypher types:
 * Date exposed as `java.time.LocalDate`
 * Time exposed as `java.time.OffsetTime`
 * LocalTime exposed as `java.time.LocalTime`
 * LocalDateTime exposed as `java.time.LocalDateTime`
 * DateTime exposed as `java.time.ZonedDateTime`
 * Duration exposed as custom `org.neo4j.driver.v1.types.Duration` interface containing months, days, seconds and nanoseconds

Also simplified couple things around the `Value` class hierarchy.


driver/src/main/java/org/neo4j/driver/v1/Values.java
```diff
@@ -87,9 +101,17 @@ public static Value value( Object value )
         if ( value instanceof Integer ) { return value( (int) value ); }
         if ( value instanceof Double ) { return value( (double) value ); }
         if ( value instanceof Float ) { return value( (float) value ); }
+        if ( value instanceof LocalDate ) { return value( (LocalDate) value ); }
+        if ( value instanceof OffsetTime ) { return value( (OffsetTime) value ); }
+        if ( value instanceof LocalTime ) { return value( (LocalTime) value ); }
+        if ( value instanceof LocalDateTime ) { return value( (LocalDateTime) value ); }
+        if ( value instanceof ZonedDateTime ) { return value( (ZonedDateTime) value ); }
+        if ( value instanceof Duration ) { return value( (Duration) value ); }
```
zhenlineo: It seems this `Duration` is our `org.neo4j.driver.v1.types.Duration` type, do we also want to support java native `Duration` and/or `Period`?

driver/src/main/java/org/neo4j/driver/internal/InternalDuration.java
```diff
@@ -0,0 +1,186 @@
...
+import static java.util.Collections.unmodifiableList;
+
+public class InternalDuration implements Duration
+{
+    private static final List<TemporalUnit> SUPPORTED_UNITS = unmodifiableList( asList( MONTHS, DAYS, SECONDS, NANOS ) );
+
+    private final long months;
+    private final long days;
+    private final long seconds;
+    private final long nanoseconds;
```
zhenlineo: Maybe instead of having `months, days, seconds, nanoseconds` 4 field, we could have `Period period` and `Duration duration`? Then the calculation such as add and minus would be easier?

driver/src/main/java/org/neo4j/driver/v1/types/Duration.java
```diff
@@ -0,0 +1,61 @@
...
+import org.neo4j.driver.v1.Values;
+import org.neo4j.driver.v1.util.Immutable;
+
+/**
+ * Represents temporal amount containing months, days, seconds and nanoseconds of the second. A duration¬ can be negative.
+ * <p>
+ * Value that represents a duration can be created using {@link Values#duration(long, long, long, long)} method.
+ */
+@Immutable
+public interface Duration extends TemporalAmount
```
zhenlineo: Maybe `IsoStandardDuration`?

driver/src/main/java/org/neo4j/driver/v1/Values.java
lutovich: Done in https://github.com/neo4j/neo4j-java-driver/pull/474/commits/a355fc4360849ea8fc8a8965ffd63d268c574dd4

driver/src/main/java/org/neo4j/driver/internal/InternalDuration.java
lutovich: We receive months and days as `long`s from the database but `java.time.Period` only allows `int`s

driver/src/main/java/org/neo4j/driver/v1/types/Duration.java
lutovich: Renamed to `IsoDuration` in https://github.com/neo4j/neo4j-java-driver/pull/474/commits/bb31fc02532fc16ac9d164fc26ed622974d74a0b

ali-ince (approved):
Changes look good to me.

@technige used `WITH $x AS x RETURN x.hour, x.minute, x.second` kind of queries in ITs which I think is a good way to check consistency across all layers. We may also consider it for the rest of the drivers.

PR #475, opened 2018-03-14
Replaced Point2D and Point3D with Point

zhenlineo:
Based on #474


driver/src/main/java/org/neo4j/driver/v1/types/Point.java
```diff
@@ -22,19 +22,20 @@
 import org.neo4j.driver.v1.util.Immutable;
 
 /**
- * Represents a single tree-dimensional point in a particular coordinate reference system.
+ * Represents a single point in a particular coordinate reference system.
  * <p>
- * Value that represents a 3D point can be created using {@link Values#point3D(long, double, double, double)} method.
+ * Value that represents a point can be created using {@link Values#point(int, double, double)} method.
```
ali-ince: Shall we also add a reference to the `Values#point(int, double, double, double)` overload?

driver/src/main/java/org/neo4j/driver/internal/InternalPoint3D.java
```diff
@@ -20,16 +20,21 @@
...
+public class InternalPoint3D implements Point
 {
-    private final long srid;
+    private final int srid;
     private final double x;
     private final double y;
     private final double z;
 
-    public InternalPoint3D( long srid, double x, double y, double z )
+    public InternalPoint3D( int srid, double x, double y )
```
ali-ince: Do we need this constructor since we already have a distinct `InternalPoint2D`?

ali-ince (approved):
I think changes are Ok.

I made a couple of small comments. Apart from this, 3D CRS support seems to be merged on the server side, so I think we can add 3D point tests to the `SpatialTypesIT` - probably as another PR?

PR #481, opened 2018-04-03
Add `Value#asXXX(defalutValue)` for primary types

zhenlineo:
Add `Value#computeOrDefault(Function<Value, T> mapper, T defaultValue)` for other types
Add `Value#computeOrNull(Function<Value, T> mapper)` for shortcuts to call `Value#computeOrDefault(Function<Value, T> mapper, T defaultValue=null)`

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/v1/Value.java
+++ b/driver/src/main/java/org/neo4j/driver/v1/Value.java
@@ -205,14 +205,6 @@ public interface Value extends MapAccessor, MapAccessorWithDefaultValue
     <T>T computeOrDefault( Function<Value, T> mapper, T defaultValue );
 
     /**
-     * Apply the mapping function on the value if the value is not a {@link NullValue}, or null if the value is a {@link NullValue}.
-     * @param mapper The mapping function defines how to map a {@link Value} to T.
-     * @param <T> The return type
-     * @return The value after applying the given mapping function or null if the value is {@link NullValue}.
-     */
-    <T>T computeOrNull( Function<Value, T> mapper );
-
-    /**
      * @return the value as a Java boolean, if possible.
      * @throws Uncoercible if value types are incompatible.
      */
@@ -230,6 +222,13 @@ public interface Value extends MapAccessor, MapAccessorWithDefaultValue
      *  @throws Uncoercible if value types are incompatible.
      */
     byte[] asByteArray();
+
+    /**
+     *  @param defaultValue default to this value if the original value is a {@link NullValue}
+     *  @return the value as a Java byte array, if possible.
+     *  @throws Uncoercible if value types are incompatible.
+     */
+    byte[] asByteArray( byte[] defaultValue );
 
     /**
      *  @return the value as a Java String, if possible.
@@ -331,6 +330,17 @@ public interface Value extends MapAccessor, MapAccessorWithDefaultValue
      */
     List<Object> asList();
 
+
+    /**
+     * If the underlying type can be viewed as a list, returns a java list of
+     * values, where each value has been converted using {@link #asObject()}.
+     *
+     * @see #asObject()
+     * @param defaultValue default to this value if the value is a {@link NullValue}
+     * @return the value as a Java list of values, if possible
+     */
+    List<Object> asList( List<Object> defaultValue );
+
     /**
      * @param mapFunction a function to map from Value to T. See {@link Values} for some predefined functions, such
      * as {@link Values#ofBoolean()}, {@link Values#ofList(Function)}.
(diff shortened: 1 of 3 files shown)


driver/src/main/java/org/neo4j/driver/v1/Value.java
```diff
@@ -194,12 +195,36 @@
...
+    <T>T computeOrNull( Function<Value, T> mapper );
+
     /**
      * @return the value as a Java boolean, if possible.
      * @throws Uncoercible if value types are incompatible.
      */
     boolean asBoolean();
 
+    /**
+     * @param defaultValue return this value if the value is a {@link NullValue}.
```
lutovich, line 214: Maybe add description and `@throws Uncoercible if value types are incompatible.`

driver/src/main/java/org/neo4j/driver/v1/Value.java
```diff
@@ -212,6 +237,12 @@
      */
     String asString();
 
+    /**
+     * @param defaultValue return this value if the value is null.
```
lutovich, line 240: Maybe add description and `@throws Uncoercible if value types are incompatible.`

driver/src/main/java/org/neo4j/driver/v1/Value.java
```diff
@@ -245,6 +294,16 @@
      */
     double asDouble();
 
+    /**
+     * Returns a Java double if no precision is lost in the conversion.
+     * @param defaultValue default to this value if the value is a {@link NullValue}.
+     * @return the value as a Java double.
+     * @throws LossyCoercion if it is not possible to convert the value without loosing precision.
+     * @throws Uncoercible if value types are incompatible.
+
```
lutovich: Empty line should probably be removed

driver/src/main/java/org/neo4j/driver/v1/Value.java
```diff
@@ -254,6 +313,16 @@
      */
     float asFloat();
 
+    /**
+     * Returns a Java float if no precision is lost in the conversion.
+     * @param defaultValue default to this value if the value is a {@link NullValue}
+     * @return the value as a Java float.
+     * @throws LossyCoercion if it is not possible to convert the value without loosing precision.
+     * @throws Uncoercible if value types are incompatible.
+
```
lutovich: Empty line should probably be removed

lutovich (commented):
@zhenlineo Changes look good to me. Made couple minor comments. Methods with default value should also be added for native temporal types from `java.time`:
 * `LocalDate asLocalDate(LocalDate defaultValue)`
 * `OffsetTime asOffsetTime(OffsetTime defaultValue)`
 * `LocalTime asLocalTime(LocalTime defaultValue)`
 * `LocalDateTime asLocalDateTime(LocalDateTime defaultValue)`
 * `ZonedDateTime asZonedDateTime(ZonedDateTime defaultValue)`

driver/src/main/java/org/neo4j/driver/v1/Value.java
zhenlineo: It was in the javadoc.

driver/src/main/java/org/neo4j/driver/v1/Value.java
zhenlineo: Done

zhenlineo:
@sarmbruster Given we have general methods
`Values#computeOrDefault(Function<Value, T> mapper, T defaultValue)` and `Value#computeOrNull(Function<Value, T> mapper)`
for all types to use, 

On which level do you suggest to have `asXXX(defaultVal)` methods?
1) primary types + String only, 
2) all Java Native types
3) all types that are not Node, Path and Rel
4) other

sarmbruster:
Hi @zhenlineo,
what's the difference between primary and native types? (1 and 2)
I basically want defaults to be supported for all property values. So `String`, primitive types, `spatial` and `temporal` types, and arrays

zhenlineo:
@sarmbruster I mean primitive types.

So basically I guess you are asking for the following API:
```
asString(String def);

// primitives 
asBoolean(boolean def);
asInt(int def);
asLong(long def);
asFloat(float def);
asDouble(double def);

asNumber(Number def);

// array, list, map
asByteArray(byte[] def);
asList(List def);
asMap(Map def);

asObject(Object def);

// node, rel, path
asEntityOrNull();
asNodeOrNull();
asRelationshipOrNull();
asPathOrNull();

// asEntity(Entity entity)   ===> Node + Rel = Entity, Node, Rel, and Path cannot be created so only default to null
// asNode(Node node)
// asRelationship(Relationship rel)
// asPath(Path path)

// spatial, temporal
asPoint(Point def);
asLocalDate(LocalDate d);
asOffsetTime(OffsetTime t);
asLocalTime(LocalTime t);
asLocalDateTime(LocalDateTime d);
asZonedDateTime(ZonedDateTime d);
asIsoDuration(IsoDuration d);
```

But how about I give you a smaller API:
```
asString(String def);

// primitives
asBoolean(boolean def);
asInt(int def);
asLong(long def);
asFloat(float def);
asDouble(double def);

computeOrDefault(Function<Value, T> mapper, T defaultValue)
computeOrNull(Function<Value, T> mapper)

// The following are equivalent:
// Point point = pointValue.asPoint(myDefaultPoint) 
// Point point = pointValue.computeOrDefault(Value::asPoint, myDefaultPoint)

// We also have an extra helper method:
// Point pointOrNull = pointValue.computeOrNull(Value::asPoint);
```
Which API is more useful to you (most of users)? or how would you improve one of them?

lutovich:
Imho a slightly trimmed version 1 would be great:

```
asString(String def);

// primitives
asBoolean(boolean def);
asInt(int def);
asLong(long def);
asFloat(float def);
asDouble(double def);

// array, list, map
asByteArray(byte[] def);
asList(List def);
asMap(Map def);

// spatial, temporal
asPoint(Point def);
asLocalDate(LocalDate def);
asOffsetTime(OffsetTime def);
asLocalTime(LocalTime def);
asLocalDateTime(LocalDateTime def);
asZonedDateTime(ZonedDateTime def);
asIsoDuration(IsoDuration def);
```

sarmbruster:
+1 on @lutovich 's previous commit. I guess we should have a `asString(String def)` as well - it seems to be missing in the list.
These `asXXXX` plus `asString` would IMHO cover >90% of all uses.

For the remainder having a type agnostic `computeOrDefault` is a good idea. `computeOrNull` is IMHO not needed since it's just as `computeOrDefault, mapper, null)`.

PR #482, opened 2018-04-03
Adding error check of unsupported types

zhenlineo:
Fixes https://github.com/neo4j/neo4j/issues/10239

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/v1/Values.java
+++ b/driver/src/main/java/org/neo4j/driver/v1/Values.java
@@ -49,8 +49,11 @@ import org.neo4j.driver.internal.value.ListValue;
 import org.neo4j.driver.internal.value.LocalDateTimeValue;
 import org.neo4j.driver.internal.value.LocalTimeValue;
 import org.neo4j.driver.internal.value.MapValue;
+import org.neo4j.driver.internal.value.NodeValue;
 import org.neo4j.driver.internal.value.NullValue;
+import org.neo4j.driver.internal.value.PathValue;
 import org.neo4j.driver.internal.value.PointValue;
+import org.neo4j.driver.internal.value.RelationshipValue;
 import org.neo4j.driver.internal.value.StringValue;
 import org.neo4j.driver.internal.value.TimeValue;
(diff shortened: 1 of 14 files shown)


driver/src/main/java/org/neo4j/driver/v1/Values.java
```diff
@@ -91,6 +94,10 @@ public static Value value( Object value )
     {
         if ( value == null ) { return NullValue.NULL; }
 
+        if ( value instanceof NodeValue || value instanceof RelationshipValue || value instanceof PathValue )
```
lutovich: I think it will be better to prohibit node, relationship and path in `PackStreamMessageFormatV1`. It is clear that struct packing code is the single point that writes structs to the channel. While `Values#value()` is just one of many factory methods.

driver/src/test/java/org/neo4j/driver/internal/value/NodeValueTest.java
```diff
@@ -70,12 +70,12 @@ public void shouldTypeAsNode()
         assertThat( value.typeConstructor(), equalTo( TypeConstructor.NODE ) );
     }
 
-    private NodeValue emptyNodeValue()
+    public static NodeValue emptyNodeValue()
```
lutovich: Please extract `NodeValueTest#emptyNodeValue()`, `PathValueTest#pathValue()` and `RelationshipValueTest#emptyRelationshipValue()` into a test utility class. Alternatively `ValuesTest` could just create it's own node, relationship and path values.

lutovich (changes requested):
@zhenlineo review completed. Made two comments. It would also be nice to add ITs which use nodes, relationships and paths as query parameters.

driver/src/main/java/org/neo4j/driver/v1/Values.java
zhenlineo: The reason to do it outside/earlier is because when we got an error there, we will kill the connection, which seems too aggressive. So if we could found the error in advance, then we will simply pop an error to the surface.

lutovich (approved):

PR #483, opened 2018-04-05
Print ISO standard format in `IsoDuration#toString()`

zhenlineo:
Fixed a bug where the size of nanos should be `int`


driver/src/main/java/org/neo4j/driver/internal/InternalIsoDuration.java
```diff
@@ -188,11 +195,65 @@ public int hashCode()
...
     public String toString()
     {
-        return "Duration{" +
-               "months=" + months +
-               ", days=" + days +
-               ", seconds=" + seconds +
-               ", nanoseconds=" + nanoseconds +
-               '}';
+        // print the duration in iso standard format.
+        if ( this == ZERO )
```
lutovich: Maybe `this.equals(ZERO)`? Then last `if` with "this was all zeros (but not ZERO for some reason)" comment can be removed.

lutovich (commented):
@zhenlineo review completed. Made one comment. Would be also nice to add tests for zero and negative duration.

PR #497, opened 2018-05-07
Handle nanosecond normalization in IsoDuration

lutovich:


sherfert (commented):
From what I can see in the tests, it looks fine to me. I can not see any obvious flaws in the implementation. The implementation is different from the one in the server, what is the reason?

lutovich:
@sherfert server also does other conversions like 12 months -> year, etc. We do not want this it drivers for now. So this is a minimal fix for seconds & nanoseconds handling.

PR #499, opened 2018-05-14
Logging API improvements

lutovich:
1. Cleanup console logging to not use not thread-safe `SimpleDateFormat` and include level and logger name in the message
2. SLF4J support via an optional Maven dependency and new logging implementation
3. More javadocs around `Logging`, `Logger` and corresponding config
4. New public API to create instances of `Logging`:
    * `Logging#slf4j()`
    * `Logging#console(Level)`
    * `Logging#javaUtilLogging(Level)`
    * `Logging#none()`

### DIFF
--- a/driver/pom.xml
+++ b/driver/pom.xml
@@ -34,6 +34,13 @@
       <artifactId>HdrHistogram</artifactId>
     </dependency>
 
+    <!-- Optional dependencies -->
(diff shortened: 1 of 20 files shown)


zhenlineo (approved):

ali-ince (approved):
👍

PR #500, opened 2018-05-16
Support hostname verification for secure connections

lutovich:
Hostname verification makes client verify that server it is trying to connect to matches server specified in the returned certificate. This is a way to protect from man in the middle attack and make sure client is talking to the correct server. Hostname will be compared to all entries in `subjectAltName` field of the certificate.

This PR makes driver support configurable hostname verification. It can be configured together with the trust strategy:

```java
Config config = Config.build()
    .withTrustStrategy(trustAllCertificates().withHostnameVerification())
    .toConfig();
```

Method `#withHostnameVerification()` is available for all trust strategies. Verification is turned off by default.

### DIFF
--- a/driver/src/test/java/org/neo4j/driver/v1/tck/DriverComplianceIT.java
+++ b/driver/src/test/java/org/neo4j/driver/v1/tck/DriverComplianceIT.java
@@ -19,10 +19,11 @@
 package org.neo4j.driver.v1.tck;
 
 import cucumber.api.CucumberOptions;
+import org.junit.AfterClass;
 import org.junit.ClassRule;
 import org.junit.runner.RunWith;
 
-import java.io.IOException;
+import java.nio.file.Files;
 
 import org.neo4j.driver.v1.util.TestNeo4j;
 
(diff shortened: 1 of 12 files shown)


driver/src/test/java/org/neo4j/driver/v1/tck/DriverComplianceIT.java
```diff
@@ -37,8 +38,11 @@
     @ClassRule
     public static TestNeo4j neo4j = new TestNeo4j();
 
-    public DriverComplianceIT() throws IOException
+    @AfterClass
+    public static void aaa() throws Exception
```
ali-ince: maybe `tearDownClass()`?

ali-ince (approved):
Looks good. I only had one small naming comment.

driver/src/test/java/org/neo4j/driver/v1/tck/DriverComplianceIT.java
lutovich: Hehe, yes! Fixed now

PR #501, opened 2018-05-21
Improve handling of RESET and ACK_FAILURE errors

lutovich:
This PR makes driver close the network channel when either `RESET` or `ACK_FAILURE` receive an error response. Message `RESET` is sent before release of the channel back to the channel pool. It might fail if database experiences a fatal error (like `OutOfMemoryError`) or a network failure occurs. In case of such errors channel will be closed and returned to the pool, which will drop it later. Message `ACK_FAILURE` is sent to acknowledge a `FAILURE` response. Failure to do so is unrecoverable and should result in a connection being dropped.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/handlers/ResetResponseHandler.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/handlers/ResetResponseHandler.java
@@ -39,13 +39,13 @@ public class ResetResponseHandler implements ResponseHandler
     @Override
     public final void onSuccess( Map<String,Value> metadata )
     {
-        resetCompleted();
+        resetCompleted( true );
     }
 
     @Override
     public final void onFailure( Throwable error )
     {
-        resetCompleted();
+        resetCompleted( false );
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


driver/src/main/java/org/neo4j/driver/internal/handlers/ResetResponseHandler.java
```diff
@@ -54,13 +54,13 @@ public final void onRecord( Value[] fields )
...
-    private void resetCompleted()
+    private void resetCompleted( boolean success )
     {
         messageDispatcher.unMuteAckFailure();
-        resetCompleted( completionFuture );
+        resetCompleted( completionFuture, success );
     }
 
-    protected void resetCompleted( CompletableFuture<Void> completionFuture )
+    protected void resetCompleted( CompletableFuture<Void> completionFuture, boolean success )
```
ali-ince, line 63: Do we loose information without using the `RESET` outcome, i.e. `success` parameter?

ali-ince (commented):
LGTM. I made one comment, which was not directly obvious to me.

driver/src/main/java/org/neo4j/driver/internal/handlers/ResetResponseHandler.java
lutovich: This response handler is used only for `Session#reset()` which is a best-effort call to clear the connection state on the database side. It might succeed, fail or do nothing regardless of a network error. That is why I thought it is okay to simply notify about the completion of the remote call here. `Session#reset()` is also deprecated :)

Does this make sense?

PR #507, opened 2018-06-20
Configurable server address resolver

lutovich:
This PR makes it possible to configure a resolver function used by the routing driver. Such function is used during the initial discovery and when all known routers have failed. Driver already had an internal facility like this which performed a DNS lookup for the hostname of the initial address. This remains the default. Users are now able to provide a custom resolver in the config.


zhenlineo (commented):
@nigelsmall Is this resolve function only available for routing driver or also applies to direct driver?

technige:
There is no reason to only restrict it to routing. There are (a few) use cases where it could be useful for direct too.

PR #511, opened 2018-07-02
Replaced ACK_FAILURE with RESET

technige:


driver/src/main/java/org/neo4j/driver/internal/handlers/ResetResponseHandler.java
```diff
@@ -56,8 +56,11 @@ public final void onRecord( Value[] fields )
 
     private void resetCompleted( boolean success )
     {
-        messageDispatcher.unMuteAckFailure();
-        resetCompleted( completionFuture, success );
+        messageDispatcher.clearCurrentError();
+        if (completionFuture != null)
```
zhenlineo: Why this null check is added? Did we changed as old behaviour here? 
Why this future could be null and is it safe to not have an `else` here?

zhenlineo (commented):
It is really happy to see `muteAckFailure` removed!
I got one question after reading this PR. 
Finally you might notice that your formatting is a bit different from the existing code.

driver/src/test/java/org/neo4j/driver/v1/integration/TransactionAsyncIT.java
```diff
@@ -1157,7 +1157,7 @@ public void shouldFailToCommitWhenRunFailureIsConsumed()
         }
         catch ( ClientException e )
         {
-            assertThat( e.getMessage(), startsWith( "Transaction rolled back" ) );
+            assertThat( e.getMessage(), startsWith( "No current transaction to commit." ) );
```
lutovich: This message change is quite weird. It exposes an implementation detail that driver resets a connection after an error. Do you think it is okay?

driver/src/main/java/org/neo4j/driver/internal/async/inbound/InboundMessageDispatcher.java
```diff
@@ -140,13 +132,14 @@ public void handleFailureMessage( String code, String message )
...
-            // fire error event back to the pipeline and avoid sending ACK_FAILURE
+            // fire error event back to the pipeline and avoid sending RESET
             channel.pipeline().fireExceptionCaught( currentError );
             return;
         }
 
-        // try to write ACK_FAILURE before notifying the next response handler
-        ackFailureIfNeeded();
+        // write a RESET to "acknowledge" the failure
+        queue( new ResetResponseHandler( this, null ) );
```
lutovich: Maybe better to create a constructor that only takes an `InboundMessageDispatcher` and uses `null` inside `ResetResponseHandler`. Just to not pass `null` explicitly as the second parameter.

lutovich (commented):
@nigelsmall review completed. Made two comments. It would also be nice to add a test that asserts valid errors are thrown in an explicit transaction after a failed run. Smth like this for blocking and async API:

```java
        
tx.run( "RETURN invalid" ).consume(); // should throw syntax error
tx.run( "RETURN 1" ).consume(); // should throw some meaningful error
        
tx.close();
```

PR #514, opened 2018-07-19
Refactoring to allow easier request message versioning

lutovich:
* Make connection more protocol version independent
 `Connection` interface now exposes methods that take ` Message` objects instead of statement and parameters for the `RUN` message

* Allow easier request message versioning
 Replaced double dispatch approach with a set of configurable message encoders. Later allows easier addition of new message types because double dispatch was locked to specific messages.
 Introduced `ValuePacker` and `ValueUnpacker` components responsible for reading and writing structs, like node, relationship, date-time or point. Previously, logic of handling structs was coupled with message writer/reader.


driver/src/main/java/org/neo4j/driver/internal/messaging/ValuePacker.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.neo4j.driver.v1.Value;
+
+public interface ValuePacker
+{
+    void packStructHeader( int size, byte signature ) throws IOException;
+
+    void pack( String string ) throws IOException;
+
+    void pack( Value value ) throws IOException;
```
zhenlineo, line 32: This method seems not used outside `ValuePackerV1`. Maybe not needed to be on the interface?

driver/src/main/java/org/neo4j/driver/internal/messaging/ValueUnpacker.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.neo4j.driver.v1.Value;
+
+public interface ValueUnpacker
+{
+    long unpackStructHeader() throws IOException;
+
+    int unpackStructSignature() throws IOException;
+
+    Value unpack() throws IOException;
```
zhenlineo: Similarly this one is also not used outside `ValueUnpackerV1`

driver/src/test/java/org/neo4j/driver/internal/messaging/KnowledgeablePackStreamMessageFormat.java
```diff
@@ -32,24 +54,51 @@
 
 /**
  * This class provides the missing server side packing methods to serialize Node, Relationship and Path.
+ * It also allows writing of server side messages like SUCCESS, FAILURE, IGNORED and RECORD.
  */
-public class KnowledgeablePackStreamMessageFormat extends PackStreamMessageFormatV2
+public class KnowledgeablePackStreamMessageFormat extends MessageFormatV2
```
zhenlineo: Rename to `KnowledgeableMessageFormat` ?

zhenlineo (approved):
Looks good.

driver/src/main/java/org/neo4j/driver/internal/messaging/ValuePacker.java
lutovich: Method is used in `RecordMessageEncoder`

PR #515, opened 2018-07-23
Create BoltProtocol abstraction

lutovich:
To encapsulate all Bolt protocol version specific operations under a single interface. This should make it easier to introduce Bolt V3 which contains multiple new messages.


driver/src/main/java/org/neo4j/driver/internal/messaging/BoltProtocol.java
```diff
@@ -0,0 +1,145 @@
...
+    void initializeChannel( String userAgent, Map<String,Value> authToken, ChannelPromise channelInitializedPromise );
+
+    /**
+     * Begin an explicit transaction.
+     *
+     * @param connection the connection to use.
+     * @param bookmark the bookmark. Never null, should be {@link Bookmark#empty()} when absent.
+     * @return a completion stage completed when transaction is started or completed exceptionally when there was a failure.
+     */
+    CompletionStage<Void> beginTransaction( Connection connection, Bookmark bookmark );
```
zhenlineo, line 66: This works fine for V1 and V2, but to make it work better with future versions, maybe we want each methods to take a `TransactionContext` instead? Also it will stop me from asking myself why some of methods require tx but some of them not :P

driver/src/main/java/org/neo4j/driver/internal/async/HandshakeHandler.java
```diff
@@ -108,36 +104,55 @@ protected void decode( ChannelHandlerContext ctx, ByteBuf in, List<Object> out )
         int serverSuggestedVersion = in.readInt();
         log.debug( "S: [Bolt Handshake] %d", serverSuggestedVersion );
 
-        ChannelPipeline pipeline = ctx.pipeline();
         // this is a one-time handler, remove it when protocol version has been read
-        pipeline.remove( this );
+        ctx.pipeline().remove( this );
 
-        switch ( serverSuggestedVersion )
+        BoltProtocol protocol = protocolForVersion( serverSuggestedVersion );
```
zhenlineo, line 110: Is it on purpose to NOT persist the protocol version globally? Are you thinking of driver talking to different members in cluster with different protocols? Is that a case that we should think of?

zhenlineo (approved):
PR looks good. Two questions that come into my mind when reading this PR.

driver/src/main/java/org/neo4j/driver/internal/messaging/BoltProtocol.java
lutovich: True, transaction is a bit confusing. It is only needed for commit to set the bookmark. I'll work on V3 soon and try to clean this up.

driver/src/main/java/org/neo4j/driver/internal/async/HandshakeHandler.java
lutovich: Yeah, driver might also talk to different single instance databases because of DNS.

PR #517, opened 2018-07-26
Bolt V3

lutovich:
Added new message types and logic for handling them. New metadata like `tx_timeout` and `tx_metadata` are not used yet.

Renamed `NettyConnection` to `DirectConnection` and `Bookmark` to `Bookmarks`.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/handlers/PullAllResponseHandler.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/handlers/PullAllResponseHandler.java
@@ -31,7 +31,7 @@ import org.neo4j.driver.internal.spi.Connection;
 import org.neo4j.driver.internal.spi.ResponseHandler;
 import org.neo4j.driver.internal.util.Futures;
 import org.neo4j.driver.internal.util.Iterables;
-import org.neo4j.driver.internal.util.MetadataUtil;
+import org.neo4j.driver.internal.util.MetadataExtractor;
 import org.neo4j.driver.v1.Record;
 import org.neo4j.driver.v1.Statement;
 import org.neo4j.driver.v1.Value;
@@ -53,7 +53,7 @@ public abstract class PullAllResponseHandler implements ResponseHandler
 
     private final Statement statement;
     private final RunResponseHandler runResponseHandler;
-    private final String resultConsumedAfterMetadataKey;
+    private final MetadataExtractor metadataExtractor;
     protected final Connection connection;
 
     // initialized lazily when first record arrives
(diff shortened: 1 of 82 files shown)


driver/src/main/java/org/neo4j/driver/internal/handlers/PullAllResponseHandler.java
```diff
@@ -66,10 +67,11 @@
     private CompletableFuture<Record> recordFuture;
     private CompletableFuture<Throwable> failureFuture;
 
-    public PullAllResponseHandler( Statement statement, RunResponseHandler runResponseHandler, Connection connection )
+    public PullAllResponseHandler( Statement statement, RunResponseHandler runResponseHandler, String resultConsumedAfterMetadataKey, Connection connection )
```
zhenlineo: I can see why you have this key here, 
but how about override this class and then have two implementation of `extractResultSummary` method? Versioning response handler sound a reasonable thing.

```
public ResultSummary extractResultSummary( Map<String,Value> metadata ) {
  long resultAvailableAfter = runResponseHandler.resultAvailableAfter();
  return MetadataUtil.extractSummary( statement, connection, resultAvailableAfter, metadata, T_LAST_KEY );
  // or RESULT_COMSUMED_AFTER_KEY for v1 and v2
}
```

Similarly for `RunResponseHandler`, we can override `onSuccess` or `onMeta`?

driver/src/main/java/org/neo4j/driver/internal/messaging/request/TransactionStartingMessage.java
```diff
@@ -0,0 +1,70 @@
...
+
+abstract class TransactionStartingMessage implements Message
+{
+    private static final String BOOKMARKS_METADATA_KEY = "bookmarks";
+    private static final String TX_TIMEOUT_METADATA_KEY = "tx_timeout";
+    private static final String TX_METADATA_METADATA_KEY = "tx_metadata";
+
+    final Map<String,Value> metadata;
+
+    TransactionStartingMessage( Bookmarks bookmarks, Duration txTimeout, Map<String,Value> txMetadata )
```
zhenlineo, line 39: aha, smart interface to take all transaction related metadata

driver/src/test/java/org/neo4j/driver/internal/handlers/HelloResponseHandlerTest.java
```diff
@@ -0,0 +1,135 @@
...
+        assertTrue( channel.closeFuture().isDone() ); // channel was closed
+    }
+
+    @Test
+    void shouldThrowWhenServerVersionCantBeParsed()
+    {
+        ChannelPromise channelPromise = channel.newPromise();
+        HelloResponseHandler handler = new HelloResponseHandler( channelPromise );
+
+        assertThrows( IllegalArgumentException.class, () -> handler.onSuccess( singletonMap( "server", value( "WrongServerVersion" ) ) ) );
```
zhenlineo, line 113: Is it still required to have the server version parsed rather than simply record it as a simple message?

zhenlineo (approved):

driver/src/main/java/org/neo4j/driver/internal/handlers/PullAllResponseHandler.java
lutovich: 👍 I'll give this a try. Might especially be valuable if we add another response metadata field

driver/src/test/java/org/neo4j/driver/internal/handlers/HelloResponseHandlerTest.java
lutovich: Server version should not be needed for version comparisons in Bolt V3. It is just easier to still keep it as `ServerVersion` object for older protocol versions

driver/src/main/java/org/neo4j/driver/internal/handlers/PullAllResponseHandler.java
lutovich: Made the metadata extraction utility versioned instead of passing string keys around. Creating classes felt a little to much - 3 new classes would need to be created just to override 2 string keys

PR #518, opened 2018-07-26
Expose transaction timeout and metadata in the API

lutovich:
This is the first draft of the API to allow transaction timeout and/or metadata be set for explicit and auto-commit transactions. Both settings can be specified in a `TransactionConfig` object which can be created like this:

```
TransactionConfig config = TransactionConfig.builder()
            .withTimeout(Duration.ofSeconds(5))
            .withMetadata(Collections.singletonMap("key", "value"))
            .build();
```

Explicit transactions can take config like this:

```
Transaction tx = session.beginTransaction(config);
```

and auto-commit transactions accept it in various overloads of `Session#run()` and `Session#runAsync()`:

```
session.run("CREATE ()", config);

session.runAsync("RETURN $x", Collections.singletonMap("x", 1) config);
```

Transaction functions accept config like this:

```
session.readTransaction(tx -> {...}, config);

session.writeTransactionAsync(tx -> {...}, config);
```

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/messaging/v1/BoltProtocolV1.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/messaging/v1/BoltProtocolV1.java
@@ -48,7 +48,9 @@ import org.neo4j.driver.internal.spi.ResponseHandler;
 import org.neo4j.driver.internal.util.Futures;
 import org.neo4j.driver.internal.util.MetadataExtractor;
 import org.neo4j.driver.v1.Statement;
+import org.neo4j.driver.v1.TransactionConfig;
 import org.neo4j.driver.v1.Value;
+import org.neo4j.driver.v1.exceptions.ClientException;
 
 import static java.util.concurrent.CompletableFuture.completedFuture;
 import static org.neo4j.driver.internal.async.ChannelAttributes.messageDispatcher;
@@ -86,8 +88,13 @@ public class BoltProtocolV1 implements BoltProtocol
     }
 
     @Override
-    public CompletionStage<Void> beginTransaction( Connection connection, Bookmarks bookmarks )
+    public CompletionStage<Void> beginTransaction( Connection connection, Bookmarks bookmarks, TransactionConfig config )
     {
+        if ( config != null && !config.isEmpty() )
+        {
+            return txConfigNotSupported();
+        }
+
         if ( bookmarks.isEmpty() )
         {
             connection.write(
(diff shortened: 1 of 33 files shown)


driver/src/main/java/org/neo4j/driver/internal/messaging/v1/BoltProtocolV1.java
```diff
@@ -86,8 +88,13 @@ public void initializeChannel( String userAgent, Map<String,Value> authToken, Ch
     }
 
     @Override
-    public CompletionStage<Void> beginTransaction( Connection connection, Bookmarks bookmarks )
+    public CompletionStage<Void> beginTransaction( Connection connection, Bookmarks bookmarks, TransactionConfig config )
     {
+        if ( config != null && !config.isEmpty() )
+        {
+            return txConfigNotSupported();
```
zhenlineo, line 95: This check is amazing! Finally, we do not need to check with server version anymore!

driver/src/main/java/org/neo4j/driver/v1/TransactionConfig.java
```diff
@@ -0,0 +1,218 @@
...
+ *                 .withMetadata(metadata)
+ *                 .build();
+ * }
+ * </pre>
+ *
+ * @see Session
+ */
+public class TransactionConfig
+{
+    private static final TransactionConfig EMPTY = builder().build();
```
zhenlineo, line 64: Since we now have this builder style, shall we also give Driver config a builder style?
So we deprecate the old method and then provides a new builder API. WDYT?

driver/src/test/java/org/neo4j/driver/v1/integration/TransactionBoltV3IT.java
```diff
@@ -0,0 +1,177 @@
...
+
+        try ( Session otherSession = session.driver().session() )
+        {
+            try ( Transaction otherTx = otherSession.beginTransaction() )
+            {
+                // lock dummy node but keep the transaction open
+                otherTx.run( "MATCH (n:Node) SET n.prop = 1" ).consume();
+
+                TransactionConfig config = TransactionConfig.builder()
+                        .withTimeout( ofSeconds( 1 ) )
```
zhenlineo: does it need 1s? Can we give it 1ms to speed up this test? This 1s might slow down my fancy new compute to run tests. (The minimal legal value server accept is 1ms)

Similar for the async one.

zhenlineo (approved):
Looks pretty good. Nice java doc

driver/src/test/java/org/neo4j/driver/v1/integration/TransactionBoltV3IT.java
lutovich: Fixed to keep your computer cool

driver/src/main/java/org/neo4j/driver/v1/TransactionConfig.java
lutovich: I would prefer to just change it in the next major release, tbh

PR #527, opened 2018-09-04
Keep bookmark returned after auto-commit tx

lutovich:
Auto-commit transactions return bookmarks in Bolt V3. They were previously unused. This PR makes driver extract such bookmarks and expose them via `Session#lastBookmark()`. It then possible for explicit and auto-commit transactions within the same session to be chained with bookmarks.


zhenlineo:
The concept of BookmarkHolder is really good. Wondering if it also makes sense to make Tx to get and set bookmark via BookmarkHolder too.

lutovich:
@zhenlineo I removed all the bookmark handling from transaction :)
`ExplicitTransaction` will now propagate bookmark directly to the session after commit.

PR #528, opened 2018-09-10
Ensure a single handler manages channel's auto-read

lutovich:
`PullAllResponseHandler` is responsible for receiving a stream of records. Stream might be very large and handler tries to not fetch if fully into memory. It applies a network-level backpressure by disabling/enabling auto-read property of the underlying Netty channel. This property tells channel to not read from the network. Auto-read is disabled when number of queued records goes beyond the threshold.

Management of auto-read turned out to be problematic for nested queries within a single transaction. Nested queries result in multiple `PullAllResponseHandler`s being added to the queue of handlers. They will try to manage auto-read concurrently and can sometimes disable it completely. Callers would then be blocked and unable to proceed. This is not a problem for `Session#run()` because every such call ensures there was a logical SYNC and the previous query has completed and its result is buffered.

This PR fixes a problem by making only a single installed handler manage the channel's auto-read property. Making sure there only exists a single such handler is the responsibility of `InboundMessageDispatcher`. Auto-read is enabled when new auto-read managing handler is installed. Previous such handler is disabled.

Also renamed `InboundMessageDispatcher#queue()` to `InboundMessageDispatcher#enqueue()`.

Fixes https://github.com/neo4j/neo4j/issues/12033

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/async/inbound/InboundMessageDispatcher.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/async/inbound/InboundMessageDispatcher.java
@@ -265,11 +265,11 @@ public class InboundMessageDispatcher implements MessageHandler
     private ResponseHandler removeHandler()
     {
         ResponseHandler handler = handlers.remove();
-        if ( autoReadManagingHandler == handler )
+        if ( handler == autoReadManagingHandler )
         {
-            // handler that is being removed is the auto-read managing handler
+            // the auto-read managing handler is being removed
             // make sure this dispatcher does not hold on to a removed handler
-            autoReadManagingHandler = null;
+            updateAutoReadManagingHandler( null );
         }
         return handler;
     }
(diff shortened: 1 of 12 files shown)


driver/src/main/java/org/neo4j/driver/internal/async/inbound/InboundMessageDispatcher.java
```diff
@@ -63,6 +66,7 @@ public void queue( ResponseHandler handler )
         else
         {
             handlers.add( handler );
+            updateAutoReadManagingHandlerIfNeeded( handler );
```
ali-ince, line 69: Even I couldn't think of a side effect, do you think that not cleaning `autoReadManagingHandler` after `handlers.remove()` would cause a problem?

ali-ince (commented):
looks good @lutovich. I only had one comment regarding the cleanup of field `autoReadManagingHandler`.

driver/src/main/java/org/neo4j/driver/internal/async/inbound/InboundMessageDispatcher.java
lutovich: Hm, I can't really see if there is a problem. However, it is probably incorrect to not clear `autoReadManagingHandler` field because `InboundMessageDispatcher` is pooled together with the channel. It's kind of strange to pool it together with some obsolete auto-read managing handler. Should be fixed in [this commit](https://github.com/neo4j/neo4j-java-driver/pull/528/commits/78cf1e3b5d27ee24862d0ec6811e41d83f3c95dd). Thanks!

ali-ince (approved):

PR #534, opened 2018-10-02
Add ability to create graph drivers fromUri with a builder

javatarz:
Allows consumers to keep a single code path to instantiate the driver for causal clusters and single node deployments.

1. If a protocol is specified, it is always used.
2. If only one URI is passed and no protocol is specified, the driver defaults to the bolt protocol.
3. If multiple URIs are passed and no protocol is specified, the driver uses the bolt+routing protocol.

### DIFF
--- a/driver/pom.xml
+++ b/driver/pom.xml
@@ -86,6 +86,10 @@
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+    </dependency>
   </dependencies>
 
   <build>
--- a/driver/src/main/java/org/neo4j/driver/v1/GraphDatabase.java
+++ b/driver/src/main/java/org/neo4j/driver/v1/GraphDatabase.java
@@ -18,13 +18,16 @@
  */
 package org.neo4j.driver.v1;
 
-import java.net.URI;
-
 import org.neo4j.driver.internal.DriverFactory;
 import org.neo4j.driver.internal.cluster.RoutingSettings;
 import org.neo4j.driver.internal.retry.RetrySettings;
 import org.neo4j.driver.v1.exceptions.ServiceUnavailableException;
 
+import java.net.URI;
+import java.util.List;
+
+import static java.util.Arrays.stream;
+import static java.util.stream.Collectors.toList;
 import static org.neo4j.driver.internal.DriverFactory.BOLT_ROUTING_URI_SCHEME;
 
 /**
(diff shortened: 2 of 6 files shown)


zhenlineo:
Thanks very much for your suggestion. We will discuss your suggested API and come back on this.

javatarz:
# Why do we need a single method to connect to single node instances and causal clusters

Our dev environments are run against single node neo4j instance. Our test environment works off a small causal cluster and our production environment works off a large causal cluster.

As a result, the connection modes used on the environments are bolt, bolt+routing and bolt+routing respectively. The lack of availability of an API in the driver that works across both these protocols requires client applications to contain an inherent code branch (if condition) that works based on the environment (or more specifically, the number of instances passed in the URL).

The `GraphDriver#routing` implementation does not support `bolt+routing`. The `GraphDriver#routingDriver` implementation allows `bolt` connections to causal clusters but does not allow bolt connections to single node instances (due to the internal call to the routing table). 

While both of these choices make sense, having the ability for an implementation to connect to clusters and single node instances is useful for enterprise users continuously delivering software across multiple environments.

## Stylistic implementation choices

A builder style implementation has been used for `GraphDriver#fromUri` for readability. This implementation is a slight departure from the current GraphDriver method implementations and I'm up for changing that if others want the style to remain consistent 😃

technige:
Hi @javatarz - thanks for your suggestion.

We recommend as a general rule that applications use the `GraphDatabase#driver` method for constructing a `Driver` object. This can accept either a `bolt:` or `bolt+routing:` URI, which would typically come from an application configuration file. For `bolt:`, the URI address would be the logical or physical address of a single server; for `bolt+routing:` this address would be the _logical_ address of the entry point of a database service that exposes the routing protocol, generally a Causal Cluster.

The `GraphDatabase#routingDriver` method is a handy construct for the few cases where it's not feasible to provide a single logical address for the cluster cores, perhaps due to DNS or load balancing limitations, and so the physical core addresses can be supplied instead. It's not intended for use in the general case as it is somewhat at odds with the general philosophy of the separation of logical and physical addressing. Furthermore, the individual addresses are not retained for the fallback bootstrap sequence.

It's also worth mentioning that the URI schemes should not be directly conflated with backend topology. While there is a strong correlation between `bolt:`/single instance and `bolt+routing:`/causal cluster, there is not a 1-to-1 mapping. You can use, for example, `bolt:` to access a cluster member directly. Consider them instead to be behavioural selectors for the `Driver` object.

You may find useful the _custom resolver_ functionality that is being introduced with the 1.7 drivers (available now in RC). This allows provision of a callback function at `Driver` construction time which can be used to implement "application level DNS". In this way, you can "resolve" the URI address into one or more derived addresses, which can be pulled from configuration or anywhere else.

We are continuing to work on URI format and `Driver` construction and have several internal design discussions ongoing. We are aware that some improvements can still be made here and expect to make some significant change in this regard with the 2.0 drivers.

So unfortunately we can't accept your PR, but I hope this explains why.

PR #539, opened 2018-10-09
Remove unnecessary restart of server to speed up test running

zhenlineo:


driver/src/test/java/org/neo4j/driver/v1/EncryptionTest.java
```diff
@@ -0,0 +1,192 @@
...
+    {
+        Encrypted,
+        Unencrypted,
+        Unknown,
+        FailedToRecognize
+    }
+
+    public final class DummyNettyServer
+    {
+        private EventLoopGroup bossGroup;
```
lutovich: Single event loop group can be used for the whole dummy server. There exists `ServerBootstrap#group(EventLoopGroup)` method and we most likely do not need separate groups in a test.

driver/src/test/java/org/neo4j/driver/v1/EncryptionTest.java
```diff
@@ -0,0 +1,192 @@
...
+            serverChannel = server.bind( 0 ).sync().channel();
+            return ((InetSocketAddress) serverChannel.localAddress()).getPort();
+        }
+
+        public void stop() throws Throwable
+        {
+            try
+            {
+                // Now shutting down
+                serverChannel.close().sync();
```
lutovich: I think it should be enough to just shutdown the event loop group. Channel gets terminated as part of the event loop group termination. `serverChannel` field might not be needed in this case.

driver/src/test/java/org/neo4j/driver/v1/EncryptionTest.java
```diff
@@ -0,0 +1,192 @@
...
+
+            ServerBootstrap server = new ServerBootstrap();
+            server.group( bossGroup, workerGroup ).channel( NioServerSocketChannel.class )
+                    .childHandler( new ChannelInitializer<SocketChannel>()
+            {
+                @Override
+                public void initChannel( SocketChannel ch ) throws Exception
+                {
+                    ChannelPipeline p = ch.pipeline();
+                    p.addLast( new ByteToMessageDecoder()
```
lutovich, line 116: `ReplayingDecoder` can be used here to not care about reading exactly 5 bytes.

driver/src/test/java/org/neo4j/driver/v1/EncryptionTest.java
```diff
@@ -0,0 +1,192 @@
...
+        Unknown,
+        FailedToRecognize
+    }
+
+    public final class DummyNettyServer
+    {
+        private EventLoopGroup bossGroup;
+        private EventLoopGroup workerGroup;
+        private Channel serverChannel;
+        private final AtomicReference<ClientSocketStatus> clientStatus = new AtomicReference<>( ClientSocketStatus.Unknown );
```
lutovich: Can this be a volatile field? It is expected to be written once, I think.

driver/src/test/java/org/neo4j/driver/v1/EncryptionTest.java
```diff
@@ -0,0 +1,192 @@
...
+            {
+                clientStatus.compareAndSet( ClientSocketStatus.Unknown, ClientSocketStatus.Unencrypted );
+            }
+            else
+            {
+                clientStatus.compareAndSet( ClientSocketStatus.Unknown, ClientSocketStatus.FailedToRecognize );
+            }
+        }
+
+        public void assertClientEncryption( boolean isEncrypted )
```
lutovich: Maybe `private`?

driver/src/test/java/org/neo4j/driver/v1/util/NetworkUtil.java
```diff
@@ -0,0 +1,34 @@
...
+import java.net.ServerSocket;
+
+public class NetworkUtil
+{
+    public static int getFreePort() throws IOException
+    {
+        ServerSocket s = new ServerSocket( 0 );
+        s.setReuseAddress( true );
+        int port = s.getLocalPort();
+        s.close();
```
lutovich: This can theoretically return an in-use port because of the `s.close()` :)

driver/src/test/java/org/neo4j/driver/v1/tck/ErrorReportingSteps.java
```diff
@@ -162,7 +163,7 @@ public void iRunANonValidCypherStatementItThrowsAnClientException() throws Throw
     @When( "^I set up a driver to an incorrect port" )
     public void iSetUpADriverToAnIncorrectPort() throws Throwable
     {
-        try ( Driver driver = GraphDatabase.driver( "bolt://localhost:7777" );
+        try ( Driver driver = GraphDatabase.driver( "bolt://localhost:" + getFreePort() );
```
lutovich: Maybe use port 80 of some other registered port that's definitely not Bolt?

PR #542, opened 2018-10-23
Fail correctly when connecting to a server with an unknown server identifier

technige:
While we currently parse the "server" string returned in the INIT metadata to extract the server version, no logic existed to validate the product part of that string. Indeed, an unexpected product name would have caused the internal regex to fail and an unhelpful error would surface. This PR raises a new `UntrustedServerException` in the case that the product string does not identify `Neo4j`.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/handlers/InitResponseHandler.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/handlers/InitResponseHandler.java
@@ -28,6 +28,7 @@ import org.neo4j.driver.internal.async.outbound.OutboundMessageHandler;
 import org.neo4j.driver.internal.spi.ResponseHandler;
 import org.neo4j.driver.internal.util.ServerVersion;
 import org.neo4j.driver.v1.Value;
+import org.neo4j.driver.v1.exceptions.UntrustedServerException;
 
 import static org.neo4j.driver.internal.async.ChannelAttributes.setServerVersion;
 
@@ -71,11 +72,25 @@ public class InitResponseHandler implements ResponseHandler
         throw new UnsupportedOperationException();
     }
 
-    private static ServerVersion extractServerVersion( Map<String,Value> metadata )
+    private static ServerVersion extractServerVersion( Map<String,Value> metadata ) throws UntrustedServerException
     {
         Value versionValue = metadata.get( "server" );
-        boolean versionAbsent = versionValue == null || versionValue.isNull();
-        return versionAbsent ? ServerVersion.v3_0_0 : ServerVersion.version( versionValue.asString() );
+        if ( versionValue == null || versionValue.isNull() )
+        {
+            return ServerVersion.v3_0_0;
+        }
+        else
+        {
+            ServerVersion server = ServerVersion.version(versionValue.asString());
+            if ( server.product().equalsIgnoreCase( "Neo4j" ) )
+            {
+                return server;
+            }
+            else
+            {
+                throw new UntrustedServerException( "Server does not identify as a genuine Neo4j instance" );
+            }
+        }
     }
 
     private static void updatePipelineIfNeeded( ServerVersion serverVersion, ChannelPipeline pipeline )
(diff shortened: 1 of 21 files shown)


driver/src/main/java/org/neo4j/driver/internal/handlers/InitResponseHandler.java
```diff
@@ -71,11 +72,25 @@ public void onRecord( Value[] fields )
         throw new UnsupportedOperationException();
     }
 
-    private static ServerVersion extractServerVersion( Map<String,Value> metadata )
+    private static ServerVersion extractServerVersion( Map<String,Value> metadata ) throws UntrustedServerException
```
lutovich: `throws` can be removed because `UntrustedServerException` is a runtime exception

driver/src/main/java/org/neo4j/driver/internal/handlers/InitResponseHandler.java
```diff
@@ -71,11 +72,25 @@ public void onRecord( Value[] fields )
...
+        else
+        {
+            ServerVersion server = ServerVersion.version(versionValue.asString());
+            if ( server.product().equalsIgnoreCase( "Neo4j" ) )
+            {
+                return server;
+            }
+            else
+            {
+                throw new UntrustedServerException( "Server does not identify as a genuine Neo4j instance" );
```
lutovich: Would be nice to include the received product string in the error message

driver/src/main/java/org/neo4j/driver/internal/util/ServerVersion.java
```diff
@@ -28,28 +29,35 @@
...
+    public static final ServerVersion v3_4_0 = new ServerVersion( "Neo4j", 3, 4, 0 );
+    public static final ServerVersion v3_2_0 = new ServerVersion( "Neo4j", 3, 2, 0 );
+    public static final ServerVersion v3_1_0 = new ServerVersion( "Neo4j", 3, 1, 0 );
+    public static final ServerVersion v3_0_0 = new ServerVersion( "Neo4j", 3, 0, 0 );
+    public static final ServerVersion vInDev = new ServerVersion( "Neo4j", Integer.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE );
 
     private static final String NEO4J_IN_DEV_VERSION_STRING = "Neo4j/dev";
     private static final Pattern PATTERN =
-            Pattern.compile( "(Neo4j/)?(\\d+)\\.(\\d+)(?:\\.)?(\\d*)(\\.|-|\\+)?([0-9A-Za-z-.]*)?" );
+            Pattern.compile( "([^/]*)/(\\d+)\\.(\\d+)(?:\\.)?(\\d*)(\\.|-|\\+)?([0-9A-Za-z-.]*)?" );
```
lutovich: This will match strings like "/3.5.0" where the product part is absent. It is okay?

driver/src/main/java/org/neo4j/driver/internal/util/ServerVersion.java
```diff
@@ -141,6 +142,10 @@ public boolean lessThanOrEqual(ServerVersion other)
 
     private int compareTo( ServerVersion o )
     {
+        if ( !product.equals( o.product ) )
+        {
+            throw new IllegalArgumentException("Comparing different products");
```
lutovich: Would be nice to include product names in the error message

driver/src/main/java/org/neo4j/driver/internal/util/ServerVersion.java
```diff
@@ -28,28 +29,35 @@
 
 public class ServerVersion
 {
-    public static final ServerVersion v3_5_0 = new ServerVersion( 3, 5, 0 );
-    public static final ServerVersion v3_4_0 = new ServerVersion( 3, 4, 0 );
-    public static final ServerVersion v3_2_0 = new ServerVersion( 3, 2, 0 );
-    public static final ServerVersion v3_1_0 = new ServerVersion( 3, 1, 0 );
-    public static final ServerVersion v3_0_0 = new ServerVersion( 3, 0, 0 );
-    public static final ServerVersion vInDev = new ServerVersion( Integer.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE );
+    public static final ServerVersion v3_5_0 = new ServerVersion( "Neo4j", 3, 5, 0 );
```
lutovich: Constant for string "Neo4j" would be nice

PR #548, opened 2018-11-30
Make channel auto-read management to work on routing driver

ali-ince:
This PR fixes an issue where the injected `RoutingResponseHandler` hides the auto-read management capabilities of the delegated `PullAllResponseHandler` and results in nested queries in a single transaction. The problem only applies to routing driver and was originally fixed with #528.

This PR removes the previously introduced `AutoReadManagingResponseHandler` interface and makes the capability a part of the `ResponseHandler` interface with default methods. Both `PullAllResponseHandler` and `RoutingResponseHandler` overrides the default implementation and the latter delegates the request to the wrapped handler.


lutovich (approved):

zhenlineo:
Nice finding. LGTM

PR #550, opened 2018-12-10
Made the code compile on java 11 and upgraded maven plugin dependencies

zhenlineo:
Fixed some errors in java docs when compiling with java 11.


driver/src/main/java/org/neo4j/driver/v1/Record.java
```diff
@@ -62,7 +62,7 @@
      * Check if the list of keys contains the given key
      *
      * @param key the key
-     * @return <tt>true</tt> if this map keys contains the given key otherwise <tt>false</tt>
+     * @return <code>true</code> if this map keys contains the given key otherwise <code>false</code>
```
michael-simons: I started to use the `{@code code}` macro quite often. Saves some chars.

PR #553, opened 2018-12-19
Removed the old tck, replaced with unit test instead.

zhenlineo:


ali-ince (approved):
LGTM.

@zhenlineo Could you please update version strings in pom.xml so that we can merge this into 1.7?

PR #557, opened 2019-01-21
Logging improvements:

PetrJanouch:
- each frame is logged only once
  - server address is now part of the logging message

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/logging/ChannelActivityLogger.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/logging/ChannelActivityLogger.java
@@ -20,11 +20,13 @@ package org.neo4j.driver.internal.logging;
 
 import io.netty.channel.Channel;
 
+import org.neo4j.driver.internal.BoltServerAddress;
 import org.neo4j.driver.internal.async.ChannelAttributes;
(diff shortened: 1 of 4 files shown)


driver/src/main/java/org/neo4j/driver/internal/logging/ChannelActivityLogger.java
```diff
@@ -70,4 +69,21 @@ private String getDbConnectionId()
...
+        if ( serverAddress == null )
+        {
+            BoltServerAddress serverAddress = ChannelAttributes.serverAddress( channel );
+            this.serverAddress = serverAddress != null ? serverAddress.toString() : null;
+        }
+
+        return serverAddress;
+    }
+
+    private String valueOrEmpty( String value )
```
ali-ince: Maybe move this into `org.neo4j.driver.internal.util.Format` class as a static method?

ali-ince (commented):
LGTM.

Only made a small comment on `valueOrEmpty` method's place.

driver/src/main/java/org/neo4j/driver/internal/logging/ChannelActivityLogger.java
PetrJanouch: Done.

PR #562, opened 2019-02-11
Use only BookmarksHolder contract instead of concrete NetworkSession.

michael-simons:
I'd fancy using only the necessary contract not only because I need this for the embedded connector, but because it reduces the exposed methods on that level.


michael-simons:
I've added a 2nd commit, that helps me to reuse the bookmark handling in future other session implementations.
As a side effect I think the intention of several tests are much better recognisable.

PR #564, opened 2019-02-12
Improve readability of compose-calls.

michael-simons:
This is a cosmetic issue, but it would help new readers like myself to understand what's actually going on.


zhenlineo:
@michael-simons According to this [resource](https://www.baeldung.com/java-completablefuture), your suggested change removes the enforced order: first we get the `cursorError`, then we release connection. 

According to the above resource, the `combine` applies in an different situation:  

> If you want to execute two independent Futures and do something with their results, use the thenCombine method that accepts a Future and a Function with two arguments to process both results`

michael-simons:
I'm closing this now due to not fully (yet) understood implications and errors. Thanks for the discussion, though.

PR #565, opened 2019-02-21
Add unresolved inital router address to list of resolved initial roter addresses

jjaderberg:
In scenarios with SNI, connecting to resolved addresses does not work.
The driver consistently fails to retrieve a routing table using the `initialRouter` because the address is explicitly resolved before making connection(s). It succeeds to get a routing table only because the same initial address is also put into the table as a router.

By adding the initial router address unresolved to the list of addresses to try, the driver should be able to work both with SNI and with multi-record DNS bootstrap addresses.


zhenlineo:
@jjaderberg Could you give a try of the fix proposed by https://github.com/neo4j/neo4j-java-driver/pull/566

PR #575, opened 2019-03-12
Driver New 2.0 API

zhenlineo:


driver/src/main/java/org/neo4j/driver/async/AsyncStatementRunner.java
```diff
@@ -0,0 +1,206 @@
...
+import java.util.concurrent.Executor;
+import java.util.function.Function;
+
+import org.neo4j.driver.Record;
+import org.neo4j.driver.Statement;
+import org.neo4j.driver.Value;
+import org.neo4j.driver.Values;
+
+/**
+ * Common interface for components that can execute Neo4j statements.
```
michael-simons: That's no longer true. "Common".

driver/src/main/java/org/neo4j/driver/internal/AbstractStatementRunner.java
```diff
@@ -21,19 +21,20 @@
...
 import org.neo4j.driver.StatementResult;
-import org.neo4j.driver.StatementResultCursor;
+import org.neo4j.driver.async.StatementResultCursor;
 import org.neo4j.driver.StatementRunner;
 import org.neo4j.driver.Value;
 import org.neo4j.driver.Values;
 import org.neo4j.driver.types.TypeSystem;
 
-public abstract class AbstractStatementRunner implements StatementRunner
+public abstract class AbstractStatementRunner implements StatementRunner, AsyncStatementRunner
```
michael-simons, line 37: I find this interesting: Bringing those two together again.
Where's the reactive statement runner in this context?

michael-simons (commented):
This is missing the reactive bits, right?
I wonder if the abstractstatementrunner than will implement all three…

driver/src/main/java/org/neo4j/driver/internal/AbstractStatementRunner.java
zhenlineo: `RxSession` takes a `NetworkSession` as input when creating. The `NetworkSession` indeed could implement `RxSession` too. But I feel it is not good to dump everything into one class. So the initial plan was to actually pull Blocking `Session` implementation out of `NetworkSession`. But as this PR only focus on making the API correct, so I did not split the implementation. Thus why this class implements two interfaces for example. Same for Session and Transaction.

zhenlineo: I also would suggest not use this internal class directly. If we split `NetworkSession` then we have two split this `AbstractStatementRunner` too.

michael-simons: I don't plan. 

I will use `StatementRunner` and `ReactiveStatementRunner` interfaces separatly. 

The platform plumbing we are writing will provide one or the other then. 

The user is supposed to use one of our yet to write "Templates". Those are Spring Datas low level operations.

Based on that, there will be repositories.

In any case: The user says give my reactive or imperative and he gets a different template and should ideally never see the driver.

ali-ince (approved):
Changes look good to me!

Shall we also add `neo4j` scheme to 1.7 (and maybe add a deprecation notice for `bolt+routing` scheme)?

Issue #571, opened 2019-03-06
Creating/closing a driver takes a long time


tobias-johansson:
- Neo4j version: Community Neo4j 3.5.3
- Neo4j Mode: Single instance
- Driver version: Java language driver 1.7.2
- Operating system: Ubuntu 18.04.2 LTS
- **Steps to reproduce**
1. Start local instance of Neo4j 3.5.3
2. Run the following class
```java
import org.neo4j.driver.v1.*;

public class DriverTest {
    public static void main(String[] args) {
        long start = System.currentTimeMillis();
        Driver driver = GraphDatabase.driver("bolt://localhost:7687", AuthTokens.basic("neo4j", "hej"));
        long created = System.currentTimeMillis();
        driver.close();
        long closed = System.currentTimeMillis();
        System.out.println("creation: " + (created - start) + "ms");
        System.out.println("closing: " + (closed - created) + "ms");
    }
}
```

- Expected behavior
```
creation: 14ms
closing: 22ms
```
- Actual behavior
```
creation: 1478ms
closing: 2248ms
```

eastlondoner:
Do you know what the latency between the driver and the neo4j server instance is? The authentication requires round-trips to the server which will depend very much on that

tobias-johansson:
@eastlondoner Not sure about creation, but it seems that closing always takes ~ 2000 ms, and that shouldn't require round-trips should it?

technige:
As of 1.7, a graceful shutdown sends a `GOODBYE` message from client to server. This indicates to the server that the client is deliberately closing the connection, as opposed to when an unexpected network event occurs. No response is sent back though, the server merely closes its side too.

Note that TCP has a graceful shutdown sequence too, so there will definitely be network activity on connection closure.

tobias-johansson:
@technige That makes sense, thanks! But still seems like we're waiting for some ~2000ms timeout on close right?
The long close times seem to always be present, also in very low latency setups (local docker neo4j and local jvm neo4j)

zhenlineo:
@tobias-johansson You can just call the async close `driver.closeAsync()` in your test environment. The blocking close instead will wait for 2s for a more "graceful" shutdown. In your test env, I do not think you need to worry about it.

tobias-johansson:
[...]

PR #576, opened 2019-03-14
Speeding up the shutdown of the driver.

PetrJanouch:


pstutz:
Thanks for looking into this, this helps with our issue https://github.com/neo4j/neo4j-java-driver/issues/571.

With a quiet period of 200ms the shutdown still takes about 1 second. 
With a not-so-graceful quiet period of 0ms the shutdown is almost instantly.

What important work wouldn't get done if the shutdown wasn't graceful? Might it make sense to either set it to 0ms or to offer a `shudownNow()` method for users that can live with a possible drawback?

PetrJanouch:
@pstutz Unfortunately, we cannot set the quiet period to 0ms right now. It seems that there are code paths that can result to some close or clean up tasks being submitted to the event loop after shutdown is called on the event loop. I think the ideal solution is to make sure that calling shutdown on the event loop is really the absolutely last step in the shut down routine. We could set the quiet period to 0ms then. Unfortunately, this is not as easy as it sounds, because of the asynchronous and event driven nature of that code. It will require revisiting and carefully ordering the shutdown logic. We will see what we can do. shutdownNow() might result in some resources not being closed (the closing events not being processed by the event loop)

driver/src/main/java/org/neo4j/driver/internal/async/pool/ConnectionPoolImpl.java
```diff
@@ -165,7 +166,10 @@ public int idleConnections( BoltServerAddress address )
             }
             finally
             {
-                eventLoopGroup().shutdownGracefully();
+                // This is an attempt to speed up the shut down procedure of the driver
+                // Feel free return this back to shutdownGracefully() method with default values
+                // if this proves troublesome!!!
+                eventLoopGroup().shutdownGracefully(200, 15_000, TimeUnit.MILLISECONDS);
```
technige, line 172: What does this do? Add a timeout? What happens if the timeout expires... what is the user supposed to do?

I am very cautious about optimising for a very niche use case. We need to optimise for the general case, wherein the driver shuts down at application shutdown. Here, everything _must_ close; there is no other code path.

PetrJanouch: A short explanation how the shutdownGracefully works. It has 2 parameters ‘quiet period’ and  ‘timeout’. The algorithm that it uses is the following:

```
shutdownStart = now()
while (true) {
	Thread.sleep(quietPeriod)
	if (nothingWasSubmittedToTheEventLoopWhileSleeping()) {
           break
	}

	if (now() - shutdownStart > timeout) {
          // timeout
	  break
        }
}
```
The main purpose of this shutdown algorithm is to compete work that is in progress.
The default value for ‘quiet period’  are 2 seconds and 15 seconds for ‘timeout’

What do we need a quiet period for anyway? The current shutdown logic is like this:

```
channels.each (ch -> closeAsync())
shutdownEventLoopWithQuietPeriod()
```

The graceful shutdown with the quiet period is needed because closing a channel async generates events submitted to the event loop.

What we would need is something like this:
```
channels.each (ch -> closeAsync())
waitForAllChannelsToBeClosed()
shutdownEventLoopWithoutQuietPeriod()
```

As simple as the pseudo code above looks, the problem is that achieving waitForAllChannelsToBeClosed() with current Netty API is not that simple (at least was not the last time I checked)

The problem is that the shutdown currently takes 4s (the quiet period is usually reset once by the tasks generated by closing the channels) which is enough to make for instance lambda users or users with a lot of integration tests sad.

The reasoning behind my PR was that the defaults in Netty are set so it would work with all possible use cases the toolkit is used for, but 200ms quiet period is more that enough for us.

PR #582, opened 2019-03-22
Adding tx with built-in retries support for RxSession

zhenlineo:
Implemented `RxSession#readTransaction` and `RxSession#writeTransaction` using `Flux#retryWhen`.
Extracted `NetworkSession` to provide basic `runAsync` and `runRx` methods, allowing other session implementations (such as AsyncSession, RxSession) build on top of it.


driver/src/main/java/org/neo4j/driver/internal/async/ExplicitTransaction.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.concurrent.CompletionStage;
+
+import org.neo4j.driver.Statement;
+import org.neo4j.driver.TransactionConfig;
+import org.neo4j.driver.async.StatementResultCursor;
+import org.neo4j.driver.internal.Bookmarks;
+import org.neo4j.driver.internal.cursor.RxStatementResultCursor;
+import org.neo4j.driver.internal.spi.Connection;
+
+public interface ExplicitTransaction
```
ali-ince: I slightly got confused with the new naming scheme. Since both `NetworkSession` and `ExplicitTransaction` are core building blocks for the rest of the surfaced API, what do you think about renaming `ExplicitTransaction` to `NetworkTransaction` - just to form a relationship between the two?

ali-ince (commented):
LGTM. 

I had one naming suggestion against the internal classes.

driver/src/main/java/org/neo4j/driver/internal/async/ExplicitTransaction.java
zhenlineo: The name `NetworkTransaction` dose not fits the Transaction very much. Will it be more clear if I remove the interfaces and directly call the implementations `NetworkSession` and `ExplicitTransaction`? 

Then all sessions will be built on top of `NetworkSession` and all transactions are built on top of `ExplicitTransaction`.

ali-ince: :+1: I think that would help.

PR #586, opened 2019-04-10
Added more examples to show how to use the new React API

zhenlineo:
Examples for integrating the driver's reactive API with Project Reactor and RxJava.

### DIFF
--- a/examples/src/main/java/org/neo4j/docs/driver/RxTransactionFunctionExample.java
+++ b/examples/src/main/java/org/neo4j/docs/driver/RxTransactionFunctionExample.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright (c) 2002-2019 "Neo4j,"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


michael-simons (commented):
Looks good to me. I didn't find the docs referencing that, but maybe it's worth writing a short sentence about why / when it's appropriate using `using` vs. `usingWhen` on the publishers.

examples/src/test/java/org/neo4j/docs/driver/ExamplesIT.java
```diff
@@ -488,4 +491,106 @@ void testHostnameVerificationExample()
...
+            // create some 'Product' nodes
+            try ( Session session = neo4j.driver().session() )
+            {
+                session.run(
+                        "UNWIND ['Tesseract', 'Orb', 'Eye of Agamotto'] AS item " +
+                                "CREATE (:Product {id: 0, title: item})" );
+            }
+
+            // read all 'Product' nodes
+            List<String> titles = await( example.readProductTitlesReactor() );
```
michael-simons, line 510: I'd use the Project reactors `StepVerifier` for Reactors test and probably similar things if there any for RxJava2.

examples/src/main/java/org/neo4j/docs/driver/RxTransactionFunctionExample.java
```diff
@@ -0,0 +1,71 @@
...
+import org.neo4j.driver.summary.ResultSummary;
+
+public class RxTransactionFunctionExample extends BaseApplication
+{
+    public RxTransactionFunctionExample( String uri, String user, String password )
+    {
+        super( uri, user, password );
+    }
+
+    public Publisher<ResultSummary> printAllProductsReactor()
```
michael-simons: Maybe return the actual implementation here? In the end it's what we wanna demonstrate with the example. 
Would also fit better with the tests for the example.

zhenlineo:
There is currently no doc using this examples. Eventually, we shall have a section in docs talking about reactive and that's where these examples will be used. Currently the purpose of this PR is merely to show how the suggested use pattern looks like.

PR #588, opened 2019-04-16
Retry tx on no writer error

zhenlineo:
Added a stub test to verify that after a leader in a cluster is removed, the driver should be able to retry the tx on 
* writer no longer available error (connection with a writer breaks in the middle of a tx)
* server no longer reachable error (a broken writer is still included in routing table)
* no writer available error (no writer found in routing table)

Changed the error `SEVERE: ServiceUnavailableException: Connection to the database terminated.` which happens on the second error condition, to `WARN: SessionExpired` to avoid noise in logs.


ali-ince (approved):
:+1:

PR #591, opened 2019-04-17
Added database name in the ResultSummary.

zhenlineo:
This requires the server to stream `db` with database name in the result summary.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/util/MetadataExtractor.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/util/MetadataExtractor.java
@@ -23,6 +23,9 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
+import org.neo4j.driver.Statement;
+import org.neo4j.driver.Value;
+import org.neo4j.driver.exceptions.UntrustedServerException;
 import org.neo4j.driver.internal.Bookmarks;
 import org.neo4j.driver.internal.spi.Connection;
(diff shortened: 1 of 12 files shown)


driver/src/main/java/org/neo4j/driver/internal/util/MetadataExtractor.java
```diff
@@ -99,12 +102,26 @@ public long extractResultAvailableAfter( Map<String,Value> metadata )
...
+        return new InternalResultSummary( statement, serverInfo, dbInfo, extractStatementType( metadata ), extractCounters( metadata ), extractPlan( metadata ),
+                extractProfiledPlan( metadata ), extractNotifications( metadata ), resultAvailableAfter,
+                extractResultConsumedAfter( metadata, resultConsumedAfterMetadataKey ) );
     }
 
-    public Bookmarks extractBookmarks( Map<String,Value> metadata )
+    public static DatabaseInfo extractDatabaseInfo( Map<String,Value> metadata )
+    {
+        Value dbValue = metadata.get( "db" );
+        if ( dbValue == null || dbValue.isNull() || !dbValue.hasType( TYPE_SYSTEM.STRING() ) )
```
ali-ince: Shouldn't we throw `ProtocolException` if `!dbValue.hasType( TYPE_SYSTEM.STRING() )` holds?

ali-ince (approved):
looks good. I only think that we should throw `ProtocolException` on unexpected types in summary.

PR #592, opened 2019-04-18
Change how BoltServerAddress equality works

ali-ince:
BoltServerAddress was checking equality on  `Hostname`, `Resolved IP` and `Port` triplet and DnsResolver was causing loss of original `Hostname`.

This commit removed public access to resolved IP Address component of BoltServerAddress and removes it from the equality check, thus resulting in same `hostname:port` pairs equal whatever IP address they are resolved to.

Also removed DnsResolver since it returns a set of addresses (which would not allow same `hostname:port` pairs resolved to different IP addresses) and offloaded the dns resolution to take place just after custom address resolution (default is identity resolver) happens.


ali-ince:
For the record, we have discussed this PR with @zhenlineo and decided to merge it in. 

The only behavioural change visible from outside will be that if the user provides a `ServerAddressResolver` implementation through configuration, the returned `ServerAddress`es will go through DNS resolution as well (and that resolution will resolve all IP addresses registered which will be tried in turn until successful). _This only applies to initial routing table lookup and fallback to initial routing table when none of the known routers are accessible._

PR #594, opened 2019-04-25
Suppress subsequent connection errors into existing error

zhenlineo:
When a server is shutting down, it is possible that a transaction is killed in the middle, followed by the termination of connection. Before the connection is shutting down, the server might send an error message if there is at least one pending message sent by client. However the server will also ignore any further messages that are queued in the server inbound message queue. Thus when the client trying to send or reading more with the closed connection, the client will receive a connection terminated error.

This PR makes sure when the above situation happens, the client will throw the original error with connection error as suppressed error.


ali-ince (approved):
LGTM.

PR #597, opened 2019-05-06
Suppress `pool was closed` errors

zhenlineo:
In a routing driver, it is possible that a thread will try to acquire a connection from a pool which might have been closed by another thread.
This happens when thread A obtains a server S to establish a connection. While thread B triggers updating routing table and removes server S from the routing table and connection pool. Then when thread A process to acquire a connection with server S, then we will end up with this pool was closed error.
This race condition cannot be avoided due to the nature of multi-threading. However when it happens, we can catch the error and treat it as a normal IO Error.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/async/pool/ConnectionPoolImpl.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/async/pool/ConnectionPoolImpl.java
@@ -45,6 +45,9 @@ import org.neo4j.driver.internal.util.Futures;
 import org.neo4j.driver.v1.Logger;
 import org.neo4j.driver.v1.Logging;
 import org.neo4j.driver.v1.exceptions.ClientException;
+import org.neo4j.driver.v1.exceptions.ServiceUnavailableException;
+
+import static java.lang.String.format;
 
 public class ConnectionPoolImpl implements ConnectionPool
 {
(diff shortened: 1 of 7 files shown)


driver/src/main/java/org/neo4j/driver/internal/async/pool/ConnectionPoolImpl.java
```diff
@@ -226,6 +230,13 @@ private void processAcquisitionError( BoltServerAddress serverAddress, Throwable
                         "Unable to acquire connection from the pool within configured maximum time of " +
                         settings.connectionAcquisitionTimeout() + "ms" );
             }
+            else if ( pool.isClosed() )
+            {
+                // There is a race condition where a thread tries to acquire a connection while the pool is closed by another concurrent thread.
+                // Treat as failed to obtain connection for a direct driver. For a routing driver, this error should be retried.
+                throw new ServiceUnavailableException( format( "Connection pool for server %s is closed while acquiring a connection. ", serverAddress ),
```
ali-ince: The extra space after `a connection.` in the exception message seems to be a typo.

driver/src/main/java/org/neo4j/driver/internal/async/pool/NettyChannelPool.java
```diff
@@ -76,4 +77,16 @@ protected ChannelFuture connectChannel( Bootstrap bootstrap )
         } );
         return channelFuture;
     }
+
+    @Override
+    public void close()
+    {
+        closed = true;
+        super.close();
```
ali-ince: Don't know if `FixedChannelPool#close` is idempotent or not, but it would be better if we could make `closed` an `AtomicBoolean` and call `super.close()` only if it wasn't called before, via `closed.compareAndSet(false, true)`?

ali-ince (commented):
LGTM. Had a couple of comments.

PR #598, opened 2019-05-10
Relax discovery logic

ali-ince:
Currently, discovery process is very picky on errors related with discovery responses. This PR relaxes the existing logic, and treats those exception as a try-next-router action and logs those exceptions at warning level.


zhenlineo (approved):
I was looking at `RoutingProcedureClusterCompositionProvider#processRoutingResponse`. Your change is better. 👍

PR #601, opened 2019-05-13
Improved error message in leaked session detection

zhenlineo:
to reflect the changes to the resource management in driver.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/LeakLoggingNetworkSession.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/LeakLoggingNetworkSession.java
@@ -48,8 +48,8 @@ class LeakLoggingNetworkSession extends NetworkSession
         Boolean isOpen = Futures.blockingGet( currentConnectionIsOpen() );
         if ( isOpen )
         {
-            logger.error( "Neo4j Session object leaked, please ensure that your application" +
-                          "calls the `close` method on Sessions before disposing of the objects.\n" +
+            logger.error( "Neo4j Session object leaked, please ensure that your application " +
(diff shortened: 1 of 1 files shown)


driver/src/main/java/org/neo4j/driver/internal/LeakLoggingNetworkSession.java
```diff
@@ -49,7 +49,7 @@ private void logLeakIfNeeded()
         if ( isOpen )
         {
             logger.error( "Neo4j Session object leaked, please ensure that your application" +
-                          "calls the `close` method on Sessions before disposing of the objects.\n" +
+                          "fully consumes results in Sessions or explicitly calls `close` on Sessions before disposing of the objects.\n" +
```
ali-ince, line 52: Does java driver releases connection when the result consumption is complete? Having looked shortly, it seems that `NetworkSession#closeAsync` does the actual release work and it's not called from within the driver at all. Maybe worth checking?

And besides, maybe add ` ` (space) just before `fully`?

ali-ince (commented):
I only have a question to verify.

driver/src/main/java/org/neo4j/driver/internal/LeakLoggingNetworkSession.java
zhenlineo: Both `SessionPullAllHandler#success` (also `failure`) and `Session#closeAsync` will release connection. The `DirectConnection#release` will ensure we release the one and only connection once. If the result is fully consumed, the `SessionPullAllHandler` will first release the connection.

PR #607, opened 2019-06-28
Fix shading of Project Reactor.

michael-simons:
Project reactors package is just reactor.

We cannot use the internal TestUtils in our ExamplesIT, as the internal TestUtils now expose reactors types under the shaded coordinates.

### DIFF
--- a/examples/src/test/java/org/neo4j/docs/driver/ExamplesIT.java
+++ b/examples/src/test/java/org/neo4j/docs/driver/ExamplesIT.java
@@ -23,7 +23,7 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
 import org.junit.jupiter.api.parallel.Execution;
 import org.junit.jupiter.api.parallel.ExecutionMode;
-import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
 
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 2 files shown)


examples/src/test/java/org/neo4j/docs/driver/ExamplesIT.java
```diff
@@ -563,14 +565,17 @@ void testShouldRunRxTransactionFunctionExampleReactor() throws Exception
...
             try ( AutoCloseable ignore = stdIOCapture.capture() )
             {
-                ResultSummary summary = await( Flux.from( example.printAllProductsReactor() ).single() );
-                assertEquals( StatementType.READ_ONLY, summary.statementType() );
+                Mono.from( example.printAllProductsReactor() )
+                        .doAfterTerminate( () -> latch.countDown() )
+                        .subscribe( summary -> assertEquals( StatementType.READ_ONLY, summary.statementType() ) );
             }
 
+            latch.await( 1, TimeUnit.MINUTES );
```
zhenlineo, line 580: I am not sure why this latch is needed? The original `await` has a timeout, which ensures the test will not blocking forever. If you wonder why the test is not written in a proper reactive way it merely because this test's focus is not to verify reactive but the result is correct.

Another note: there is a difference between `Mono.from(manyPublisher)` and `Flux.from(manyPublisher).single()`. If the `manyPublisher` publishes more than one result, the first Mono code will success, the second Flux code will error.

michael-simons: The latch has the same purpose as the `await`, but we cannot use `await` from `TestUtils` as they expose Reactor Types from the shaded package. Concrete: `org.neo4j.docs.driver.RxTransactionFunctionExample#printAllProductsReactor` returns `reactor.core.publisher` which now clashes with `org.neo4j.driver.internal.shaded.*`

The `await` on the latch has the same timeout.

Regarding Mono and Flux: You're right, I'll fix this.

PR #587, opened 2019-04-15
Multi-databases support for routing driver

zhenlineo:
`GetRoutingTable` procedure now accepts an optional parameter `database` for fetching routing table for the specified database. To obtain routing table for default database, use `database = null` rather than `database = ''`.

`GetRoutingTable` procedure should be executed on system database as it is a DBMS procedure.

Basics:
`RoutingTableRegistry` keeps a concurrent map from database names to `RoutingTableHandler`.
`RoutingTableHandler` holds a `RoutingTable` (empty at creation) and handles removal of servers from routing table in case of connection errors.
`Discovery` provides discovery service. Given a routing table, it reads available routers and finds a connection from connection pool to perform discovery. If success, it updates the routing table and connection pool. If failed, it throws error.

`RoutingTableRegistry`, `Discovery` and `ConnectionPool` are shared resources by all threads in drivers. `RoutingTableHandler` and `RoutingTable` is (mostly) used by a single thread.

PR #609, opened 2019-07-01
Make connectivity verification not part of driver creation.

zhenlineo:
Based on top of #587


ali-ince (commented):
I would vote for skipping connectivity verification for all and leave it to the user to explicitly call `Driver.verifyConnectivity` or `Driver.verifyConnectivityAsync` at his will. 

I think there's no value of adding a specific configuration setting for this.

ali-ince (approved):
LGTM @zhenlineo.

Maybe we can add an explicit warning note that even if the `verifyConnectivity` throws an exception, the user should still call `Driver.close()` to clean-up resources.

PR #615, opened 2019-07-20
Used channel tracking: Increment idle instead of use after channel is created.

michael-simons:
I'm pretty sure that needs to be idle to be incremented. 

I'm playing with Netty 4.1.36 and there timing seems a bit different, and as a result when returning the used channel, an `IllegalStateException` is thrown.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/async/pool/NettyChannelTracker.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/async/pool/NettyChannelTracker.java
@@ -90,7 +90,7 @@ public class NettyChannelTracker implements ChannelPoolHandler
         log.debug( "Channel [%s] created. Local address: %s, remote address: %s",
                 channel.id(), channel.localAddress(), channel.remoteAddress() );
 
-        incrementInUse( channel );
+        incrementIdle( channel );
         metricsListener.afterCreated( serverAddress( channel ), creatingEvent );
 
         allChannels.add( channel );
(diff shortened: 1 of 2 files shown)


michael-simons:
I have added a 2nd commit that fixes the test.

Correct me if I'm wrong, but i think their assumptions have been wrong. I read the code of the channel tracker that "creating" is not "acquiring" and as such, a creation leads to a new idle channel, that turns into "in use" when acquired, which will be signaled via Netty mechanism (`public void channelAcquired( Channel channel ))`)

zhenlineo:
@michael-simons In the netty 4.1.22 implementation, the following code is called when acquire a connection.

```
    private Future<Channel> acquireHealthyFromPoolOrNew(final Promise<Channel> promise) {
        try {
            final Channel ch = pollChannel();
            if (ch == null) {
                // No Channel left in the pool bootstrap a new Channel
                Bootstrap bs = bootstrap.clone();
                bs.attr(POOL_KEY, this);
                ChannelFuture f = connectChannel(bs);    // calls `channelCreated` 
                if (f.isDone()) {
                    notifyConnect(f, promise);
                } else {
                    f.addListener(new ChannelFutureListener() {
                        @Override
                        public void operationComplete(ChannelFuture future) throws Exception {
                            notifyConnect(future, promise);
                        }
                    });
                }
                return promise;
            }
            EventLoop loop = ch.eventLoop();
            if (loop.inEventLoop()) {
                doHealthCheck(ch, promise);               // calls `channelAcquired`
            } else {
                loop.execute(new Runnable() {
                    @Override
                    public void run() {
                        doHealthCheck(ch, promise);
                    }
                });
            }
        } catch (Throwable cause) {
            promise.tryFailure(cause);
        }
        return promise;
    }
```

As you can see, after this call, the connection is directly in state `InUse`. Thus both `channelCreated` and `channelAcquired` should increase `InUse` counter.

It could be totally true that the later netty changed how this code works. I would like to re-visit this while upgrading netty version.

michael-simons:
Thanks for the explanation! 

If that's changed in Netty, I find it quite a big change for a small revision upgrade.

PR #618, opened 2019-07-23
Avoid a queue for shaded record publisher

zhenlineo:
By default, when creating a publisher via `Flux.create` a `BufferAsyncSink` will be used and a queue is created to buffer all data the client cannot process in time.
With reactor being shaded, the reactor users cannot benifit from this queue and will have to create this queue again when creating from driver's record publisher.
Thus we configure the driver to create a publisher without any active buffering. The application will be fully in charge of buffer consumed data instead.


michael-simons (commented):
I think I understand the intention (give the user control), but don't we lose the buffering, too?

Or are we avoiding double buffering (once inside shaded reactor and once outside, when for example our publisher is turned into a non-shaded flux)?

If it's the later, than I'm all for this change.

zhenlineo:
It is indeed to avoid double buffering.

PR #619, opened 2019-07-26
System database bookmarks.

zhenlineo:
Passing system bookmarks to `getRT` call for bolt v4 connections. The main problem to solve is to find the routing table for a newly created database. When a new database is created, the system database is updated and then the new database is reconciled into existence asynchronously. The system database bookmarks allow us to make a causal link from the created database to the updated routing tables that included it. 

System Bookmarks are only used if a routing table update is needed when creating a session. If there is already a valid routing table, then the driver has no problem to route to the specified database in a cluster.

Bookmark is changed from a `String` to `Bookmark` object.
Here is the story how this change happened. `Session#lastBookmark` need to be able to return one or multiple bookmark strings. Normally, if we get a bookmark from server, the bookmark is a single string. However if the bookmark is provided by a user, then the bookmark can be a list of strings. As `Session#lastBookmark` can be called any time, so the bookmark can be one or multiple bookmark strings. If we simply return a `List` of strings rather than a `Bookmark` object, then `SessionConfig#withBookmarks` has to accept a `List` of `List` of string. This API dose not look good. Thus we come to the conclusion to make a bookmark object.

In the driver implementation to support system bookmark, a `ConnectionContext` is added to carry this system bookmarks down to rediscovery service. It is kind like a `SessionConfig` but it carries current connection access mode rather than default access mode.


driver/src/main/java/org/neo4j/driver/internal/Bookmark.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * When starting a session with initial bookmarks, the first transaction will be ensured to run at least after
+ * the database is as up-to-date as the latest transaction referenced by the supplied bookmarks.
+ *
+ * Within a session, bookmark propagation is carried out automatically.
+ * Thus all transactions in a session including explicit and implicit transactions are ensured to be carried out one after another.
+ *
+ * To opt out of this mechanism for unrelated units of work, applications can use multiple sessions.
+ */
+public interface Bookmark
```
ali-ince: I think a placeholder interface doesn't help all use cases we could encounter - assume that the user puts bookmarks into a cache service to use in further requests - it was very easy to do so when we were returning a `String` value, but now has no means to get the actual value in a simple type.

I strongly suggest that we expose a means of getting a simple value out of `Bookmark` interface and provide a means of constructing one from that exposed simple value.

ali-ince (commented):
The overall PR looks good to me, with the exception of exposing bookmarks as a placeholder interface.

PR #620, opened 2019-08-06
Default to unencrypted connection on the driver.

zhenlineo:
Removed insecure TrustAllCerts SecurityPlan.
Default to always perform client host name verification when encryption is on.
Deleted some code used to support server version earlier than 3.3.0.

Update boltkit to install a self-signed cert for 4.0 servers after download.


ali-ince (commented):
@zhenlineo AFAIK, a user still is allowed to use `TRUST_ALL_CERTIFICATES` but our default value is desired to be `TRUST_SYSTEM_CA_CERTIFICATES`. 

Maybe worth confirming and acting accordingly?

zhenlineo:
@ali-ince You are right, fixing>>>

PR #622, opened 2019-08-22
`Transaction#commit` and `Transaction#rollback`

zhenlineo:
Replaced `Transaction#success` and `Transaction#failure`
Changed the behaviour of `AsyncTransaction#commitAsync` and `AsyncTransaction#rollbackAsync` to throw error when committing/rolling back second time.
A transaction can be allowed to commit or rollback once, but closed multiple times.
This bring the behaviour of commit and rollback the same as 4.0 core API, as well as all other language drivers.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/async/ExplicitTransaction.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/async/ExplicitTransaction.java
@@ -45,12 +45,6 @@ public class ExplicitTransaction
         /** The transaction is running with no explicit success or failure marked */
         ACTIVE,
 
-        /** Running, user marked for success, meaning it'll value committed */
-        MARKED_SUCCESS,
-
-        /** User marked as failed, meaning it'll be rolled back. */
-        MARKED_FAILED,
-
         /**
          * This transaction has been terminated either because of explicit {@link Session#reset()} or because of a
          * fatal connection error.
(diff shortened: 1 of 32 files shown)


driver/src/main/java/org/neo4j/driver/internal/async/ExplicitTransaction.java
```diff
@@ -130,7 +104,7 @@ else if ( state != State.COMMITTED && state != State.ROLLED_BACK )
     {
         if ( state == State.COMMITTED )
         {
-            return completedWithNull();
+            return failedFuture( new ClientException( "Can't commit, transaction has been committed" ) );
```
ali-ince, line 107: I'm definitely not an English guru, but I would find `Can't commit, because transaction has already been committed.` more descriptive. And it would be nice if we could standardise the error messages like using always `Can't` or `Cannot` in the same context.

driver/src/main/java/org/neo4j/driver/internal/async/ExplicitTransaction.java
```diff
@@ -152,7 +126,7 @@ else if ( state == State.ROLLED_BACK )
         }
         else if ( state == State.ROLLED_BACK )
         {
-            return completedWithNull();
+            return failedFuture( new ClientException( "Can't rollback, transaction has been rolled back" ) );
```
ali-ince, line 129: Same here, and also the following (unchanged) messages with regards to standardisation,

ali-ince (approved):
LGTM. I've a couple of comments regarding error messages.

PR #624, opened 2019-09-03
Improved java doc for metrics

zhenlineo:
Added Driver#isMetricsEnabled method
Removed snapshot methods on metrics.
Bound pool metrics with connection pool. If a pool is removed, then the pool's metrics is also removed.
Changed the connection pool metrics to be a Collection instead of a Map, as the keys are mutable while pool is added and removed.


driver/src/main/java/org/neo4j/driver/ConnectionPoolMetrics.java
```diff
@@ -124,15 +120,15 @@
     long totalAcquisitionTime();
```
michael-simons, line 120: I wonder if we should compute this or just delegate to a monitoring tool. Something like Prometheus or Kibana should be able to compute this, based on the available counters and gauges.

driver/src/main/java/org/neo4j/driver/ConnectionPoolMetrics.java
```diff
@@ -124,15 +120,15 @@
...
     /**
-     * The total time in milliseconds spent to establishing new socket connections since the pool is created.
+     * A counter to record the total time in milliseconds spent to establishing new socket connections since the pool is created.
+     * This number increases every time when a connection is established.
      * See {@link ConnectionPoolMetrics#created()} for the total amount of connections established since the pool is created.
      * The average connection time can be calculated using the code below:
      * <h2>Example</h2>
      * <pre>
      * {@code
-     * ConnectionPoolMetrics previous = ConnectionPoolMetrics.EMPTY;
```
michael-simons, line 133: See above.

driver/src/main/java/org/neo4j/driver/Driver.java
```diff
@@ -155,6 +155,14 @@
     @Experimental
     Metrics metrics();
 
+    /**
+     * Returns true if the driver metrics reporting is enabled via {@link Config.ConfigBuilder#withDriverMetrics()}, otherwise false.
+     *
+     * @return true if the metrics reporting is enabled.
+     */
+    @Experimental
```
michael-simons, line 163: Remove `@Experimental`?

```suggestion
    @Experimental
```

driver/src/main/java/org/neo4j/driver/Metrics.java
```diff
@@ -18,20 +18,17 @@
...
-
-    /**
-     * Returns a snapshot of this metrics.
-     * @return a snapshot of this metrics.
+     * Connection pool metrics records metrics of connection pools that are currently in use.
+     * As the connection pools are dynamically added and removed while the server topology changes, the metrics collection changes over time.
+     * @return Connection pool metrics for all current active pools.
      */
-    Metrics snapshot();
+    @Experimental
```
michael-simons, line 32:
```suggestion
```

driver/src/main/java/org/neo4j/driver/internal/async/connection/ChannelAttributes.java
```diff
@@ -52,6 +53,16 @@ public static void setConnectionId( Channel channel, String id )
         setOnce( channel, CONNECTION_ID, id );
     }
 
+    public static String poolId( Channel channel )
+    {
+        return get( channel, POOL_ID );
+    }
+
+    public static void setPoolId( Channel channel, String id )
```
michael-simons, line 61: Is that a method to be called by a user? Asking because `public`.

michael-simons (changes requested):
Looks good apart from the minor comments I had.

driver/src/main/java/org/neo4j/driver/Driver.java
zhenlineo: The metrics is still experimental as it is only available in java. We want to introduce metrics in all drivers with a unified API. As it is only made into java for now. We mark it experimental so that the signature might be changed if we need to adjust for a unified API in the future.

driver/src/main/java/org/neo4j/driver/ConnectionPoolMetrics.java
zhenlineo: I am not sure how to calculate this acquisition time with other existing status.
This method basically give the acquisition time. Exposing it in a totally number is just how we expose it. A user would typically be interested in average acquisition time, which I give out the code to compute in javadoc

Could you give me some an example how we could we deduce this acquisition time with other data? Monitoring tool is smart but it has to use some math to compute this value, right? How the math would looks like???

driver/src/main/java/org/neo4j/driver/Metrics.java
zhenlineo: Same as the other experimental marker.

michael-simons: Ok :)

driver/src/main/java/org/neo4j/driver/ConnectionPoolMetrics.java
michael-simons: I guess in Prometheus, this query method would do the trick:

https://prometheus.io/docs/prometheus/latest/querying/functions/#aggregation_over_time

The idea is that those tools just collect all the counters and gauges and do the aggregration etc. on that abstract level.

zhenlineo: No, this is the connection acquisition time. For example, during 1s, we have 4 connection acquire operation and their acquisition time are `1ms, 2ms, 1ms, 3ms`. Then we can use that tool to calculate the total acquisition time, average acquisition time etc. However there is no good way to give a user these acquisition time `1ms, 2ms, 1ms, 3ms`. 

Imagine, in order to give a user these values, then we need to give them a list. But if we give them a list then there are too many questions one can ask: How long should we keep an acquisition time in the list? How to record the time when a new number is captured? 

Thus we give them a sum of `1ms, 2ms, 1ms, 3ms`. If the user query every 1s, then the user would know that the sum has increased 7ms since last time he queried, and the total acquisition count has increased 4, so he can compute that the average of the acquisition time over 1s is around 2ms. 

I do not think a user can deduce this average acquisition time from other counters and gauges. We have gauge `acquiring` and counter `acquired`.  It is totally possible, when a user query every 1s, the `acquiring` is always 0. However the `acquired` increased by 4. I do not see a way to compute the average acquisition time.

michael-simons: I see. Thanks for the explanation, that makes sense. I think that I thought there was another way to get the individual acquisition time.

PR #625, opened 2019-09-05
Prevent NPE in SessionConfig#equals.

michael-simons:
While `SessionConfig.builder()` disallows the user to input the literal value of `null`as database name, the name is by default null and when not configured else, `SessionConfig` has literal `null` as value for the database.

Under this conditions, the `equals` method is broken and will throw an NPE in any case where the receiver of equals has no database name configured, i.e.:

`SessionConfig.builder().build().equals( SessionConfig.builder().build())`

This PR fixes the behaviour and adds tests.


zhenlineo (approved):
❤️

PR #630, opened 2019-09-23
Bookmark Parsing

zhenlineo:
Moved `Bookmark` into public API and then provided a string conversion method to convert string bookmarks to `Bookmark`


michael-simons (approved):
That looks great and is exactly what we need for SDN. Thanks.

PR #631, opened 2019-09-24
Update netty version to the latest.

zhenlineo:
## Connection Counters
In the updated netty pool,
`ChannelPoolHandler#channelCreated` is called at channel creation, and then `ChannelPoolHandler#channelAcquired` is called when the connection is borrowed out of the pool. Whereas in previous netty version, when a connection is created, then it is directly lent out of the pool without being acquired. Thus some changes are made to maintain the counters for in use and idle connections.

## Deadlock Issues Caused by Netty Blocking `ChannelPool#close`.

The new netty version provides both blocking pool close `FixedChannelPool#close` and async pool close `FixedChannelPool#closeAsync`.
The blocking close causes deadlock issues in our code.

One deadlock issue happens when `ConnectionPoolImpl#retainAll` trying to close a netty pool inside an IO thread that is used by one of the channels owned by the closing pool.

The deadlock happens because
1) The channel cannot be closed until the IO thread is fully free.
2) The pool cannot be closed until the channel is closed.
3) However the IO thread cannot be free until the pool is closed.
`pool --wait-on--> channel --wait-on--> IO thread --wait-on--> pool`

The fix is to change 3) to let the IO thread to call async pool close to not block the close operation.

Another deadlock issue happens in `Driver#close`.

In `Driver#close`, we need to chain the event loop group shutdown after all pool close.
As pool close tasks are executed on event loop group, if we shutdown event loop group first,
then the pool close will be stuck because of no thread to pick up any shutdown tasks.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/async/pool/ConnectionPoolImpl.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/async/pool/ConnectionPoolImpl.java
@@ -21,8 +21,6 @@ package org.neo4j.driver.internal.async.pool;
 import io.netty.bootstrap.Bootstrap;
 import io.netty.channel.Channel;
 import io.netty.channel.EventLoopGroup;
-import io.netty.channel.pool.ChannelPool;
-import io.netty.util.concurrent.Future;
 
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
@@ -49,7 +47,7 @@ import org.neo4j.driver.internal.util.Futures;
 
 import static java.lang.String.format;
 import static org.neo4j.driver.internal.util.Futures.combineErrors;
-import static org.neo4j.driver.internal.util.Futures.completedWithNullIfNonError;
+import static org.neo4j.driver.internal.util.Futures.completeWithNullIfNoError;
 
 public class ConnectionPoolImpl implements ConnectionPool
 {
@@ -98,9 +96,9 @@ public class ConnectionPoolImpl implements ConnectionPool
 
         ListenerEvent acquireEvent = metricsListener.createListenerEvent();
         metricsListener.beforeAcquiringOrCreating( pool.id(), acquireEvent );
-        Future<Channel> connectionFuture = pool.acquire();
+        CompletionStage<Channel> channelFuture = pool.acquire();
 
-        return Futures.asCompletionStage( connectionFuture ).handle( ( channel, error ) ->
+        return channelFuture.handle( ( channel, error ) ->
         {
             try
             {
@@ -135,13 +133,7 @@ public class ConnectionPoolImpl implements ConnectionPool
                     {
                         log.info( "Closing connection pool towards %s, it has no active connections " +
                                   "and is not in the routing table registry.", address );
-                        // Close in the background
-                        closePool( pool ).whenComplete( ( ignored, error ) -> {
-                            if ( error != null )
-                            {
-                                log.warn( format( "An error occurred while closing connection pool towards %s.", address ), error );
-                            }
-                        } );
+                        closePoolInBackground( address, pool );
                     }
                 }
             }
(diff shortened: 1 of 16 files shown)


michael-simons:
Very happy to see this working under way.
I'll will add the additional steps needed to make it work with native compilation, once this is merged.

I'd love to see this and the additional steps into the final 4.0 release.

driver/src/main/java/org/neo4j/driver/internal/async/pool/ConnectionPoolImpl.java
```diff
@@ -156,37 +165,40 @@ public int idleConnections( BoltServerAddress address )
...
-                    log.info( "Closing connection pool towards %s", address );
-                    closePool( pool );
-                }
+            nettyChannelTracker.prepareToCloseChannels();
 
-                pools.clear();
-            }
-            finally
-            {
+            CompletableFuture<Void> allPoolClosedFuture = CompletableFuture.allOf(
```
lutovich: Maybe extract creation of this `CompletableFuture` in a method?

driver/src/main/java/org/neo4j/driver/internal/async/pool/ConnectionPoolImpl.java
```diff
@@ -156,37 +165,40 @@ public int idleConnections( BoltServerAddress address )
...
+                {
+                    completedWithNullIfNonError( closeFuture, pollCloseError );
+                }
+                else
+                {
                     // This is an attempt to speed up the shut down procedure of the driver
                     // Feel free return this back to shutdownGracefully() method with default values
                     // if this proves troublesome!!!
-                    eventLoopGroup().shutdownGracefully(200, 15_000, TimeUnit.MILLISECONDS);
+                    eventLoopGroup().shutdownGracefully( 200, 15_000, TimeUnit.MILLISECONDS );
```
lutovich: Maybe extract all event loop shutdown logic in a separate method? (both `shutdownGracefully` and `whenComplete` parts)

driver/src/main/java/org/neo4j/driver/internal/util/Futures.java
```diff
@@ -45,9 +45,27 @@ private Futures()
         return (CompletableFuture) COMPLETED_WITH_NULL;
     }
 
+    public static <T> CompletableFuture<T> completedWithNullIfNonError( CompletableFuture<T> future, Throwable error )
```
lutovich: I'd suggest changing the name slightly to `completeWithNullIfNoError`

driver/src/main/java/org/neo4j/driver/internal/async/pool/ExtendedChannelPool.java
```diff
@@ -20,9 +20,13 @@
 
 import io.netty.channel.pool.ChannelPool;
 
+import java.util.concurrent.CompletionStage;
+
 public interface ExtendedChannelPool extends ChannelPool
```
lutovich: Wdyt about removing `extends ChannelPool` and exposing only the methods `ConnectionPoolImpl` really needs?
For example, this would hide `close()` and `closeAsync()` so `ConnectionPoolImpl` would only have a possibility to call `repeatableCloseAsync()`

driver/src/main/java/org/neo4j/driver/internal/async/pool/ExtendedChannelPool.java
```diff
@@ -20,9 +20,13 @@
...
 {
     boolean isClosed();
 
     String id();
+
+    CompletionStage<Void> repeatableCloseAsync();
```
lutovich: Wdyt about calling this `destroyAsync`? I do not quite understand the "repeatable" part of the name :)

lutovich (commented):
Looks good, made a few really minor comments

lutovich (approved):

PR #634, opened 2019-09-27
Add support for native compilation.

michael-simons:
This commit adds the following:

- Oracle SVM, part of the Graal VM that allows substitutes for classes and methods, deleting them and aliasing them on the GraalVM.
- Substitutions to make Netty work natively with SSL. Those substitutions live under internal/svm and are not visible when used outside GraalVM

The PR also removes Nettys own native-image properties and replaces them with the setting needed for our driver.

Those settings cannot be easily rewritten using the Shade transformer, so it’s more simple to do it by hand.

The changes have been tested sucessfully on Graal 19.2.0.1 with Quarkus and Helidon microservices framework compiled natively.

### DIFF
--- a/driver/pom.xml
+++ b/driver/pom.xml
@@ -78,6 +78,10 @@
       <groupId>io.projectreactor</groupId>
       <artifactId>reactor-test</artifactId>
     </dependency>
+    <dependency>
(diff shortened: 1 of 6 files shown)


michael-simons:
Awesome 🎉, thanks for including it @zhenlineo

PR #637, opened 2019-10-10
Pulling in batches for async and blocking API with BOLT V4.

zhenlineo:
This provides a simple implementation to pull records in batches.
The idea is that when we start, we prefetch default fetch size (1000) records, once the driver receives `SUCCESS {has_more=true}`, then we automatically request a new batch from the server.

A driver user could cancel the streaming at any time by calling `StatementResult#summary` or `StatementRunner#close`. 

Changes to Driver API
* `StatementResult#summary` will actually behave like `StatementResult#consume`. It cancels streaming if the streaming has not been finished.
* `StatementResult#consume` is removed from public API.
* Unconsumed records in `StatementResult` cannot be accessed once `StatementRunner#close` is called.

Nested queries with `StatementRunner#run` is still supported.


driver/src/main/java/org/neo4j/driver/StatementResult.java
```diff
@@ -138,29 +138,12 @@
...
-     * @return a summary for the whole query result
-     */
-    ResultSummary consume();
-
     /**
      * Return the result summary.
      *
-     * If the records in the result is not fully consumed, then calling this method will force to pull all remaining
-     * records into buffer to yield the summary.
+     * If the records in the result is not fully consumed, then calling this method will exhausts the result.
```
michael-simons, line 144: Maybe clarify that exhausting means exhausting the current batch, not the whole possible result set / size.

driver/src/main/java/org/neo4j/driver/internal/InternalDriver.java
```diff
@@ -164,10 +164,10 @@ private static RuntimeException driverCloseException()
         return new IllegalStateException( "This driver instance has already been closed" );
     }
 
-    public NetworkSession newSession( SessionConfig parameters )
+    public NetworkSession newSession( SessionConfig config )
```
michael-simons, line 167: Unrelated, but good catch 👍

driver/src/main/java/org/neo4j/driver/internal/FailableCursor.java
```diff
@@ -22,5 +22,6 @@
 
 public interface FailableCursor
 {
+    CompletionStage<Throwable> consumeAsync();
     CompletionStage<Throwable> failureAsync();
```
michael-simons, line 26: Is this still needed? At least from what I read, you using `consumeAsync` now where `failureAsync` on the failable cursor has been used before.

I found that the failable cursor is used as interface for the `RxStatementResultCursor` as well… Hmm, it feels like that area could need some unification as well.

driver/src/main/java/org/neo4j/driver/internal/async/NetworkSession.java
```diff
@@ -194,7 +196,7 @@ public boolean isOpen()
                 if ( cursor != null )
                 {
                     // there exists a cursor with potentially unconsumed error, try to extract and propagate it
-                    return cursor.failureAsync();
+                    return cursor.consumeAsync();
```
michael-simons, line 199: This is one of the places I mentioned above… 
Wouldn't that change be applicable to other places where `failureAsync` is used?

driver/src/main/java/org/neo4j/driver/internal/async/ResultCursorsHolder.java
```diff
@@ -74,6 +74,6 @@ private static Throwable findFirstFailure( CompletableFuture<Throwable>[] comple
     {
         return cursorStage
                 .exceptionally( cursor -> null )
-                .thenCompose( cursor -> cursor == null ? completedWithNull() : cursor.failureAsync() );
+                .thenCompose( cursor -> cursor == null ? completedWithNull() : cursor.consumeAsync() );
```
michael-simons, line 77: Same.

driver/src/main/java/org/neo4j/driver/internal/cursor/AsyncStatementResultCursorImpl.java
```diff
@@ -16,37 +16,36 @@
...
 import org.neo4j.driver.internal.handlers.PullAllResponseHandler;
 import org.neo4j.driver.internal.handlers.RunResponseHandler;
 import org.neo4j.driver.internal.util.Futures;
-import org.neo4j.driver.internal.cursor.InternalStatementResultCursor;
-import org.neo4j.driver.Record;
-import org.neo4j.driver.exceptions.NoSuchRecordException;
 import org.neo4j.driver.summary.ResultSummary;
 
-public class AsyncStatementResultCursor implements InternalStatementResultCursor
+public class AsyncStatementResultCursorImpl implements AsyncStatementResultCursor
```
michael-simons, line 34: Big 👍

driver/src/main/java/org/neo4j/driver/internal/handlers/pulln/BasicPullResponseHandler.java
```diff
@@ -18,38 +18,262 @@
...
+ * |                    | DONE | FAILED | STREAMING                      | READY              | CANCELED       |
+ * |--------------------|------|--------|--------------------------------|--------------------|----------------|
+ * | request            | X    | X      | toRequest++ ->STREAMING        | PULL ->STREAMING   | X              |
+ * | cancel             | X    | X      | ->CANCELED                     | DISCARD ->CANCELED | ->CANCELED     |
+ * | onSuccess has_more | X    | X      | ->READY request if toRequest>0 | X                  | ->READY cancel |
+ * | onSuccess          | X    | X      | summary ->DONE                 | X                  | summary ->DONE |
+ * | onRecord           | X    | X      | yield record ->STREAMING       | X                  | ->CANCELED     |
+ * | onFailure          | X    | X      | ->FAILED                       | X                  | ->FAILED       |
+ *
+ * Currently the error state (marked with X on the table above) might not be enforced.
```
michael-simons, line 54: `DONE` is an error state?

driver/src/main/java/org/neo4j/driver/internal/handlers/pulln/PullResponseHandler.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Register a summary consumer to be notified when a summary is received.
+     * STREAMING shall not be started before this consumer is registered.
+     * A null summary with no error indicates a SUCCESS message with has_more=true has arrived.
+     * @param summaryConsumer register a summary consumer
+     */
+    void installSummaryConsumer( BiConsumer<ResultSummary,Throwable> summaryConsumer );
+
+    enum Status
```
michael-simons, line 47: It would be useful to find the state transitions here.

michael-simons (commented):
General remark: I like the removal of `consume`. It will break things for people, but will eventually be much more clear than the dualism of `summary()` and `consume()`. 

Offering a higher level method like the following

```
	@Override
		public ResultSummary run() {

			try (AutoCloseableStatementRunner statementRunner = getStatementRunner(this.targetDatabase)) {
				StatementResult result = runnableStatement.runWith(statementRunner);
				return result.consume();
			}
		}
```

will not be longer possible, right?

And would need to look like this

```
@Override
		public ResultSummary run() {

			try (AutoCloseableStatementRunner statementRunner = getStatementRunner(this.targetDatabase)) {
				StatementResult result = runnableStatement.runWith(statementRunner);
				while(result.hasNext()) {
					result.next();
				}
				return result.summary();
			}
		}
```

or doing it with `.list()` and drop the returned values.

Having said that, I would like an overloaded `summary` taking in a `boolean` to indicate to actually not only exhaust the current batch, but everything.
But than, why removing `consume()`?

driver/src/main/java/org/neo4j/driver/internal/FailableCursor.java
zhenlineo: The different of these two method:
consumeAsync: Discard all unconsumed records and return if any error for the whole execution and streaming. Used in `StatementRunner#close`, where after this boundary, all records are discarded.
failureAsync: Buffer all unconsumed records into memory and return if any error for the whole execution and streaming. Used for nested queries between `session#run`s, so that the second run will not discard all previous unconsumed run records.

driver/src/main/java/org/neo4j/driver/StatementResult.java
zhenlineo: No, it will exhausting the whole result. It is equivalent to `Subscription#cancel`.

We currently cannot discard N, we can only discard ALL.

The following code is equivalent to a driver user:
```
// Code with 1.7 driver
public ResultSummary run() {
	try (AutoCloseableStatementRunner statementRunner = getStatementRunner(this.targetDatabase)) {
		StatementResult result = runnableStatement.runWith(statementRunner);
		return result.consume();
	}
}

// Code with 4.0 driver
public ResultSummary run() {
	try (AutoCloseableStatementRunner statementRunner = getStatementRunner(this.targetDatabase)) {
		StatementResult result = runnableStatement.runWith(statementRunner);
		return result.summary();
	}
}
```

driver/src/main/java/org/neo4j/driver/internal/handlers/pulln/BasicPullResponseHandler.java
zhenlineo: If it is Done, then there is no further transition you can go from Done state. 

The top row of the table defines what state you are currently in. Then the first column defines what action you can perform.

I know this is a bit shaky. I will add a card to improve this class with a proper state machine as you pointed out.

PR #641, opened 2019-10-28
Fix duplicated fields on ProfiledPlan

zhenlineo:
The current PR removes duplicated keys from fields in `ProfiledPlan`.

If we want to remove duplicated keys from `argument` map instead, then we shall add missing fields `pageCacheHits`, `pageCacheMisses`, `pageCacheHitRatio`, `time`.


zhenlineo:
https://github.com/neo4j/neo4j-java-driver/pull/642

PR #649, opened 2019-11-17
Simplify bookmark impl by exposing internal values

zhenlineo:
This makes serialization of the bookmark easier to end users.
End users could serialize and deserialize bookmark in whatever way they like using `Set<String> values`

Bookmark is defined as follows:
```
Bookmark {
Set<String> values(); // can be used to serialize bookmark
static Bookmark parse(Set<String> values); // can be used to deserialize bookmark
boolean isEmpty();
}
```


michael-simons (changes requested):
I'd remove `static Bookmark merge(Iterable<String> bookmarks);` to avoid confusion.  It's not needed with `.parse` and `.merge` in combination and clutters API (at least to me, maybe that's personal taste).

I'm also a bit confused as I thought we spoke about dumping the bookmarks as `byte[]` from the driver.

zhenlineo:
@michael-simons Very good input on the `merge` methods. Would you think this API will be good enough for your use case if we have `merge` methods removed?

We talked about `toBytes`, but it is not very idiomatic to Javascript, where most of objects are directly visible to users. .Net driver gives this solution which gives me some new thoughts. The original purpose to hide Bookmark implementation was to prevent users from using the bookmark content, however if users (OGM and/or Fabric) need to use the content anyway, I see fewer points to hide it.

PetrJanouch (approved):

michael-simons:
Thanks, Zhen. I'd only remove the `static Bookmark merge(Iterable<String> bookmarks);` 

With the Strings is totally fine with me this way.

michael-simons (approved):
LGTM! Thanks for your work on that.

PR #652, opened 2019-11-25
Config option to allow config netty event loop threads

zhenlineo:
Added configuration option to allow users to config how many netty event loop threads they want to use.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/Config.java
+++ b/driver/src/main/java/org/neo4j/driver/Config.java
@@ -94,13 +94,14 @@ public class Config
     private final int routingFailureLimit;
     private final long routingRetryDelayMillis;
     private final long fetchSize;
-    private long routingTablePurgeDelayMillis;
+    private final long routingTablePurgeDelayMillis;
 
     private final int connectionTimeoutMillis;
     private final RetrySettings retrySettings;
     private final ServerAddressResolver resolver;
 
     private final boolean isMetricsEnabled;
+    private final int eventLoopThreads;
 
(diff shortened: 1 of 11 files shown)


PetrJanouch (approved):
Nice!

elR1co:
Hello. Thanks a lot for the work done.
I am currently using version 1.7.5. In my application, I need to use a separate pool for a service, in order to restrict the number of connections, only for this case. In this case, I have two separate Driver instances, one for my generic pool, and one for this specific service. I noticed that every time I create a Driver instance, there are 2 * CPUs Neo4jDriverIO threads which are created, and in my case where I would just need maybe 5 connections max, I would not need more than 2 threads for example.
My question is : does this feature will solve my problem ? Will it be also merged to 1.7 branch soon ?

Thanks a lot again.

zhenlineo:
Hi @elR1co,
Yes, this PR will allow you to set the thread count used by netty thread pool used by this driver. You will be able to give 2 threads per driver. However currently there is no plan to back port this feature into 1.7. Our 4.0 driver is ready for a final release soon. The 4.0 driver will support all in-support Neo4j versions. Though you might need to change some code as we broke some public API.

zhenlineo:
Hi @elR1co,
The default value is `2 * availableProcessors`. The 0 we used is an implementation detail. Deep down the stack, Netty will modify it to `2 * availableProcessors` if it is 0. The constant is not exposed by Netty, thus we will not support the computation of processor count either. If you are interested, you can have a look at `MultithreadEventLoopGroup`.

PR #661, opened 2019-12-10
Refactored BasicPullResponseHandler to use an explicit state machine

gjmwoods:
This PR refactors `BasicPullResponseHandler` to make its state machine explicit.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/handlers/pulln/BasicPullResponseHandler.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/handlers/pulln/BasicPullResponseHandler.java
@@ -40,18 +40,7 @@ import static org.neo4j.driver.internal.handlers.pulln.FetchSizeUtil.UNLIMITED_F
 import static org.neo4j.driver.internal.messaging.request.DiscardMessage.newDiscardAllMessage;
 
 /**
- * In this class we have a hidden state machine.
- * Here is how it looks like:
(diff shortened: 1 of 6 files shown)


driver/src/main/java/org/neo4j/driver/internal/handlers/pulln/BasicPullResponseHandler.java
```diff
@@ -267,13 +229,217 @@ private void dispose()
         this.summaryConsumer = null;
     }
 
-    protected Status status()
+    protected State getState()
```
zhenlineo: The driver uses naming style `void status(Status s)`, and `Status status()`. This is because when we are tying to give a unified API across all languages, `getXX` and `setXX` does not apply in many languages such as python or Js. So we decided we use getter and setters in this driver without `get` and `set`. I would suggest to change the name back so that we keep the naming style in the whole codebase to be consistent.

driver/src/main/java/org/neo4j/driver/internal/handlers/pulln/BasicPullResponseHandler.java
```diff
@@ -267,13 +229,217 @@ private void dispose()
...
     {
-        return this.status;
+        return state;
     }
 
-    protected void status( Status status )
+    protected void setState( State state )
```
zhenlineo: Same for the setter.

zhenlineo (approved):
Looks pretty good.

PR #662, opened 2019-12-13
Allow simple/async sessions to request data as needed

gjmwoods:
Optimizing record buffer for `AutoPullResponseHandler`. Simple and Async sessions will now buffer records in a similar way to the reactive sessions by only requesting more records when the Query's consumer is close to catching up. This prevents excessive records being buffered by the driver/client.

The driver will request additional records when the current buffer contains less than 30% of the fetchSize (`LOW_RECORD_WATERMARK`). When the driver has buffered more than 70% of the fetchSize (`HIGH_RECORD_WATERMARK`), the driver will stop requesting more records until it drops below the `LOW_RECORD_WATERMARK`. A `fetchSize` of zero effectively disables this buffering.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/handlers/pulln/AutoPullResponseHandler.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/handlers/pulln/AutoPullResponseHandler.java
@@ -49,12 +49,15 @@ public class AutoPullResponseHandler extends BasicPullResponseHandler implements
 {
     private static final Queue<Record> UNINITIALIZED_RECORDS = Iterables.emptyQueue();
     private final long fetchSize;
+    private final long lowRecordWatermark;
+    private final long highRecordWatermark;
 
     // initialized lazily when first record arrives
     private Queue<Record> records = UNINITIALIZED_RECORDS;
 
     private ResultSummary summary;
     private Throwable failure;
+    private boolean isAutoPullEnabled = true;
 
     private CompletableFuture<Record> recordFuture;
     private CompletableFuture<ResultSummary> summaryFuture;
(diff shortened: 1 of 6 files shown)


zhenlineo:
Can you have a test for `ResultCursor#listAsync`, where we define fetchSize = 10?
I am thinking in this situation. we will pre-pull 10 records locally. However as it exceed the `HIGH_RECORD_WATERMARK`, then we might not be able to send PULL(n=-1) until the user has consumed any records locally. The user will not be able to consume records locally as the user is waiting for all records arrive into a list.

driver/src/main/java/org/neo4j/driver/internal/handlers/pulln/AutoPullResponseHandler.java
```diff
@@ -64,6 +67,19 @@ public AutoPullResponseHandler(Query query, RunResponseHandler runResponseHandle
     {
         super(query, runResponseHandler, connection, metadataExtractor, completionListener );
         this.fetchSize = fetchSize;
+
+        //For pull everything ensure conditions for disabling auto pull are never met
+        if ( fetchSize == UNLIMITED_FETCH_SIZE )
+        {
+            this.HIGH_RECORD_WATERMARK = Long.MAX_VALUE;
+            this.LOW_RECORD_WATERMARK = Long.MAX_VALUE;
```
zhenlineo: The vars are not static, so they do not need to be capitalised.

driver/src/main/java/org/neo4j/driver/internal/handlers/pulln/AutoPullResponseHandler.java
```diff
@@ -198,11 +217,27 @@ private void enqueueRecord( Record record )
...
+        {
+            isAutoPullEnabled = false;
+        }
+
         records.add( record );
     }
 
     private Record dequeueRecord()
     {
+        if ( records.size() < LOW_RECORD_WATERMARK )
```
zhenlineo: "<=" for the case when fetchSize = 1. (or any case where the HIGH_RECORD_WATERMARK == LOW_RECORD_WATERMARK)

zhenlineo (commented):
A small case where fetchSize = 1 may not work as expected.

PR #672, opened 2020-02-05
use Synchronized List to fix flaky test

gjmwoods:


michael-simons (commented):
I would probably have used a `Collections.synchronizedList(new ArrayList())`, just to keep the list semantics.

PR #685, opened 2020-03-02
Extended URI Schemes

gjmwoods:
This PR makes encryption and trust settings configurable using the URI.

We introduce 4 additional schemes:
- `bolt+s` and `neo4j+s` which enables encryption and configures the trust settings to use the local system's certificate store
- `bolt+ssc` and `neo4j+ssc` which configures trust to use all certificates


driver/src/main/java/org/neo4j/driver/internal/DriverFactory.java
```diff
@@ -148,21 +133,23 @@ protected ChannelConnector createConnector( ConnectionSettings settings, Securit
...
             {
-            case BOLT_URI_SCHEME:
                 assertNoRoutingContext( uri, routingSettings );
                 return createDirectDriver( securityPlan, address, connectionPool, retryLogic, metricsProvider, config );
-            case BOLT_ROUTING_URI_SCHEME:
-                return createRoutingDriver( securityPlan, address, connectionPool, eventExecutorGroup, routingSettings, retryLogic, metricsProvider, config );
-            default:
-                throw new ClientException( format( "Unsupported URI scheme: %s", scheme ) );
             }
+            throw new ClientException( format( "Unsupported URI scheme: %s", scheme ) );
```
zhenlineo: haha. I am not fooled by these lines :D 
If-else should be good enough here as we've validated the scheme `Scheme.validateScheme( uri.getScheme() );`.

```
if( isRoutingScheme(scheme) ){...}
else {...}
```

driver/src/main/java/org/neo4j/driver/internal/SecuritySettings.java
```diff
@@ -0,0 +1,199 @@
...
+
+    private void assertSecuritySettingsNotUserConfigured( String uriScheme )
+    {
+        if ( isCustomized() )
+        {
+            throw new ClientException( String.format( "Scheme %s is not configurable with manual encryption and trust settings", uriScheme ) );
+        }
+    }
+
+    private SecurityPlan createSecurityPlanFromScheme( String scheme ) throws GeneralSecurityException, IOException
```
zhenlineo, line 90: Can we simplify this method as:
```
if ( isHighTrustScheme(scheme) )
{ 
    return SecurityPlanImpl.forSystemCASignedCertificates( trustStrategy.isHostnameVerificationEnabled() ); 
}
else 
{
    return SecurityPlanImpl.forAllCertificates( trustStrategy.isHostnameVerificationEnabled() );
}
```

We've already checked `isSecurityScheme` and `assertSecuritySettingsNotUserConfigured`.

zhenlineo (commented):
All looks good. Just a suggestion to simplify the code.

driver/src/main/java/org/neo4j/driver/internal/SecuritySettings.java
gjmwoods: of course! I will update this now

PR #698, opened 2020-04-23
Foundational Changes for Driver and Bolt 4.1

gjmwoods:
Creates baseline changes for 4.1 branch including:
- updating POM version
- Adding Bolt Protocol V4.1 (currently identical to 4)
- Refactoring protocol version to handle MAJOR.MINOR versions of the protocol

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/async/connection/BoltProtocolUtil.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/async/connection/BoltProtocolUtil.java
@@ -20,10 +20,12 @@ package org.neo4j.driver.internal.async.connection;
 
 import io.netty.buffer.ByteBuf;
 
+import org.neo4j.driver.internal.messaging.BoltProtocolVersion;
 import org.neo4j.driver.internal.messaging.v1.BoltProtocolV1;
 import org.neo4j.driver.internal.messaging.v2.BoltProtocolV2;
 import org.neo4j.driver.internal.messaging.v3.BoltProtocolV3;
 import org.neo4j.driver.internal.messaging.v4.BoltProtocolV4;
+import org.neo4j.driver.internal.messaging.v41.BoltProtocolV41;
 
 import static io.netty.buffer.Unpooled.copyInt;
 import static io.netty.buffer.Unpooled.unreleasableBuffer;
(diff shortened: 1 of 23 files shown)


driver/src/main/java/org/neo4j/driver/internal/async/connection/BoltProtocolUtil.java
```diff
@@ -20,32 +20,35 @@
...
 
 import static io.netty.buffer.Unpooled.copyInt;
 import static io.netty.buffer.Unpooled.unreleasableBuffer;
 import static java.lang.Integer.toHexString;
 
 public final class BoltProtocolUtil
 {
-    public static final int HTTP = 1213486160; //== 0x48545450 == "HTTP"
+    // Represents the `TP` part of `HTTP` (0x48545450) which can be returned if connected to HTTP port
+    public static final BoltProtocolVersion HTTP = new BoltProtocolVersion(80, 84 );
```
meistermeier:
```suggestion
    public static final BoltProtocolVersion HTTP = new BoltProtocolVersion( 80, 84 );
```
Also it feels at least a little bit hacky to use the `BoltProtocolVersion` for a `T,P`.
Maybe only as an inner class or similar of `HandshakeHandler` or a method in `BoltProtocolVersion` like `isHttp`.

meistermeier (commented):
There are already some removals of Bolt protocol V1 that might be necessary here and cannot wait until #700 got merged.
But on the other hand there is code here https://github.com/neo4j/neo4j-java-driver/pull/698/files#diff-b93e1004a9bd13ae99ba4ea26a24c802R528 , https://github.com/neo4j/neo4j-java-driver/pull/698/files#diff-cf7dd80310fc175afb1f0be7a76e7b21R155 and https://github.com/neo4j/neo4j-java-driver/pull/698/files#diff-cf7dd80310fc175afb1f0be7a76e7b21R155

driver/src/main/java/org/neo4j/driver/internal/async/connection/BoltProtocolUtil.java
gjmwoods: move this into a `isHttp()` method on `BoltProtocolVersion`

PR #700, opened 2020-05-11
Removing Bolt Versions 1 and 2 for 4.1 Driver

gjmwoods:
Since version 4.1 of the driver will no longer support Bolt V1 and V2 we remove unused code for those protocols.


michael-simons:
In general, looks good to me.

I wonder if we should take a day next week to consolidate message format and value (un)packers:

* Remove the v1 and v2
* Create unified (Un)Packer with the information from which bolt version types are available into the type itself.
* Create a factory for (Un)Packers that can be asked for an (Un)Packer supporting a bolt version. That factory would provide the unified unpacker with a enum set of the supported types
* Use that unified packers for a single message format. I'm sure the field keys and length of the message formats find a better home than the message format classes.

Only removing the protocol classes here feels incomplete.

michael-simons (commented):
In general, looks good to me.

I wonder if we should take a day next week to consolidate message format and value (un)packers:

* Remove the v1 and v2
* Create unified (Un)Packer with the information from which bolt version types are available into the type itself.
* Create a factory for (Un)Packers that can be asked for an (Un)Packer supporting a bolt version. That factory would provide the unified unpacker with a enum set of the supported types
* Use that unified packers for a single message format. I'm sure the field keys and length of the message formats find a better home than the message format classes.

Only removing the protocol classes here feels incomplete.

gjmwoods:
Closing as superseded by https://github.com/neo4j/neo4j-java-driver/pull/765

PR #701, opened 2020-05-13
4.1 add routing context to hello message

gjmwoods:
Include the routing context configured in the connection URI in the Hello message. This enables the server to make routing decisions with the same information provided to the driver.

For example with the URI `neo4j://localhost:7687?policy=fast&region=eu` the driver will send the map: `{'policy': 'fast', 'region': 'eu'}` in the Hello message's metadata with the key `routing`.

Since `bolt` is for direct connections the map is configured with `{'routing': null}`inform that the sever should not route queries.


michael-simons (approved):
👍 👍

PR #707, opened 2020-05-21
Removing Bolt V2 from handshake options

gjmwoods:


michael-simons (approved):
👍 Leave the headroom for the full refactoring.

PR #713, opened 2020-05-27
Fix JavaDoc on JDK 11. (#702)

gjmwoods:
- JDK 11 JavaDoc is less lenient with HTML tags inside a <code>-block, so this needed a fix.
- We don’t use java-modules, so all modules are undefined. Therefor the search is broken.

There are 2 solutions to the later:
1. Disabling modules in java-doc with `--no-module-directories`, but that also applies to references to JDK classes, which would than be broken.

2. (The one applied here): Check if `useModuleDirectories` is set to something different than literal `undefined`: If so, leave it and let JavaDocs JavaScript magic handle things, otherwise set it to false, causing JavaDocs JavaScript todo nothing. For reference: https://stackoverflow.com/a/57284322/1547989

### DIFF
--- a/driver/pom.xml
+++ b/driver/pom.xml
@@ -120,7 +120,17 @@
           <sourceFileIncludes>
             <sourceFileInclude>org/neo4j/driver/**/*.java</sourceFileInclude>
           </sourceFileIncludes>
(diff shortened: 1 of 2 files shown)


michael-simons (approved):
Unasked for, but I guess I like the PR :D

PR #729, opened 2020-06-07
Differentiate between transactions with open resources and committable transactions

gjmwoods:
Differentiate between a transaction which is open with respect to the resources it has and open in the sense it is in a committable state

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/async/UnmanagedTransaction.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/async/UnmanagedTransaction.java
@@ -92,7 +92,7 @@ public class UnmanagedTransaction
 
     public CompletionStage<Void> closeAsync()
     {
-        if ( isOpen() )
+        if ( isOpen() || state.equals( State.TERMINATED ))
         {
             return rollbackAsync();
         }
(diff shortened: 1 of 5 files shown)


driver/src/main/java/org/neo4j/driver/internal/async/UnmanagedTransaction.java
```diff
@@ -158,7 +158,13 @@ else if ( state == State.ROLLED_BACK )
 
     public boolean isOpen()
     {
-        return state != State.COMMITTED && state != State.ROLLED_BACK;
+        // TERMINATED is still considered open w.r.t to resources available
+        return state == State.ACTIVE || state == State.TERMINATED;
```
michael-simons, line 162: That doesn't actually change anything (ACTIVE, TERMINATED) are the inverse of (COMMITED;ROLLED_BACK).

gjmwoods: That was intended to make it more readable (plus makes the comment easier to make sense of). Happy to roll it back.

gjmwoods:
closing, no longer needed

PR #737, opened 2020-06-10
Don’t swallow the cause of a TX termination. (#731)

gjmwoods:
* Don’t swallow the cause of a TX termination.

There are a couple of scenarions in which a transaction gets terminated. The cause will be propagated by the pull handler and the transaction will be marked accordingly. There might be a chance that a degration from a leader to a follower on the server side happens in between to calls to a run method on that transaction: The transaction is still open, but cannot run queries any longer. Outside a transactional function this leads correctly to a ClientException. Inside a transactional function, this must not happen.

The retry logic must be able to find the cause of the termination and if there’s any, it should judge on the cause if it retries or not.

This PR changes the following:

- Add a StateHolder to the UnmangedTransaction
- The holder is necessary to keep the single field volatie
- The holder holds the state and a possible cause of termination
- The holder is able to determine whether a session is still open or not.
- It removes markTerminated from InternalAsyncTransaction as it was used only for tests.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/async/InternalAsyncTransaction.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/async/InternalAsyncTransaction.java
@@ -50,11 +50,6 @@ public class InternalAsyncTransaction extends AsyncAbstractQueryRunner implement
         return tx.runAsync(query, true );
     }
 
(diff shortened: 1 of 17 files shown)


michael-simons (approved):
LGTM ;) 👍

PR #740, opened 2020-06-10
Improve channel tracking.

michael-simons:
This uses a reentrant read/write lock to avoid synchronizing and a mixture of concurrent hashmaps and atomic integers.


gjmwoods (approved):
:thumbsup: :thumbsup:

PR #744, opened 2020-06-30
Aura example

mneedham:


mneedham:
This is how it'll look on the Aura console:

image

mneedham:
Updated screenshot:

image

examples/src/main/java/org/neo4j/docs/driver/DriverIntroductionExample.java
```diff
@@ -0,0 +1,115 @@
...
+
+    public static void main(String... args) throws Exception {
+        // Aura uses the "neo4j" protocol
+        String boltUrl = "%%BOLT_URL_PLACEHOLDER%%";
+
+        String user = "<Username for Neo4j Aura database>";
+        String password = "<Password for Neo4j Aura database>";
+
+        // Aura queries use an encrypted connection
+        Config config = Config.builder().withEncryption().build();
```
gjmwoods, line 107: just a thought about dropping this config in favor of `neo4j+s` scheme

gjmwoods (approved):
looks like a nice clear example to me. cheers!

examples/src/main/java/org/neo4j/docs/driver/DriverIntroductionExample.java
mneedham: I didn't know about that scheme! But yes we can use that. I've updated the example

PR #760, opened 2020-09-10
Rename tests end with BoltKitTest to BoltKitIT

bigmontz:
The BoltKitTests depends on the TeamCity environment with the new test environment be running to not be skipped,
so it should be moved to the integration tests phase avoiding being always skipped during the unit tests phase.

Rename it with the suffix `BoltKitIT` does this job.

### DIFF
--- a/driver/src/test/java/org/neo4j/driver/integration/RoutingDriverBoltKitIT.java
+++ b/driver/src/test/java/org/neo4j/driver/integration/RoutingDriverBoltKitIT.java
@@ -36,7 +36,6 @@ import java.util.TreeSet;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.Function;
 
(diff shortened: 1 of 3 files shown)


michael-simons (approved):
LGTM. I guess you needed to explicitly include the `*IT.java` and `*ITCase.java` because one entry would overwrite all the others, right?

bigmontz:
> LGTM. I guess you needed to explicitly include the `*IT.java` and `*ITCase.java` because one entry would overwrite all the others, right?

Exactly

2hdddg:
Would it be a bigger but more safe change to append IT to the boltkit tests?
We need to be sure that we dont lose any tests in the process (the test count in Teamcity can be used to verifiy)

bigmontz:
> Would it be a bigger but more safe change to append IT to the boltkit tests?
> We need to be sure that we dont lose any tests in the process (the test count in Teamcity can be used to verifiy)

I could named `BoltKitIT` instead of `BoltKitTest`. I also get a bit more explicit the intention in the code.

PR #764, opened 2020-09-29
Make canRetryOn function public.

meistermeier:
Make it static and rename it to isRetryable.
To be used as experimental API.


technige:
@meistermeier Yes, this seems fine. Both an internal package and an experimental annotation clarify that this still will not become part of the public API.

michael-simons:
For comparision @nigelsmall https://github.com/spring-projects/spring-data-neo4j/blob/master/src/main/java/org/springframework/data/neo4j/core/support/RetryExceptionPredicate.java

gjmwoods (approved):

PR #765, opened 2020-09-30
Remove protocol 1 and 2

bigmontz:
The reference this protocols are not needed since it's not used by the driver anymore.

During the remotion of these protocol version, the tests were refactoried to be specific per version instead of
assume the re-use done on the protocols implementation.


driver/src/test/java/org/neo4j/driver/internal/InternalResultTest.java
```diff
@@ -19,46 +19,42 @@
...
 import static java.util.Collections.emptyMap;
 import static java.util.Collections.singletonMap;
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.collection.IsCollectionWithSize.hasSize;
 import static org.hamcrest.junit.MatcherAssert.assertThat;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.*;
```
gjmwoods: Could you double-check the neo4j checkstyles is being applied? Pretty sure its doesn't use *. (Personally not fussed either way just makes reviews easier)

driver/src/test/java/org/neo4j/driver/internal/messaging/v42/MessageFormatV42Test.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * The MessageFormat under tests is the one provided by the {@link BoltProtocolV42} and not an specific class implementation.
+ *
+ * It's done on this way to make easy to replace the implementation and still getting the same behaviour.
+ *
+ */
+class MessageFormatV42Test
+{
+    private static final MessageFormat format = BoltProtocolV4.INSTANCE.createMessageFormat();
```
gjmwoods: `BoltProtocolV42`?

gjmwoods (commented):
Awesome stuff! Just double-check the check styles setup before merging.

driver/src/test/java/org/neo4j/driver/internal/InternalResultTest.java
bigmontz: I will check again.

bigmontz: It is, interesting it is not being applied correctly.

driver/src/test/java/org/neo4j/driver/internal/messaging/v42/MessageFormatV42Test.java
bigmontz: Yes, I will fix it.

driver/src/test/java/org/neo4j/driver/internal/InternalResultTest.java
bigmontz: I've reseted the commit, and then applied the reformatting on all the modified files.

gjmwoods (approved):

PR #768, opened 2020-10-05
Update dependencies

gjmwoods:


pom.xml
```diff
@@ -132,10 +132,24 @@
...
+      <dependency>
+        <groupId>com.fasterxml.jackson.core</groupId>
+        <artifactId>jackson-core</artifactId>
+        <version>2.11.3</version>
+      </dependency>
+      <dependency>
+        <groupId>com.fasterxml.jackson.core</groupId>
+        <artifactId>jackson-databind</artifactId>
+        <version>2.11.3</version>
+      </dependency>
```
bigmontz, line 146: Do we need Jackson for the driver it self? Because, if it's only needed for the test kit, it should be only on the test kit project.

bigmontz: My mistake, it's just the dependency management .

bigmontz (approved):
💯

PR #770, opened 2020-10-14
Creating protocol 4.3 and uses the Route Message to get the routing table

bigmontz:
The objective of the commit/pr is create the infrastructure needed to support the version 4.3 of bolt protocol, including the handshake and the serialisation and use of the ROUTE message to get the routing table.

In manner to support the version 4.3 of the protocol, the handshake was adjusted and the version 4.0 was removed from it.

Some adjustments was done into the RoutingProcedureRunner which now instead of be a base class with the implementation of the single database routing procedure runner,  it's the interface used by the runners.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/messaging/BoltProtocol.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/messaging/BoltProtocol.java
@@ -39,6 +39,7 @@ import org.neo4j.driver.internal.messaging.v3.BoltProtocolV3;
 import org.neo4j.driver.internal.messaging.v4.BoltProtocolV4;
 import org.neo4j.driver.internal.messaging.v41.BoltProtocolV41;
 import org.neo4j.driver.internal.messaging.v42.BoltProtocolV42;
+import org.neo4j.driver.internal.messaging.v43.BoltProtocolV43;
 import org.neo4j.driver.internal.spi.Connection;
 
 import static org.neo4j.driver.internal.async.connection.ChannelAttributes.protocolVersion;
@@ -167,6 +168,10 @@ public interface BoltProtocol
         else if ( BoltProtocolV42.VERSION.equals( version ) )
         {
             return BoltProtocolV42.INSTANCE;
+        }
+        else if ( BoltProtocolV43.VERSION.equals( version ) )
+        {
+            return BoltProtocolV43.INSTANCE;
         }
         throw new ClientException( "Unknown protocol version: " + version );
     }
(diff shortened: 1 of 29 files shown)


driver/src/test/java/org/neo4j/driver/integration/RoutingDriverBoltKitIT.java
```diff
@@ -613,6 +614,41 @@ void shouldHandleLeaderSwitchAndRetryWhenWritingInTxFunctionAsync() throws IOExc
         assertThat( writeServer.exitStatus(), equalTo( 0 ) );
     }
 
+    @Test
+    @Disabled("Disabled due the issues with the PR pipeline")
+    void shouldHandleLeaderSwitchAndRetryWhenWritingInTxFunctionAsyncV43() throws IOException, InterruptedException
```
gjmwoods: do we have the testkit test already? if so maybe we can remove this, otherwise we can leave it as a cleanup task for later

driver/src/main/java/org/neo4j/driver/internal/messaging/request/RouteMessage.java
```diff
@@ -0,0 +1,97 @@
...
+public class RouteMessage implements Message
+{
+    public final static byte SIGNATURE = 0x66;
+    private final Map<String,Value> routingContext;
+    private final String databaseName;
+
+    /**
+     * Constructor
+     *
+     * @param routingContext The routing context used for multi-data center deployments. It should be used in combination with multi-data center load
```
gjmwoods: nit: possibly routing context will have other applications beyond multi-data center in the future so we could be more generic here

driver/src/main/java/org/neo4j/driver/internal/messaging/BoltProtocol.java
```diff
@@ -168,6 +169,10 @@ else if ( BoltProtocolV42.VERSION.equals( version ) )
         {
             return BoltProtocolV42.INSTANCE;
         }
+        else if ( BoltProtocolV43.VERSION.equals( version ) )
```
gjmwoods, line 172: is there more cleanup we can do for removing 4.0? possibly not since its the first in the series

gjmwoods (approved):
another masterpiece 🎨

driver/src/test/java/org/neo4j/driver/integration/RoutingDriverBoltKitIT.java
bigmontz: It's already in testkit, so i will remove. 😢

driver/src/main/java/org/neo4j/driver/internal/messaging/BoltProtocol.java
bigmontz: I kept the 4.0 there since we don't have a final decision about the versions we will support in 4.3

PR #774, opened 2020-10-26
Refactor Testkit-backend to use jackson for databinding

bigmontz:
The objective of this PR is to support the current enabled tests on the Java Testkit Suite in more extensible and maintainable structure.

The commands were separated in different objects which are mapped by Jackson and ran by the CommandProcessor. Jackson is also responsible for serializing the response using custom serializers to map values and records to CypherValues, structure used by teskit to return the values with the type information.


testkit-backend/pom.xml
```diff
@@ -32,6 +32,24 @@
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
         </dependency>
+        <dependency>
```
michael-simons, line 35: We try to avoid Lombok…

testkit-backend/src/main/java/neo4j/org/testkit/backend/Runner.java
```diff
@@ -16,41 +16,59 @@
...
 import java.net.Socket;
+import java.util.concurrent.CompletableFuture;
 
 public class Runner
 {
     public static void main( String[] args ) throws IOException
     {
-        ServerSocket serverSocket = new ServerSocket(9876);
-        System.out.println("Starting Java Testkit Backend Started");
+        ServerSocket serverSocket = new ServerSocket( 9876 );
```
michael-simons, line 35: I would strongly recommend using the local loop back address instead of the wildcard (this is what the standard ctor does, using 0.0.0.0). I have had issues with several projects on a couple of machines.

testkit-backend/src/main/java/neo4j/org/testkit/backend/messages/requests/SessionReadTransaction.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Override
+    public TestkitResponse process( TestkitState testkitState )
+    {
+        return Optional.ofNullable( testkitState.getSessionStates().getOrDefault( data.sessionId, null ) )
+                       .map( sessionState ->
+                             {
+                                 Session session = sessionState.getSession();
+                                 session.readTransaction( tx ->
+                                                          {
```
michael-simons: Such handlers should be refactored out into a method that return the fitting SAM type

i.e

```
public TransactionWork<TestkitResponse> handler() {
        return tx -> new TestkitResponse(); // Whatever
   }
// and

session.readTransaction(handler());
```

I call those lambdas the lambda pyramid of death ;)

testkit-backend/src/main/java/neo4j/org/testkit/backend/messages/requests/TestkitRequest.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package neo4j.org.testkit.backend.messages.requests;
+
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import neo4j.org.testkit.backend.TestkitState;
+import neo4j.org.testkit.backend.messages.responses.TestkitResponse;
+
+@JsonTypeInfo( use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "name" )
+@JsonSubTypes( {
```
michael-simons, line 27: Cool! Learned something new!

testkit-backend/src/main/java/neo4j/org/testkit/backend/messages/requests/TransactionCommit.java
```diff
@@ -0,0 +1,61 @@
...
+@NoArgsConstructor
+@Setter
+public class TransactionCommit implements TestkitRequest
+{
+    private TransactionCommitBody data;
+
+    @Override
+    public TestkitResponse process( TestkitState testkitState )
+    {
+        return Optional.ofNullable( testkitState.getTransactions().getOrDefault( data.txId, null ) )
```
michael-simons: Unnecessary.

If the key is not in the map, the map will return `null` anyway.

testkit-backend/src/main/java/neo4j/org/testkit/backend/messages/requests/TransactionRun.java
```diff
@@ -0,0 +1,64 @@
...
+@Getter
+@NoArgsConstructor
+public class TransactionRun implements TestkitRequest
+{
+    private TransactionRunBody data;
+
+    @Override
+    public TestkitResponse process( TestkitState testkitState )
+    {
+        return Optional.ofNullable( testkitState.getTransactions().getOrDefault( data.txId, null ) )
```
michael-simons: Same as above.

testkit-backend/src/main/java/neo4j/org/testkit/backend/messages/responses/RetryableDone.java
```diff
@@ -0,0 +1,37 @@
...
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Setter;
+
+@Setter
+@Getter
+@Builder
+public class RetryableDone implements TestkitResponse
+{
+    private final Object data = null;
```
michael-simons: Is this a left over or needed? 

That's the thing with Lombok and the magic getters… One never knows.

michael-simons (changes requested):
I get the basic idea and I like it.
I have a couple of remarks added and some things I would change.

I definitiv recommend removing Lombok from any library. 

Lombok is ok to do in application code, but not in something that is shared as a library. It's a pita to debug, as source and classes won't much if you don't delombok during build.

testkit-backend/src/main/java/neo4j/org/testkit/backend/Runner.java
```diff
@@ -1,4 +1,4 @@
-/*
+package neo4j.org.testkit.backend;/*
```
meistermeier: Formatting :)

testkit-backend/pom.xml
bigmontz: I could delombok the code, it's not an issue. But the `testkit-backend` is not a library, it's an application.

testkit-backend/src/main/java/neo4j/org/testkit/backend/Runner.java
2hdddg: Local loop back will not work when the backend is running in a container (which it normally is)

testkit-backend/src/main/java/neo4j/org/testkit/backend/messages/responses/RetryableDone.java
bigmontz: It's something with the testkit, it needs to have always the `data` entry on all the responses, also in the ones which doesn't make any sense to have it. I could fix this thing on the testkit, i will check with Peter if the change will not have undesirable effects.

2hdddg: There is no check in python handler for if data exists or not, we just splat it into the constructor of the response with the same name. I think it is better to have it there... Shouldn't it be part of base class?

bigmontz: Done.

testkit-backend/src/main/java/neo4j/org/testkit/backend/Runner.java
bigmontz: Just to be sure, I tested and It didn't work.

michael-simons: Ok, makes sense @2hdddg 

The clients connecting against the backend: Are they on the same network docker network? Or coming from the external? If so,  shouldn't it than use the containers actual network address? 

But alas, if it works, it's all good. I just had this issue with the maven daemon like 3 days ago and therefore I noticed it.

2hdddg: It will work with 0.0.0.0, can't see that it should be a problem in docker, could be when running on host. If you want you can add a commandline param to control the interface to bind to and bind to localhost when running outside of docker.
The clients are on the same network (dynamically created), we connect through host names so we would need to figure out (or configure) the ip address of the driver container if we for some reason think we must bind to that exact interface.
It would work to bind to localhost if we use host network instead of a bridge but that is like asking for problems.

michael-simons (approved):
Thanks for looking closely into my remarks. 👍

PR #793, opened 2020-11-16
Additional Testkit support

gjmwoods:


testkit-backend/src/main/java/neo4j/org/testkit/backend/messages/requests/SessionBeginTransaction.java
```diff
@@ -47,7 +48,12 @@ public TestkitResponse process( TestkitState testkitState )
                              {
                                  TransactionConfig.Builder builder = TransactionConfig.builder();
                                  Optional.ofNullable( data.txMeta ).ifPresent( builder::withMetadata );
-                                 // Optional.ofNullable( data.timeout ).ifPresent( builder::withTimeout );
```
bigmontz, line 50: Please, remove the commented line.

bigmontz (approved):
🔧

.gitignore
```diff
@@ -26,3 +26,4 @@ community/server-plugin-test/neo4j-home
 enterprise/server-enterprise/neo4j-home
 integrationtests/data
 dependency-reduced-pom.xml
+venv
```
fbiville, line 29: out of curiosity, can `venv` work with JDK installations?
I'm currently using https://www.jenv.be/, but if there is a better alternative, I'm eager to try :)

gjmwoods: venv I think is generated by IntelliJ for python. Haven't seen jenv before, I'm just swithcing between java versions using aliases "alias j8='export JAVA_HOME=`/usr/libexec/java_home -v 1.8`'"

PR #798, opened 2020-12-08
Add support to 4.2 by implementing the extended handshake

bigmontz:
The Bolt handshake is a short exchange that occurs after the opening of every Bolt connection. It is used to establish a protocol version for subsequent messaging activity. The sequence is currently limited to four fixed protocol versions; these changes implement a mechanism to extend that to protocol version ranges.

In Bolt v1, the four bytes were treated as a single big-endian unsigned 32-bit integer, with the protocol versioned only via a single number. In Bolt 4.0, the scheme evolved such that bytes D and C represented the major and minor version numbers respectively; bytes A and B were ignored, reserved for future use.

The scheme is to be extended such that byte B can be used to allow the sequence to encode an inclusive range of versions in the client-to-server request. The server-to-client response would remain the same, specifying only a single version.

Specifically, the range minimum and maximum can be computed as follows:

Range maximum = D.C
Range minimum = D.(C - B)

This denotes byte B as a minor version difference, backwards from the minor version in byte C. In other words, zero denotes "zero versions back", one denotes "one version back", and so on. Note that this specifically requires the range to exist within the same major version, but does provide backwards compatibility with existing implementations, which fill bytes A and B with zeroes.

Note also that a range with an identical min and max should be considered semantically equivalent to a single version.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/async/connection/BoltProtocolUtil.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/async/connection/BoltProtocolUtil.java
@@ -42,9 +42,9 @@ public final class BoltProtocolUtil
 
     private static final ByteBuf HANDSHAKE_BUF = unreleasableBuffer( copyInt(
             BOLT_MAGIC_PREAMBLE,
-            BoltProtocolV43.VERSION.toInt(),
-            BoltProtocolV42.VERSION.toInt(),
+            BoltProtocolV43.VERSION.toIntRange( BoltProtocolV42.VERSION ),
             BoltProtocolV41.VERSION.toInt(),
+            BoltProtocolV4.VERSION.toInt(),
             BoltProtocolV3.VERSION.toInt() ) ).asReadOnly();
 
     private static final String HANDSHAKE_STRING = createHandshakeString();
--- a/driver/src/main/java/org/neo4j/driver/internal/messaging/BoltProtocolVersion.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/messaging/BoltProtocolVersion.java
@@ -34,7 +34,7 @@ public class BoltProtocolVersion implements Comparable<BoltProtocolVersion>
     public static BoltProtocolVersion fromRawBytes( int rawVersion )
     {
         int major = rawVersion & 0x000000FF;
-        int minor = ( rawVersion >> 8 ) & 0x000000FF;
+        int minor = (rawVersion >> 8) & 0x000000FF;
 
         return new BoltProtocolVersion( major, minor );
     }
@@ -55,15 +55,15 @@ public class BoltProtocolVersion implements Comparable<BoltProtocolVersion>
         return shiftedMinor | majorVersion;
     }
 
-    public int toIntRange(BoltProtocolVersion minVersion)
+    public int toIntRange( BoltProtocolVersion minVersion )
     {
-        if(majorVersion != minVersion.majorVersion)
+        if ( majorVersion != minVersion.majorVersion )
         {
-            throw new IllegalArgumentException( "Versions should be from the same version" );
+            throw new IllegalArgumentException( "Versions should be from the same major version" );
         }
-        else if (minorVersion < minVersion.minorVersion)
+        else if ( minorVersion < minVersion.minorVersion )
         {
-            throw new IllegalArgumentException("Max version should be newest than min version");
+            throw new IllegalArgumentException( "Max version should be newer than min version" );
         }
         int range = minorVersion - minVersion.minorVersion;
         int shiftedRange = range << 16;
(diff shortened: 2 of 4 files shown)


driver/src/main/java/org/neo4j/driver/internal/messaging/BoltProtocolVersion.java
```diff
@@ -55,6 +55,21 @@ public int toInt()
         return shiftedMinor | majorVersion;
     }
 
+    public int toIntRange(BoltProtocolVersion minVersion)
+    {
+        if(majorVersion != minVersion.majorVersion)
+        {
+            throw new IllegalArgumentException( "Versions should be from the same version" );
```
fbiville:
```suggestion
            throw new IllegalArgumentException( "Versions should be from the same major version" );
```

driver/src/main/java/org/neo4j/driver/internal/messaging/BoltProtocolVersion.java
```diff
@@ -55,6 +55,21 @@ public int toInt()
...
+        }
+        else if (minorVersion < minVersion.minorVersion)
+        {
+            throw new IllegalArgumentException("Max version should be newest than min version");
```
fbiville:
```suggestion
            throw new IllegalArgumentException("Max version should be newer than min version");
```

driver/src/main/java/org/neo4j/driver/internal/async/connection/BoltProtocolUtil.java
```diff
@@ -42,9 +42,9 @@
 
     private static final ByteBuf HANDSHAKE_BUF = unreleasableBuffer( copyInt(
             BOLT_MAGIC_PREAMBLE,
-            BoltProtocolV43.VERSION.toInt(),
-            BoltProtocolV42.VERSION.toInt(),
+            BoltProtocolV43.VERSION.toIntRange( BoltProtocolV42.VERSION ),
```
gjmwoods, line 45: just wondering if we should push the range option to the bottom for 4.3 -> 4.2. So the handshake is 4.3, 4.2, 4.1-> 4.0, 3. When 4.3 driver talks to 4.2 server it would skip over 4.2 and negotiate 4.1.

driver/src/main/java/org/neo4j/driver/internal/messaging/BoltProtocolVersion.java
gjmwoods: nit: I know its less consistently applied in the driver but can you format the changes using the server checkstyles

driver/src/main/java/org/neo4j/driver/internal/async/connection/BoltProtocolUtil.java
bigmontz: Then we will lose the multi database support on the version 4.0, since i will negotiate as 3.5. 
Uses 4.1 to talk with a 4.2.x server will not represent any feature drop.

gjmwoods: Ah good point. I guess we just have to be a little careful whilst we switch to the new handshake

gjmwoods (approved):
🐒

driver/src/main/java/org/neo4j/driver/internal/async/connection/BoltProtocolUtil.java
bigmontz: I've made a comment on the protocol definition document, to fix the example to consider the version range in decreasing order.

driver/src/main/java/org/neo4j/driver/internal/messaging/BoltProtocolVersion.java
bigmontz: I've did a format on the file, what's off in the current format?

gjmwoods: the space before and after the bracket : `throw new IllegalArgumentException( "Max version should be newest than min version" );`

driver/src/main/java/org/neo4j/driver/internal/async/connection/BoltProtocolUtil.java
AndyHeap-NeoTech: Is the idea that we should go to BoltProtocolV43.VERSION.toIntRange(BoltProtocolVersionV40.VERSION)? That way we are saying all version 4 protocols are accepted by this driver and save 2 slots in the handshake.

bigmontz: it could be done in the protocol, but I think we should keep using the slots for the 4.0 and 4.1, just in case we connect to a non-update server which do not know how to interpret the extend handshake.

PR #799, opened 2020-12-09
Improve examples and use RxJava '.using' for managing the session.

michael-simons:
This aligns the RxJava examples with the Project Reactor ones and makes sure that the session is closed.

Much more important however is the fact that the session is not opened in the first place prior to subscription to the flows.

Also take note that I have rewritten the unmanaged variant for Project Reactor. While the previous setup does work in our test rig, I had issues by using a subscriber and not the `await` testing tool.


gjmwoods (approved):
🐒

PR #803, opened 2021-01-08
Initial stress test in testkit setup

2hdddg:
Support for starting existing Java stress test suite in testkit.
Moves existing testkit glue code from testkit repo to this repository.
Note that the stress test suite doesn't run properly in testkit at this point but shouldn't break existing usage of the suite.


gjmwoods (approved):
🚀

PR #813, opened 2021-01-22
Updating stress tests exception and session handling

injectives:
Exception handling is updated to expose the leader switching exceptions for managed transactions.
Async session handling is updated to close session and report problems if they arise.


gjmwoods (approved):
👍

PR #819, opened 2021-02-08
DNS change fix

injectives:
This update fixes an issue when the driver fails to update the routing table if the router DNS entry returns a different IP address.

This PR cherry-picks the changes from: https://github.com/neo4j/neo4j-java-driver/pull/818


gjmwoods (approved):
🚀

PR #829, opened 2021-02-19
Removed the year component from all copyright notices

dotStart:


dotStart:
2.0 branch has been deprecated - No longer needed

Issue #833, opened 2021-02-25
DNS resolution fails when cluster members are offline


ekampp:
Hi team. 

First off, thanks for a super nice tool, and sorry if we're just not understanding it correctly. We have recently migrated to JRuby to leverage the java driver for a variety of reasons. 

We realized that neo4j-java-driver is incorrectly operating with DNS resolution mode. 

- Neo4j version: Enterprise 4.2.2
- Neo4j Mode: A cluster with 5 members/Casual cluster with 5 core 0 read-replica
- Driver version: neo4j-java-driver-4.2.0.beta.0-java 
- Operating system: Ubuntu 20.04.2 (official neo4j AMI)

### Steps to reproduce

The following connection URI was used:
```
NEO4J_URL=neo4j://<user>:<password>@neo4j.infra.prod.internal:7687
```

The DNS query for neo4j.infra.prod.internal returns 3 addresses:
```
neo4j.infra.prod.internal has address 10.26.219.30
neo4j.infra.prod.internal has address 10.26.221.188
neo4j.infra.prod.internal has address 10.26.204.91
```

When all aforementioned members are available, everything is fine. The driver connects to one of them successfully and gets the neo4j routing table:
```
INFO: Closing connection pool towards neo4j.infra.prod.internal(10.26.204.91):7687, it has no active connections and is not in the routing table registry.
```

However, if one of those members is down and the `gethostbyname()` library call returns its address at the top of the resulting list, then the neo4j-java-driver will fail to bootstrap:
```
Caused by: org.neo4j.driver.exceptions.ServiceUnavailableException: Unable to connect to neo4j.infra.prod.internal(10.26.221.188):7687, ensure the database is running and that there is a working network connection
to it. 
```

### Expected behavior

When a node member goes away, the connection should roll over to the next online member.

### Actual behavior

The driver is not trying to connect to other hosts returned in the list.

### Logs

[logs.txt](https://github.com/neo4j/neo4j-java-driver/files/6041359/logs.txt)

injectives:
Hi @ekampp,

Thank you for reporting this issue. It has actually been discovered internally too and we are going to fix it. An update will be posted here later on.

For the time being you might want to try the latest 4.2.1 version, which fixed another issue, but may result in a more reliable behaviour (especially in combination with the retry mechanism in the tx functions).

ekampp:
@injectives, thank you for the quick and positive feedback. 

[...]

PR #849, opened 2021-03-10
Fix for initial routers DNS resolution

injectives:
Fix for initial routers DNS resolution

This update fixes the following issue: https://github.com/neo4j/neo4j-java-driver/issues/833

The desired behaviour for getting a routing table from the initial router (either on bootstrap or when all known routers have failed) is:
- resolve the domain name to all IPs
- attempt getting a routing table from all of them until first one succeeds by:
  - getting a connection
  - trying to get a successful routing table response

Prior to this change, the connection pools were created for host and port pairs. When domain name of the host resolves to multiple IP addresses, such pools provide connections to those IPs as a group. While this works for readers and writers, it negatively impacts the routing table fetching process as there is no guarantee which IP address the provided connection is setup for.

This update delivers the following changes:
- connection pools for routers are IP address based, which allows for deterministic connection retrieval
- the resolved IP address set is kept up-to-date (in case known router IPs change) to make sure that the unused connection pools are flushed
- the domain name resolution logic has been made configurable (it is private at the moment and is used to facilitate testing)
- the testkit backend has been updated to support the domain name resolution configuration (a new test has been added to testkit to cover the issue described above)
- the testkit backend has been updated to support connection timeout driver configuration
- several tests have been updated to adopt the new changes

### DIFF
--- a/driver/src/test/java/org/neo4j/driver/internal/cluster/RediscoveryTest.java
+++ b/driver/src/test/java/org/neo4j/driver/internal/cluster/RediscoveryTest.java
@@ -23,6 +23,8 @@ import org.junit.jupiter.api.Test;
 import org.mockito.ArgumentCaptor;
 
 import java.io.IOException;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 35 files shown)


gjmwoods (approved):
💪

driver/src/test/java/org/neo4j/driver/internal/cluster/RediscoveryTest.java
```diff
@@ -412,16 +417,20 @@ void shouldNotLogWhenSingleRetryAttemptFails()
     }
 
     @Test
-    void shouldNotResolveToIPs()
+    void shouldNotResolveToIPs() throws UnknownHostException
```
gjmwoods: maybe we should rename this test :)

gjmwoods (approved):

driver/src/test/java/org/neo4j/driver/internal/cluster/RediscoveryTest.java
injectives: Good catch, thank you! I've updated the test name.

PR #851, opened 2021-03-17
Fixed SSL handling

injectives:
This update fixes a number of SSL-related tests in testkit and CausalClusteringIT.shouldDropBrokenOldConnections test.
The connection pooling strategy has been updated to use the same connection pool when the connection host is unambiguous.
Removed hardcoded domain name resolution from the BoltServerAddress and moved the logic to ChannelConnectorImpl that uses the DomainNameResolver.


gjmwoods (approved):
💯

gjmwoods (approved):
☎️

PR #861, opened 2021-03-30
Include bookmarks in the ROUTE message.

gjmwoods:
Adding bookmarks to ROUTE message

The new format is: `ROUTE {routingContext} [listOfBookmarks] dbname`
 where (in order of occurrence):
- `routingContext` provided by the driver.
- A list of bookmarks. If none are provided then an empty list should be supplied.
- the `dbname` of the database for which a new routing table is being requested. This is a string value and can be replaced with `null` to return the default database's routing table.


injectives (approved):
LGTM

PR #886, opened 2021-04-27
Added authorization expired response handling

injectives:
This update adds support for authorization expired response handling. When such error arrives the driver will mark all connections that were created on or before the creation timestamp of the receiving connection for disposal. The connections will be allowed to finish what they are doing and released back to the pool and will be disposed on the next acquire call.

This update also makes sure that we always register the HELLO response handler so that we can handle the response appropriately.

In addition, it adds the ROLLBACK message support to the driver backend.

### DIFF
--- a/driver/src/test/java/org/neo4j/driver/integration/ConnectionHandlingIT.java
+++ b/driver/src/test/java/org/neo4j/driver/integration/ConnectionHandlingIT.java
@@ -32,15 +32,18 @@ import reactor.test.StepVerifier;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
+import java.util.logging.Level;
 
 import org.neo4j.driver.AuthToken;
 import org.neo4j.driver.Config;
 import org.neo4j.driver.Driver;
 import org.neo4j.driver.Logging;
-import org.neo4j.driver.Record;
-import org.neo4j.driver.Session;
-import org.neo4j.driver.Result;
 import org.neo4j.driver.QueryRunner;
+import org.neo4j.driver.Record;
+import org.neo4j.driver.Result;
+import org.neo4j.driver.Session;
 import org.neo4j.driver.Transaction;
 import org.neo4j.driver.exceptions.ClientException;
 import org.neo4j.driver.internal.BoltServerAddress;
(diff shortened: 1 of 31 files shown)


driver/src/test/java/org/neo4j/driver/integration/ConnectionHandlingIT.java
```diff
@@ -100,7 +102,7 @@ void createDriver()
         AuthToken auth = neo4j.authToken();
         RoutingSettings routingSettings = RoutingSettings.DEFAULT;
         RetrySettings retrySettings = RetrySettings.DEFAULT;
-        driver = driverFactory.newInstance( neo4j.uri(), auth, routingSettings, retrySettings, defaultConfig(), SecurityPlanImpl.insecure() );
+        driver = driverFactory.newInstance( neo4j.uri(), auth, routingSettings, retrySettings, Config.builder().withLogging( Logging.console( Level.FINE ) ).build(), SecurityPlanImpl.insecure() );
```
gjmwoods: just checking the logging is Intentional

driver/src/test/java/org/neo4j/driver/internal/async/pool/NettyChannelHealthCheckerTest.java
```diff
@@ -91,6 +97,55 @@ void shouldAllowVeryOldChannelsWhenMaxLifetimeDisabled()
...
+
+        for ( int i = 0; i < channels.size(); i++ )
+        {
+            Channel channel = channels.get( i );
+            boolean health = Objects.requireNonNull( await( healthChecker.isHealthy( channel ) ) );
+            boolean expectedHealth = i > authorizationExpiredChannelIndex;
+            assertEquals( expectedHealth, health, String.format( "Channel %d has failed the check", i ) );
+        }
+    }
+
```
gjmwoods, line 127: Maybe a test that checks newer connections are still healthy?

gjmwoods (approved):

driver/src/test/java/org/neo4j/driver/integration/ConnectionHandlingIT.java
injectives: Sorry, this is a trace from debugging. I have updated it now.

driver/src/test/java/org/neo4j/driver/internal/async/pool/NettyChannelHealthCheckerTest.java
injectives: It actually checks both, see the `expectedHealth` variable definition - it depends on the position in the list, which is in chronological order.

gjmwoods: ah of course, I missread it

PR #895, opened 2021-05-21
Remove circular references in exceptions during the commit

### LINKED ISSUE #773: [Bug] Error construction leads to StackOverflowException when logging it
**Neo4j Version:** 4.1.3 Community  
**Neo4j Mode**: Single instance  
**Driver version**: Java driver 4.1.1  
**Operating System:** Ubuntu 18   

Description of the bug: 
When making a request with old syntax like "MATCH(n:Element) WHERE n.name = {param} RETURN n" through the java driver
The driver throws a ClientException 
```
org.neo4j.driver.exceptions.ClientException: The old parameter syntax `{param}` is no longer supported. Please use `$param` instead (line 1, column 34 (offset: 33))
"MATCH (n:Element) WHERE n.name = {param} RETURN n"
```

This exception has a suppressed exception (ClientException: Transaction can't be committed)
This exception has cause org.neo4j.driver.exceptions.ClientException: The old parameter syntax `{param}` is no longer supported 
.....

This is an infinite reccursion, which leads to Stackoverflow when logging the Exception

Expected behaviour: 
There should not be an infinite recursion in the Exception construction

Code sample to reproduce
`private static final Logger LOGGER = LoggerFactory.getLogger(Neo4jBoltClientTest.class);

    @Test
    public void shouldTerminate() {
        CompletionStage<List<ResultCursor>> listCompletionStage = new TmpClient().executionInTransaction(tx -> Flux.from(
                Mono.fromCompletionStage(
                        tx.runAsync("MATCH (n:Element) WHERE n.name = {param} RETURN n", Map.of("param", "Luke"))
                )
        ));

        List<ResultCursor> block = Mono.fromCompletionStage(listCompletionStage).block();

    }

    private class TmpClient {

        // here wNeo is a test rule, you can put any database URL. The database doesn't matter much, and the data in it either, since the request can't be executed anyways
        private String url = wNeo.boltURI().toString(); 
        private Driver driver = GraphDatabase.driver(url,
                AuthTokens.basic("neo4j", "neo4j"));

        public <T> CompletionStage<List<T>> executionInTransaction(Function<AsyncTransaction, Flux<T>> txFunction) {
            return driver.asyncSession()
                    .writeTransactionAsync(tx -> txFunction.apply(tx).collectList().toFuture())
                    .whenComplete((val, thr) -> {
                        try {
                            LOGGER.error("error", thr);
                        } catch (Throwable t) {
                            LOGGER.error("error in log", t); // this will catch a Stackoverflow Exception
                        }
                    });
        }
    }`

Log output: 
`18:18:03.998 [main ] INFO  Driver - Direct driver instance 814684352 created for server address 127.0.0.1:46295
18:18:14.544 [Neo4jDriverIO-3-3] ERROR c.l.l.n.i.c.Neo4jBoltClientTest - error in log
java.lang.StackOverflowError: null
	at java.base/java.lang.reflect.InvocationTargetException.<init>(InvocationTargetException.java:73)
	at java.base/jdk.internal.reflect.GeneratedMethodAccessor6.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at ch.qos.logback.classic.spi.ThrowableProxy.<init>(ThrowableProxy.java:66)
	at ch.qos.logback.classic.spi.ThrowableProxy.<init>(ThrowableProxy.java:60)
	at ch.qos.logback.classic.spi.ThrowableProxy.<init>(ThrowableProxy.java:72)
	at ch.qos.logback.classic.spi.ThrowableProxy.<init>(ThrowableProxy.java:60)
	at ch.qos.logback.classic.spi.ThrowableProxy.<init>(ThrowableProxy.java:72)
        .... (lines 60/72 alternate until Stackoverflow)
`

NOTE: this behaviour doesn't happen with some other syntax errors.
For example, "MATCH(n:ELEMENT)" - which will fail since it has no return or update clause) will return a ClientException without any suppressed exception, thus it can be logged without problems

bigmontz:
The error issue happened because the cursor failure is already the cause of the commit or rollback failure. This generates a circular reference when both of the exceptions are combined.

This situation can create stack overflow when the exception is logged in some logging libraries.

The solution don't add the cursor exception as the cause of the exception, this way it will not be wrongly combined creating a circular reference.

### DIFF
--- a/driver/src/test/java/org/neo4j/driver/integration/async/AsyncQueryIT.java
+++ b/driver/src/test/java/org/neo4j/driver/integration/async/AsyncQueryIT.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.neo4j.driver.integration.async;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.ArrayList;
+import java.util.concurrent.ExecutionException;
+
+import org.neo4j.driver.async.AsyncSession;
+import org.neo4j.driver.util.DatabaseExtension;
+import org.neo4j.driver.util.ParallelizableIT;
+
+import static org.neo4j.driver.Values.parameters;
+import static org.neo4j.driver.util.TestUtil.assertNoCircularReferences;
+
+@ParallelizableIT
+public class AsyncQueryIT
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger( AsyncQueryIT.class );
+
+    @RegisterExtension
+    static final DatabaseExtension neo4j = new DatabaseExtension();
+
+    private AsyncSession session;
+
+    @BeforeEach
+    void setUp()
+    {
+        session = neo4j.driver().asyncSession();
+    }
+
+    @AfterEach
+    void tearDown()
+    {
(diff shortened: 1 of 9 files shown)


driver/src/test/java/org/neo4j/driver/integration/async/AsyncQueryIT.java
```diff
@@ -0,0 +1,81 @@
...
+    void tearDown()
+    {
+        session.closeAsync();
+    }
+
+    @Test
+    void shouldBeAbleToLogSemanticWrongExceptions() throws ExecutionException, InterruptedException
+    {
+        session.writeTransactionAsync( tx -> Flux.from(
+                Mono.fromCompletionStage(
```
injectives, line 69: How about using plain async style here? Without RX.

injectives: Or was it needed for something?

bigmontz: It comes from the bug description.

https://github.com/neo4j/neo4j-java-driver/issues/773

I like to keep the original code or scenario which revealed the bug in the test case.

injectives: I see, cool 👍 
At least, we have documented this here.

injectives (approved):
LGTM

PR #897, opened 2021-05-24
Run response immediate processing update

injectives:
This update makes `run` methods (session, tx) wait for successful request acknowledgement by the server. Specifically, the `Result` object will only be returned when there is a successful response from the server to the `RUN` message, meaning that the request was at least accepted by the server for processing. Otherwise, an appropriate error will be thrown.

Additionally, this update also makes the `Result.keys` method non-blocking and free from potential communication errors.


driver/src/main/java/org/neo4j/driver/internal/handlers/RunResponseHandler.java
```diff
@@ -22,28 +22,31 @@
...
+import org.neo4j.driver.internal.async.UnmanagedTransaction;
+import org.neo4j.driver.internal.spi.Connection;
 import org.neo4j.driver.internal.spi.ResponseHandler;
 import org.neo4j.driver.internal.util.MetadataExtractor;
 import org.neo4j.driver.internal.util.QueryKeys;
 
 public class RunResponseHandler implements ResponseHandler
 {
-    private final CompletableFuture<Throwable> runCompletedFuture;
+    private final CompletableFuture<Void> runFuture;
```
michael-simons, line 34: Very nice change… Completable future of throwable was weird.

driver/src/main/java/org/neo4j/driver/internal/handlers/TransactionPullResponseCompletionListener.java
```diff
@@ -45,6 +45,9 @@ public void afterFailure( Throwable error )
         // always mark transaction as terminated because every error is "acknowledged" with a RESET message
         // so database forgets about the transaction after the first error
         // such transaction should not attempt to commit and can be considered as rolled back
-        tx.markTerminated( error );
+        if ( tx.isOpen() )
```
michael-simons: Is there a danger that we lose the error if the tx isn't open?

driver/src/test/java/org/neo4j/driver/integration/TransactionIT.java
```diff
@@ -324,15 +320,13 @@ void shouldRollBackTxIfErrorWithConsume()
     }
 
     @Test
-    void shouldPropagateFailureFromSummary()
+    void shouldFailRun()
     {
         try ( Transaction tx = session.beginTransaction() )
         {
-            Result result = tx.run( "RETURN Wrong" );
+            ClientException e = assertThrows( ClientException.class, () -> tx.run( "RETURN Wrong" ) );
```
michael-simons, line 327: that test would be a nice example to document the changed behavior for example.

driver/src/test/java/org/neo4j/driver/integration/async/AsyncTransactionIT.java
```diff
@@ -679,53 +669,53 @@ void shouldFailToCommitWhenQueriesFailAndErrorNotConsumed() throws InterruptedEx
...
+    void shouldFailToCommitWhenRunFailed()
     {
         AsyncTransaction tx = await( session.beginTransactionAsync() );
 
         tx.runAsync( "RETURN ILLEGAL" );
 
         ClientException e = assertThrows( ClientException.class, () -> await( tx.commitAsync() ) );
         assertNoCircularReferences( e );
-        assertThat( e.getMessage(), containsString( "ILLEGAL" ) );
+        assertThat( e.getMessage(), containsString( "Transaction can't be committed" ) );
```
michael-simons, line 685: I wonder how that even works in a non flaky way tbh.

driver/src/main/java/org/neo4j/driver/internal/cursor/ResultCursorFactoryImpl.java
```diff
@@ -39,24 +39,25 @@
 
     private final PullResponseHandler pullHandler;
     private final PullAllResponseHandler pullAllHandler;
-    private final boolean waitForRunResponse;
     private final Message runMessage;
+    private final CompletableFuture<Void> runFuture;
```
michael-simons, line 43: The purpose of that `runFuture` isn't clear to me at all just from looking at that bit.

michael-simons (changes requested):
From what I understand it seems to make sense. Looks at least simpler in a positive way. 
The core idea is not to pass that `boolean` to `ResultCursorFactoryImpl` and a new `runFuture`, right? 
Could you elaborate a bit more on the purpose of that `runFucture`? By maybe choosing a different name?

injectives:
> The core idea is not to pass that `boolean` to `ResultCursorFactoryImpl` and a new `runFuture`, right?

Basically, yes. The `waitForRunResponse` did not make sense anymore given the goal of these changes. From API perspective, the main objective is to improve the `run` methods handling by making sure we only return `Result` object when server successfully accepts the `RUN` Bolt message ([details](https://7687.org/bolt/bolt-protocol-message-specification-4.html#request-message---run)). In such way the `Result` object represents a successful acceptance of the request by the server. Otherwise, an error must be thrown.
Previously, it would return `Result` object in most cases, even if `waitForRunResponse` was set to `true`. For instance, see [the previous implementation](https://github.com/neo4j/neo4j-java-driver/pull/897/files#diff-b88986c83241d14cf14fcaf5d3a7221848469d0954b1597fa9e1f99262ede3f3L72). Keep in mind that `runFuture()` used to return `CompletableFuture<Throwable>`.
The new `runFuture` if part of achieving the aforementioned goal.

injectives:
> Could you elaborate a bit more on the purpose of that `runFucture`? By maybe choosing a different name?

Basic `ResponseHandler` implementations take `CompletableFuture` (or something similar) that is meant to contain the processed response. They are responsible for completing it when appropriate. For instance:
- `beginTxFuture` in (BeginTxResponseHandler)[https://github.com/neo4j/neo4j-java-driver/blob/4.3/driver/src/main/java/org/neo4j/driver/internal/handlers/BeginTxResponseHandler.java#L32]
- `commitFuture` in (CommitTxResponseHandler)[https://github.com/neo4j/neo4j-java-driver/blob/4.3/driver/src/main/java/org/neo4j/driver/internal/handlers/CommitTxResponseHandler.java#L34]
- `rollbackFuture` in (RollbackTxResponseHandler)[https://github.com/neo4j/neo4j-java-driver/blob/4.3/driver/src/main/java/org/neo4j/driver/internal/handlers/RollbackTxResponseHandler.java#L32]
- `completionFuture` in (ResetResponseHandler)[https://github.com/neo4j/neo4j-java-driver/blob/4.3/driver/src/main/java/org/neo4j/driver/internal/handlers/ResetResponseHandler.java#L31]
- `completableFuture` in (RouteMessageResponseHandler)[https://github.com/neo4j/neo4j-java-driver/blob/4.3/driver/src/main/java/org/neo4j/driver/internal/handlers/RouteMessageResponseHandler.java#L38]
- etc.

Following this approach, the `runFuture` is meant to contain the processed response. I am happy to rename it to something else, I just though it could be better to keep the naming similar to other handlers in this change. While there is actually some discrepancy in names, this 'action'Future pattern seems to be a fairly common in the current implementation.

How about introducing doing a separate pull request and introducing a new common naming, like 'action`CompletionFuture (subject to discussion in that PR)? It might be easier from review standpoint.

driver/src/main/java/org/neo4j/driver/internal/handlers/RunResponseHandler.java
injectives: Yeah 👍

driver/src/main/java/org/neo4j/driver/internal/cursor/ResultCursorFactoryImpl.java
injectives: To avoid duplication, could you take a look at the following [message](https://github.com/neo4j/neo4j-java-driver/pull/897#issuecomment-852917461) please?

michael-simons:
Following this approach, the `runFuture` is meant to contain the processed response. I am happy to rename it to something else, I just though it could be better to keep the naming similar to other handlers in this change. While there is actually some discrepancy in names, this 'action'Future pattern seems to be a fairly common in the current implementation.

I think `responseFuture` would fit than.

driver/src/test/java/org/neo4j/driver/integration/async/AsyncTransactionIT.java
injectives: As far as I am aware, this test in not known as flaky at the moment.........

The internal TX keeps track of all result cursors. For instance, `runAsync` registers them [here](https://github.com/neo4j/neo4j-java-driver/blob/4.3/driver/src/main/java/org/neo4j/driver/internal/async/UnmanagedTransaction.java#L212).
The `commitAsync` executes [after](https://github.com/neo4j/neo4j-java-driver/blob/4.3/driver/src/main/java/org/neo4j/driver/internal/async/UnmanagedTransaction.java#L183) they complete.

The `RunResponseHandler` will mark the transaction as [failed](https://github.com/neo4j/neo4j-java-driver/pull/897/files#diff-2b8f9fbf88ad5c053b5900b59440e00c7a6e318908f9e3c247a95407facc9048R67) and the `doCommitAsync` will [fail](https://github.com/neo4j/neo4j-java-driver/blob/4.3/driver/src/main/java/org/neo4j/driver/internal/async/UnmanagedTransaction.java#L261).

injectives:
> Following this approach, the `runFuture` is meant to contain the processed response. I am happy to rename it to something else, I just though it could be better to keep the naming similar to other handlers in this change. While there is actually some discrepancy in names, this 'action'Future pattern seems to be a fairly common in the current implementation.
> 
> I think `responseFuture` would fit than.

As discussed, a separate PR will be created with renames to make them easier to review.

driver/src/main/java/org/neo4j/driver/internal/cursor/ResultCursorFactoryImpl.java
injectives: Agreed the [approach](https://github.com/neo4j/neo4j-java-driver/pull/897#issuecomment-855839119).

driver/src/main/java/org/neo4j/driver/internal/handlers/TransactionPullResponseCompletionListener.java
injectives: This was mainly to account for the fact that `RunResponseHandler` may mark transaction before this gets executed as part of pull response processing.
However, the approach has been changed now and this update has been reverted.
Both `NetworkSession` and `UnmanagedTransaction` are expected to let all pull handlers execute before completing the result cursors, meaning that neither commit or rollback transitions are expected before this status is marked. Transaction may be marked from both `RunResponseHandler` and `TransactionPullResponseCompletionListener`.

driver/src/test/java/org/neo4j/driver/integration/TransactionIT.java
injectives: Yes, this demonstrates the main goal of this update well.

driver/src/test/java/org/neo4j/driver/integration/BookmarkIT.java
```diff
@@ -133,10 +132,8 @@ void bookmarkRemainsAfterTxFailure()
         assertBookmarkContainsSingleValue( bookmark );
 
         Transaction tx = session.beginTransaction();
-        tx.run( "RETURN" );
 
-        ClientException e = assertThrows( ClientException.class, tx::commit );
-        assertNoCircularReferences( e );
```
gjmwoods, line 139: are some of these checks not needed anymore?

gjmwoods (approved):

driver/src/test/java/org/neo4j/driver/integration/BookmarkIT.java
injectives: As far as I remember, those were added for `commit` and `rollback` calls. Since this test does not use `commit` anymore, the check has been removed.

michael-simons (approved):
Thanks for the excellent walkthrough. I have tested the change afterwards with
- Neo4j-OGM 3.2.25-SNAPSHOT (no failures)
- Cypher-DSL main (no failures)
- Spring Data Neo4j 6.0.10-SNAPSHOT (one failure)

The failure was caused indirectly by our autoclosable query runner as it swallowed the target exception (see change here https://github.com/spring-projects/spring-data-neo4j/commit/dcd422d637effb944ae95189ef07af0d2ff0e756). 

We wouldn't have noticed the change in behavior otherwise, as we only consume via 

```java
try {
    var result = s.run();
    result.consume(); // or list, whatever
} catch(Whatever e) {
}
```

So in any case an exception would be handled.

Having said that, I would agree bringing this into the next patch.

PR #913, opened 2021-06-09
Add value coercion examples (#912)

bigmontz:


examples/src/main/java/org/neo4j/docs/driver/ReadingValuesExample.java
```diff
@@ -0,0 +1,180 @@
...
+    {
+        super( uri, user, password );
+    }
+
+    public Boolean nullIsNull()
+    {
+        return echo( null, record ->
+        {
+            // tag::java-driver-reading-values-null[]
+            Value nullValue = record.get( nullFieldName );
```
michael-simons: Just a suggestion on naming

`nullValue` -> `possibleNullValue`
`trueBoolean` -> `wasNull`

michael-simons (changes requested):
Hi. Nice examples.

On the `null` things I would go with a sightly different naming please.

michael-simons:
I just realized there is already one on 4.3… Maybe we adapt both of them?

michael-simons (approved):

PR #920, opened 2021-06-14
Upgrade netty to 4.1.65-Final

gjmwoods:


michael-simons (approved):
This time I ran it locally on GraalVM, too: Works as expected.

PR #923, opened 2021-06-14
Daemonize event loop threads.

michael-simons:
This instructs the `EventLoopGroupFactory` to daemonize newly created
event loop group threads as per Netty's `DefaultThreadFactory`. This
allows applications to shutdown when the last user thread ends, without
explicitly calling `close` to the Neo4j driver instance, which is in
line with other database drivers such as Postgres JDBC or MySQL JDBC.

In addition, this might become relevant with Netty 5 at some point in
the future, too. See: https://github.com/netty/netty/issues/2832.


michael-simons:
If this is needed at some point in the future, we might just open this up again. It's in the driver since 4.1.4 and this can be used as drop-in replacement for 4.0.x

PR #928, opened 2021-06-15
Make QueryRunner interface extend AutoCloseable

injectives:


michael-simons (approved):
I do appreciate that change a lot and for 4.3 this is great.

I wonder if we should - as you already hinted at - rethink this: 

image

Transactions and sessions are both query runners and both resources. The interface inheritance should be a straight line in 5.0

injectives:
Thank you 👍 We should consider removing `Resource` in the next major release unless a reasonable use-case is discovered.

PR #940, opened 2021-06-22
Introduce support for logging debug entries with Throwable

injectives:


michael-simons (approved):
Looks good to me. I briefly thought why not adding the new method as default method to the logger interface but then, the default would only be sensible for the dev null logger.
So in the end: Nothing to complain, I'm sorry.

PR #943, opened 2021-06-24
Remove stacktrace from recoverable discovery log warnings

injectives:
This is to reduce the noise in the logs when recoverable discovery exceptions occur. Complete discovery failures are expected to be reported separately.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/cluster/Rediscovery.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/cluster/Rediscovery.java
@@ -26,8 +26,21 @@ import org.neo4j.driver.Bookmark;
 import org.neo4j.driver.internal.BoltServerAddress;
 import org.neo4j.driver.internal.spi.ConnectionPool;
 
+/**
+ * Provides cluster composition lookup capabilities and initial router address resolution.
+ */
 public interface Rediscovery
 {
+    /**
+     * Fetches cluster composition using the provided routing table.
+     * <p>
+     * Implementation must be thread safe to be called with distinct routing tables concurrently. The routing table instance may be modified.
+     *
+     * @param routingTable   the routing table for cluster composition lookup
+     * @param connectionPool the connection pool for connection acquisition
+     * @param bookmark       the bookmark that is presented to the server
+     * @return cluster composition lookup result
+     */
     CompletionStage<ClusterCompositionLookupResult> lookupClusterComposition( RoutingTable routingTable, ConnectionPool connectionPool, Bookmark bookmark );
 
     List<BoltServerAddress> resolve() throws UnknownHostException;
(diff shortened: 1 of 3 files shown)


driver/src/main/java/org/neo4j/driver/internal/cluster/RediscoveryImpl.java
```diff
@@ -53,13 +53,15 @@
 import static org.neo4j.driver.internal.util.Futures.failedFuture;
 
 /**
- * This class is used by all router tables to perform discovery.
- * In other words, the methods in this class could be called by multiple threads concurrently.
+ * This class is used by all router tables to perform discovery. In other words, the methods in this class could be called by multiple threads concurrently.
```
michael-simons: Normally, I wouldn't complain, but as you touched it: Why not make this say: "The methods of this class must be thread safe as an instance of this is used by all router tables concurrently.."

michael-simons (changes requested):

driver/src/main/java/org/neo4j/driver/internal/cluster/RediscoveryImpl.java
```diff
@@ -291,8 +293,9 @@ private ClusterComposition handleRoutingProcedureError( Throwable error, Routing
         // Retriable error happened during discovery.
         DiscoveryException discoveryError = new DiscoveryException( format( RECOVERABLE_ROUTING_ERROR, routerAddress ), error );
         Futures.combineErrors( baseError, discoveryError ); // we record each failure here
-        logger.warn( format( "Received a recoverable discovery error with server '%s', will continue discovery with other routing servers if available.",
-                routerAddress ), discoveryError );
+        String warningMessage = format( RECOVERABLE_DISCOVERY_ERROR_WITH_SERVER, routerAddress );
+        logger.warn( warningMessage );
+        logger.debug( warningMessage, discoveryError );
```
michael-simons, line 295: It always feels kinda odd tbh to have a conditional debug log hidden away. Normally I would wrap this is in `if(debugIsEnabled)`, but our logger does that already internally…

driver/src/main/java/org/neo4j/driver/internal/cluster/RediscoveryImpl.java
injectives: I actually did not intentionally change it, the formatter did and I did not mind it formatting this block according to its rules.

injectives: But I can update this comment if you wish.

driver/src/main/java/org/neo4j/driver/internal/cluster/RediscoveryImpl.java
injectives: I actually expect loggers to skip this if debug is not enabled, but this is probably because I tended to work with SLF4J compliant loggers. In any case, our logger does that as you mentioned.

driver/src/main/java/org/neo4j/driver/internal/cluster/Rediscovery.java
```diff
@@ -26,8 +26,21 @@
...
 
+/**
+ * Provides cluster composition lookup capabilities and initial router address resolution.
+ */
 public interface Rediscovery
 {
+    /**
+     * Fetches cluster composition using the provided routing table.
+     * <p>
+     * Implementation must be thread safe to be called with distinct routing tables concurrently. The routing table instance may be modified.
```
michael-simons, line 37: 👍

michael-simons (approved):

PR #944, opened 2021-06-28
Remove stacktrace from connection acquisition attempts in LoadBalancer

injectives:
This is to reduce the noise in the logs on connection acquisition errors. Complete failures are reported separately.


gjmwoods (approved):
💪

PR #945, opened 2021-06-30
Move dependency versions to properties

injectives:


testkit-backend/pom.xml
```diff
@@ -35,8 +35,6 @@
         <dependency>
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
-            <version>1.18.20</version>
-            <scope>provided</scope>
```
fbiville, line 39: personal preference again: I usually let the scope visible

pom.xml
```diff
@@ -60,23 +81,17 @@
       <dependency>
         <groupId>org.reactivestreams</groupId>
         <artifactId>reactive-streams</artifactId>
-        <version>1.0.3</version>
+        <version>${reactive-streams.version}</version>
```
fbiville, line 84: personal preference: I'd only extract the versions that are common across a set of deps

pom.xml
```diff
@@ -85,27 +100,27 @@
       <dependency>
         <groupId>io.reactivex.rxjava2</groupId>
         <artifactId>rxjava</artifactId>
-        <version>2.2.21</version>
+        <version>${rxjava.version}</version>
```
fbiville, line 103: out of curiosity: why do we have RX Java AND the Reactor BOM?

fbiville (approved):

testkit-backend/pom.xml
injectives: The `svm` dependency is currently used in the same way. I feel like we should keep it consistent or have a separate PR to update them both.

pom.xml
injectives: `rxjava` is used in the examples and `reactor-bom` provides versions for `reactor-core` and `reactor-test` that are used in this project. I started working on moving away from bom to latest versions to fully match the server, but there are some methods that were removed and that work needs more time.

pom.xml
injectives: Given that we review versions quite often, I find it easier when they are consolidated in one section.

PR #955, opened 2021-07-08
Remove stacktrace from connection acquisition attempts in LoadBalancer

injectives:
Cherry-pick of #944.

This is to reduce the noise in the logs on connection acquisition errors. Complete failures are reported separately.


michael-simons (approved):
As @gjmwoods already checked the 4.3 I just compared the PRs.

PR #964, opened 2021-07-13
Added value conversion for char[] and short[]

DirkMahler:
A user of jQAssistant reported a problem when using a dedicated Neo4j instance (https://github.com/jQAssistant/jqa-core-framework/issues/60). 

The mapping framework behind jQAssistant (XO) sets properties of a node/relation using `Value.parameters` where the value is a Map of properties. In the described case one of the properties is of type `char[]` but testing revealed that `short[]` is not covered as well.

Creating the pull request against 1.6, the version used by jQAssistant actually is 4.3.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/v1/Values.java
+++ b/driver/src/main/java/org/neo4j/driver/v1/Values.java
@@ -117,11 +117,13 @@ public abstract class Values
         if ( value instanceof Iterable<?> ) { return value( (Iterable<Object>) value ); }
         if ( value instanceof Iterator<?> ) { return value( (Iterator<Object>) value ); }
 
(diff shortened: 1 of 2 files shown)


injectives:
Thank you for opening the pull request. 👍 

We are currently working on 4.3 branch and backporting changes to 4.2 and 4.1. 1.6 branch is no longer updated.

Given that the motivation came from the `char[]` support, I recommend updating to 4.3 driver, which seems to have it. In addition, 4.2 and 4.1 also seem to provide it.

DirkMahler:
The problem actually came up with 4.2 and still exists in 4.3. Shall I update this (or probably better create a new PR) for 4.3? I used 1.6 as target because it is still mentioned as active in https://github.com/neo4j/neo4j-java-driver#java-runtime.

PR #968, opened 2021-07-27
Add support for `connection.recv_timeout_seconds` connection hint

injectives:
This update brings support for the `connection.recv_timeout_seconds` connection hint that the server may provide in response to the `HELLO` Bolt message.

By providing this value, the server instructs the client on when it may consider connection broken when the client does not hear anything from the server for the specified duration. The client should drop the connection and remove the server from its routing table when it is running with routing enabled. The timeout is considered retryable by the transaction functions.

In addition, this update adds support for the `GetRoutingTable` Testkit request. The `ttl` field is not currently populated.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/async/inbound/ConnectionReadTimeoutHandler.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/async/inbound/ConnectionReadTimeoutHandler.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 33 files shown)


driver/src/main/java/org/neo4j/driver/internal/async/inbound/ConnectionReadTimeoutHandler.java
```diff
@@ -0,0 +1,47 @@
...
+    {
+        super( timeout, unit );
+    }
+
+    @Override
+    protected void readTimedOut( ChannelHandlerContext ctx )
+    {
+        if ( !triggered )
+        {
+            ctx.fireExceptionCaught( ConnectionReadTimeoutException.INSTANCE );
```
gjmwoods, line 42: might be done elsewhere and missed it but perhaps we want a debug log here to say the timeout has triggered

gjmwoods (approved):
🎨

driver/src/main/java/org/neo4j/driver/internal/async/inbound/ConnectionReadTimeoutHandler.java
injectives: As agreed, this does not seem to be necessary at the moment. Timeout is reported as exception to the user. In case of transaction function retries, it should be logged by the general logging there.

PR #987, opened 2021-08-17
Add Async Testkit Backend support

injectives:
This update brings async Testkit backend support that uses async driver for testing and is compatible with the current Testkit protocol. This allows async driver testing using the same Testkit tests transparently.

Not all requests are implemented in this PR. For instance, the resolver implementation will be added separately. The current update enables a substantial amount of tests to be executed.

In addition, a separate PR is planned to migrate the sync backend to the new Netty based implementation.


michael-simons (approved):
Approved as agreed on after the nice introduction.

bigmontz (commented):
🥇

PR #993, opened 2021-08-23
Decouple TestkitState from CommandProcessor.

michael-simons:
This commit remoces the command processor from the TestKit state. The processor is supposed to use that state, not to be part of it.
However, some requests, like `NewDriver` need the processor to trigger further state.
Therefor the processor is now provided as injectable value via Jackson, so that any request can indicate that it needs a processor via a constructor argument.


injectives (approved):
LGTM

PR #998, opened 2021-09-03
Add reactive backend support

injectives:
This update brings reactive backend support.

Its scope is limited to providing partial transparent support for existing test cases. More updates are expected in future PRs.


gjmwoods (approved):
🎺

PR #1003, opened 2021-09-09
Make reactive subscription request with Long.MAX_VALUE unbounded

injectives:
This update ensures that unbounded reactive request is translated to unbounded Bolt request.


michael-simons (approved):
Fine as of

> A Subscription MUST support an unbounded number of calls to request and MUST support a demand up to 2^63-1 (java.lang.Long.MAX_VALUE). A demand equal or greater than 2^63-1 (java.lang.Long.MAX_VALUE) MAY be considered by the Publisher as “effectively unbounded”.

See: https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.3/README.md#3-subscription-code

PR #1015, opened 2021-09-16
Prevent duplicate summary error reporting on session closure in reactive

injectives:
Reactive driver might emit the same error both on unconsumed result stream disposal and session closure, this update is intended to fix this.


michael-simons (approved):
Agreeing that a potential error should not necessary be propagated inside a `CompletionStage<>` (for future improvements).

PR #1029, opened 2021-10-08
Create a dedicated bundle / UberJAR module.

michael-simons:
This provides the driver UberJAR as it is known and used plus a slim deployment. It works by turning the previous main artifact into `neo4j-java-driver-slim` and the new bundle will be `neo4j-java-driver`.

The bundle depends on the slim artifact, extracts it and attaches it during the build so that a valid JavaDoc artifact (and sources artifact) are created and than shades the rest into it as before.

Care must be taken of the optional dependencies so that they are not included as well as the Reactive Streams API (so that it is a transitive dependency (the slim module is optional, otherwise the whole exercise would be in vain for downstream modules)).

This is an alternative (and IMHO) better approach than https://github.com/neo4j/neo4j-java-driver/pull/1024, thanks @injectives for pointing this out.


bundle/pom.xml
```diff
@@ -0,0 +1,189 @@
...
+    <plugins>
+      <!-- Extract the original sources again -->
+      <plugin>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>unpack-sources</id>
+            <phase>generate-sources</phase>
+            <goals>
+              <goal>unpack-dependencies</goal>
```
injectives: Need to discuss this bit when you are back.

injectives (approved):

gjmwoods (approved):
🚀

PR #1037, opened 2021-10-18
Replace AddressSet with standard Java collections

injectives:
This update also prevents unnecessary connection pool closures by keeping `disused` addresses until next routing table update and reusing them.


michael-simons (approved):
Thanks again for a nice collaboration.

PR #1065, opened 2021-11-10
Call close with the appropriate flag to commit or rollback on UnmanagedTransaction where possible to avoid double state acquisition

injectives:
Calling `close` instead of separate `isOpen` and `commitAsync` requires less lock acquisitions and is safer.


michael-simons (approved):
Maybe add to the commit message

> Call `close` with the appropriate flag to commit or rollback  on `UnmanagedTransaction` where possible to avoid double state aquisition.

Otherwise looks great.

PR #1067, opened 2021-11-10
Prevent duplicate summary error reporting on session closure in reactive

injectives:
Cherry-picks: #1015

Reactive driver might emit the same error both on unconsumed result stream disposal and session closure, this update is intended to fix this.


michael-simons (approved):
I like that you take the time to fix all of those reoccurring pattern. Well done.

injectives:
Thank you.

PR #1084, opened 2021-11-17
TestKit backend output goes to stdout/stderr

robsdedude:
From there, TestKit can pick it up and write it into logfiles if desired.
https://github.com/neo4j-drivers/testkit/pull/309


bigmontz (approved):
🎮

PR #1086, opened 2021-11-18
Update clirr configuration

injectives:


michael-simons (approved):
That is an easy enough fix. Maybe refer to the original PR #1029

PR #1088, opened 2021-11-22
Remove remaining stub tests and dependency on legacy stub server

injectives:
Reactive Testkit backend enabled a lot of Testkit tests to run on reactive driver. This update removes dependency on legacy stub server that is no longer maintained.

The remaining stub tests have been removed since there are similar tests in Testkit already, the tests and their respective Testkit alternatives are provided below:
- `DirectDriverBoltKitIT.shouldStreamingRecordsInBatchesRx` -> `TestIterationSessionRun.test_half_batch` - Both test result consumption in batches.
- `DirectDriverBoltKitIT.shouldDiscardIfPullNotFinished` -> `TestIterationSessionRun.test_discards_on_session_close` - Unlike the original, Testkit test consumes one record and discards the rest using session `run` intead of tx function. Based on its name, the original test purpose was to check if discard happens when result stream is not consumed in full and the suggested test covers this.
- `RoutingDriverBoltKitIT.shouldHandleLeaderSwitchAndRetryWhenWritingInTxFunctionRX` -> `RoutingV4x4.test_should_write_successfully_on_leader_switch_using_tx_function` - Both check leader failure handling.

One of examples that dependended on stub server has been brought in line with the rest of examples and uses `DatabaseExtension` to manage server.


driver/src/test/java/org/neo4j/driver/integration/RoutingDriverBoltKitIT.java
```diff
@@ -18,108 +18,35 @@
...
 
-/**
- * New tests should be added to testkit (https://github.com/neo4j-drivers/testkit).
- *
- * This class exists only for the following:
- * - to keep the remaining tests that are due to be migrated
- * - to keep the tests that are currently not portable
- */
-@Deprecated
 class RoutingDriverBoltKitIT
```
gjmwoods: Maybe we should rename this one to something like `ResolverFailureIT` or perhaps theres a suitable class to shift it to

gjmwoods (approved):
🎉

PR #1100, opened 2021-12-09
fix: Check for equality of value instead of equality of instance.

michael-simons:
isCustomized returned the wrong value for all instances that have been serialized and read back as they obviously does not point to the default instance.

The solution is to check for equality of the values, not for equality of the instances. This is a minor change in behaviour: A setting that is created with the builder matching the exact same defaults is now considered to be not customized.
This solution has been preferred over a custom readResolve returning the DEFAULT when things match as it would otherwise be able to create an instance that identifies itself as customized but changed to not customized when read back, basically the oppositte of what is fixed here: An instances created as not customized must never be read back as customized.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/SecuritySettings.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/SecuritySettings.java
@@ -21,6 +21,7 @@ package org.neo4j.driver.internal;
 import java.io.IOException;
 import java.io.Serializable;
 import java.security.GeneralSecurityException;
+import java.util.Objects;
 
 import org.neo4j.driver.Config;
 import org.neo4j.driver.exceptions.ClientException;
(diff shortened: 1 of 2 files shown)


driver/src/main/java/org/neo4j/driver/internal/SecuritySettings.java
```diff
@@ -59,7 +60,20 @@ public boolean encrypted()
 
     private boolean isCustomized()
     {
-        return this != DEFAULT;
+        return !(DEFAULT.encrypted() == this.encrypted() && DEFAULT.hasEqualTrustStrategy( this ));
```
mattcasters, line 63: Indeed this is the source of contention across serialization exercises. This change would fix that.

gjmwoods (approved):

PR #1103, opened 2021-12-14
Upgrade to Logback Classic 1.2.8 for test dependencies.

michael-simons:
Reason: https://jira.qos.ch/browse/LOGBACK-1591

It's test dependencies only, but still.

PR #1104, opened 2021-12-15
Upgrade to Logback Classic 1.2.8 for test dependencies.

michael-simons:
Forward cherry-picked from #1103. Forward because the 5.0 build was still broken yesterday and I wanted to have this in the 4.4 branch in case we do put out another (bug) fix release for it in time.


dotStart (approved):
:hourglass_flowing_sand:

PR #1117, opened 2022-01-18
Skip bookmark tests

robsdedude:
until returning no bookmarks on new sessions has been implemented.

Towards (but does not depend on) neo4j-drivers/testkit#364


robsdedude:
Obsolete, as Java already implements, what we agreed upon.

PR #1119, opened 2022-01-18
Add transaction close support to async and reactive APIs

injectives:


michael-simons (approved):
Good addition / unification. Thanks.

PR #1137, opened 2022-01-29
Micrometer metrics

injectives:


injectives:
@shakuzen, big thanks for your effort on this. 👍 
We were wondering if you could have a look at this update and let us know if you are happy to have it merged please?
It would be great if you could also sign the [CLA](https://neo4j.com/developer/cla/) agreement.

shakuzen:
Thank you for the mention and for getting this work in shape for merging. I am checking internally about the CLA and will get back to you.

driver/src/main/java/org/neo4j/driver/Config.java
```diff
@@ -95,6 +99,7 @@
     private final boolean isMetricsEnabled;
     private final int eventLoopThreads;
     private final String userAgent;
+    private final MeterRegistry meterRegistry;
```
michael-simons: Are we sure we can have the dependency as optional when we have direct imports / usages in our config?
I bet that will blow up, especially in Graal.

michael-simons (changes requested):
Can we rethink the optional dependency and go for provided?
Also I'm sceptic whether it will work as expected when Micrometer classes are on the `Config` class itself depending on the classloader used.

Options I see is using the Java service loader (define or use the existing metrics interface, and try to load it via build Java service loader, find either default or micrometer implementation) or something like I did here https://github.com/neo4j-contrib/cypher-dsl/blob/7c8ba2d4892c2933041fe7343d34c7f0a4b26f43/neo4j-cypher-dsl/src/main/java/org/neo4j/cypherdsl/core/Cypher.java#L1113 (this lets you use QueryDSL with CypherDSL).

I think the most suitable approach here would be service loader, tbh.

driver/src/main/java/org/neo4j/driver/Config.java
```diff
@@ -714,7 +723,37 @@ public ConfigBuilder withDriverMetrics()
...
         {
-            this.isMetricsEnabled = false;
+            return withMetricsEnabled( false );
+        }
+
+        private ConfigBuilder withMetricsEnabled( boolean enabled )
+        {
+            if ( !enabled )
+            {
+                this.metricsAdapter = MetricsAdapter.DEV_NULL;
```
meistermeier:
```suggestion
                return withMetricsAdapter( MetricsAdapter.DEV_NULL );
```

driver/src/main/java/org/neo4j/driver/Config.java
```diff
@@ -714,7 +723,37 @@ public ConfigBuilder withDriverMetrics()
...
+            }
+            else if ( this.metricsAdapter == null || this.metricsAdapter == MetricsAdapter.DEV_NULL )
+            {
+                this.metricsAdapter = MetricsAdapter.DEFAULT;
```
meistermeier:
```suggestion
                return withMetricsAdapter( MetricsAdapter.DEFAULT );
```

driver/src/main/java/org/neo4j/driver/ConnectionPoolMetrics.java
```diff
@@ -59,7 +60,7 @@
     /**
      * A counter to record how many connections have been successfully created with this pool since the pool is created.
      * This number increases every time when a connection is successfully created.
-     * @return The amount of connections have ever been created by this pool.
+     * @return The amount of connections has ever been created by this pool.
```
meistermeier:
```suggestion
     * @return The amount of connections that has ever been created by this pool.
```

driver/src/test/java/org/neo4j/driver/ConfigTest.java
```diff
@@ -370,7 +400,7 @@ void shouldErrorWithInvalidUserAgent()
         void shouldSerialize() throws Exception
         {
             Config config = Config.builder()
-                    .withMaxConnectionPoolSize( 123 )
+                                  .withMaxConnectionPoolSize( 123 )
```
meistermeier, line 401:
```suggestion
                    .withMaxConnectionPoolSize( 123 )
```

meistermeier (commented):
Do we have to mention somewhere which version(s) of Micrometer is/are supported?
Even though I like the style and would do the same I suggest a refactoring (**not** part of this PR) away from `/dev/null` (there is also logging) naming to something different, more understandable for non-neerdy code readers.

driver/src/test/java/org/neo4j/driver/ConfigTest.java
injectives: Updated formatting in https://github.com/neo4j/neo4j-java-driver/pull/1137/commits/d942e0deb9a5f26a6dda13fe2258723aea80980d.

driver/src/main/java/org/neo4j/driver/internal/metrics/DevNullMetricsListener.java
```diff
@@ -0,0 +1,100 @@
...
+
+    @Override
+    public void removePoolMetrics( String poolId )
+    {
+    }
+
+    @Override
+    public String toString()
+    {
+        return "Driver metrics are not available when they are not enabled.";
```
meistermeier:
```suggestion
        return "Driver metrics are not available if they are not enabled.";
```
?

driver/src/main/java/org/neo4j/driver/internal/metrics/MicrometerConnectionPoolMetrics.java
```diff
@@ -0,0 +1,277 @@
...
+    {
+        return format( "%s=[created=%s, closed=%s, creating=%s, failedToCreate=%s, acquiring=%s, acquired=%s, " +
+                       "timedOutToAcquire=%s, inUse=%s, idle=%s, " +
+                       "totalAcquisitionTime=%s, totalConnectionTime=%s, totalInUseTime=%s, totalInUseCount=%s]",
+                       id(), created(), closed(), creating(), failedToCreate(), acquiring(), acquired(),
+                       timedOutToAcquire(), inUse(), idle(),
+                       totalAcquisitionTime(), totalConnectionTime(), totalInUseTime(), totalInUseCount() );
+    }
+
+    private long toLong( Counter counter )
```
meistermeier:
```suggestion
    private long count( Counter counter )
```
or something similar, more "business logic" oriented?

meistermeier (approved):
Minor renaming. But if you think that it is better as it is right now, feel free to ignore it.

driver/src/main/java/org/neo4j/driver/internal/metrics/MicrometerConnectionPoolMetrics.java
```diff
@@ -0,0 +1,277 @@
...
+    {
+        failedToCreate.increment();
+        creating.decrementAndGet();
+    }
+
+    @Override
+    public void afterCreated( ListenerEvent<?> connEvent )
+    {
+        creating.decrementAndGet();
+        created.increment();
```
shakuzen: Timers include a count, so a separate Counter is not needed if it counts the same events being timed.

driver/src/main/java/org/neo4j/driver/internal/metrics/MicrometerConnectionPoolMetrics.java
```diff
@@ -0,0 +1,277 @@
...
+    @Override
+    public void afterAcquiringOrCreating()
+    {
+        acquiring.decrementAndGet();
+    }
+
+    @Override
+    public void afterAcquiredOrCreated( ListenerEvent<?> acquireEvent )
+    {
+        acquired.increment();
```
shakuzen: The acquired count can be obtained from the Timer for the same thing.

driver/src/main/java/org/neo4j/driver/internal/metrics/MicrometerConnectionPoolMetrics.java
```diff
@@ -0,0 +1,277 @@
...
+    @Override
+    public void acquired( ListenerEvent<?> inUseEvent )
+    {
+        inUseEvent.start();
+    }
+
+    @Override
+    public void released( ListenerEvent<?> inUseEvent )
+    {
+        released.increment();
```
shakuzen: Likewise, this can be derived from the Timer.

shakuzen (commented):
This is looking very nice. I see the `MicrometerMetricsProvider` can take an arbitrary `MeterRegistry` instance - this is good. How does a user configure things so a specific `MeterRegistry` is used?

driver/src/main/java/org/neo4j/driver/internal/metrics/MicrometerConnectionPoolMetrics.java
injectives: Good point, thanks.

driver/src/main/java/org/neo4j/driver/internal/metrics/MicrometerConnectionPoolMetrics.java
injectives: Thank you.

driver/src/main/java/org/neo4j/driver/internal/metrics/MicrometerConnectionPoolMetrics.java
injectives: Thanks.

injectives:
> This is looking very nice. I see the `MicrometerMetricsProvider` can take an arbitrary `MeterRegistry` instance - this is good. How does a user configure things so a specific `MeterRegistry` is used?

Conceptually, it can take any instance. However, at this iteration and for initial period we are only going to support the global registry.

@michael-simons, please correct me if I am wrong.

michael-simons (approved):
Looks good from my site, and yes, @injectives that was the intention: If this feature gets traction we shall revaluate custom registries.

meistermeier (approved):

PR #1148, opened 2022-02-08
Make MapAccessorWithDefaultValue extend MapAccessor

injectives:


michael-simons (approved):
Love it. But please don't extend or implement map. :)

PR #1153, opened 2022-02-11
Add support for multiple certificate files

injectives:
This update brings support for specifying multiple certificate files via `TrustStrategy#trustCustomCertificateSignedBy(File...)`. Also, it deprecates `TrustStrategy#certFiles()` that is superseded by `TrustStrategy#certFiles()`.

In addition, it adds support for the following Testkit feature flags:
- `Feature:API:SSLConfig`
- `Detail:DefaultSecurityConfigValueEquality`

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/Config.java
+++ b/driver/src/main/java/org/neo4j/driver/Config.java
@@ -21,6 +21,7 @@ package org.neo4j.driver;
 import java.io.File;
 import java.io.Serializable;
 import java.net.InetAddress;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
@@ -728,7 +729,7 @@ public class Config implements Serializable
         {
             Objects.requireNonNull( certFiles, "certFiles can't be null" );
             this.strategy = strategy;
-            this.certFiles = certFiles;
+            this.certFiles = Collections.unmodifiableList( new ArrayList<>( certFiles ) );
         }
 
         /**
--- a/driver/src/main/java/org/neo4j/driver/Config.java
+++ b/driver/src/main/java/org/neo4j/driver/Config.java
@@ -21,6 +21,9 @@ package org.neo4j.driver;
 import java.io.File;
 import java.io.Serializable;
 import java.net.InetAddress;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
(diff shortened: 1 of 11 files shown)


driver/src/test/java/org/neo4j/driver/ConfigTest.java
```diff
@@ -80,7 +80,7 @@ void shouldChangeToTrustedCert()
 
         // Then
         assertEquals( authConfig.strategy(), Config.TrustStrategy.Strategy.TRUST_CUSTOM_CA_SIGNED_CERTIFICATES );
-        assertEquals( trustedCert.getAbsolutePath(), authConfig.certFile().getAbsolutePath() );
+        assertEquals( trustedCert.getAbsolutePath(), authConfig.certFiles().get( 0 ).getAbsolutePath() );
```
michael-simons, line 83: We aren't using AssertJ? That makes those assertions so much nicer.

driver/src/main/java/org/neo4j/driver/Config.java
```diff
@@ -712,19 +715,20 @@ public Config build()
...
+            this( strategy, Collections.emptyList() );
         }
 
-        private TrustStrategy( Strategy strategy, File certFile )
+        private TrustStrategy( Strategy strategy, List<File> certFiles )
         {
+            Objects.requireNonNull( certFiles, "certFiles can't be null" );
             this.strategy = strategy;
-            this.certFile = certFile;
+            this.certFiles = certFiles;
```
michael-simons: Please make a defensive copy here.

driver/src/main/java/org/neo4j/driver/Config.java
```diff
@@ -741,10 +745,22 @@ public Strategy strategy()
...
+        }
+
+        /**
+         * Return the configured certificate files.
+         *
+         * @return configured certificate files or empty list if trust strategy does not require certificates.
+         */
+        public List<File> certFiles()
+        {
+            return certFiles;
```
michael-simons, line 764: Please return an immutable view here.

michael-simons (changes requested):

driver/src/main/java/org/neo4j/driver/Config.java
injectives: I've made `certFiles` unmodifiable.

driver/src/main/java/org/neo4j/driver/Config.java
injectives: Done.

driver/src/test/java/org/neo4j/driver/ConfigTest.java
injectives: Not sure, probably just because it started this way a while ago.
However, we can probably consider introducing it in some areas.

michael-simons (approved):

PR #1178, opened 2022-03-09
Update dependencies

injectives:


michael-simons (approved):
Ever thought that we invite dependabot?

PR #1181, opened 2022-03-11
Add support for Bolt 5

injectives:


driver/src/main/java/org/neo4j/driver/internal/messaging/common/CommonValueUnpacker.java
```diff
@@ -417,6 +418,14 @@ private static ZonedDateTime newZonedDateTime( long epochSecondLocal, long nano,
...
-}
 
+    protected int getNodeFields()
+    {
+        return NODE_FIELDS;
+    }
 
+    protected int getRelationshipFields()
+    {
+        return 5;
```
gjmwoods: nit: shall we go constant here too?

driver/src/main/java/org/neo4j/driver/types/Entity.java
```diff
@@ -28,12 +28,22 @@
...
+     * unique identifier is a better choice.
      *
      * @return the id of this entity
+     * @deprecated superseded by {@link #elementId()}
      */
+    @Deprecated
     long id();
+
+    /**
+     * A unique id for this Entity. Ids are guaranteed to remain stable for the duration of the session they were found in, but may be re-used for other
```
gjmwoods: Is the same still true for element Id?

gjmwoods (approved):
Looks great.

driver/src/main/java/org/neo4j/driver/types/Entity.java
gjmwoods: also shall we add a note to `id()` to say a value of `-1` indicates unsupported or something util we work it out

driver/src/main/java/org/neo4j/driver/internal/messaging/common/CommonValueUnpacker.java
injectives: Updated.

driver/src/main/java/org/neo4j/driver/types/Entity.java
injectives: Updated.

PR #1192, opened 2022-03-18
Add exceptions to numeric id accessors in nodes and relationships

injectives:
Depending on server configuration numeric id might not be available and accessing it will result in `IllegalStateException`.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/InternalEntity.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/InternalEntity.java
@@ -32,7 +32,8 @@ import static org.neo4j.driver.Values.ofObject;
 
 public abstract class InternalEntity implements Entity, AsValue
 {
-    public static final String INVALID_ID_ERROR = "Numeric id is not available, please use string based element id alternative";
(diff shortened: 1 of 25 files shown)


driver/src/main/java/org/neo4j/driver/internal/InternalEntity.java
```diff
@@ -32,20 +32,25 @@
 
 public abstract class InternalEntity implements Entity, AsValue
 {
+    public static final String INVALID_ID_ERROR = "Numeric id is not available, please use string based element id alternative";
```
gjmwoods: Did we settle on the content of the error message? I'm worried that refering to it as "Numeric" might be confusing. How about: "id() is not available for this storage engine, please use elementId() instead."

gjmwoods (approved):
🆔

driver/src/main/java/org/neo4j/driver/internal/InternalEntity.java
injectives: Updated the message. It needs to remain flexible enough to be used in multiple methods.

PR #1194, opened 2022-03-23
Add ability of checking if result handle is open

injectives:
This update introduces a new method to result handles for checking if result is open:
- `Result.isOpen()`
- `ResultCursor.isOpenAsync()`
- `RxResult.isOpen()`

Result is considered to be open if it has not been consumed via the consume method and its creator object (e.g. session or transaction) has not been closed (including committed or rolled back).

Attempts to access data on closed result will produce {@link ResultConsumedException}.

Implementation is based on existing internal logic.

### DIFF
--- a/driver/clirr-ignored-differences.xml
+++ b/driver/clirr-ignored-differences.xml
@@ -104,4 +104,22 @@
         <method>java.lang.String startNodeElementId()</method>
     </difference>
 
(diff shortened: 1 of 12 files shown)


michael-simons (approved):
LGTM (the all time classic, but we did actually discuss this).

PR #1203, opened 2022-04-05
Emit run errors on `RxResult.keys()`

injectives:
This update ensures that `RUN` errors are reported on `RxResult.keys()` instead of publishing an empty list.


injectives:
Superseded by https://github.com/neo4j/neo4j-java-driver/pull/1208.

PR #1208, opened 2022-04-12
Introduce new reactive session with updated API

injectives:
This update introduces a new reactive session called `ReactiveSession` and deprecates the existing `RxSession`.

The new session can be created via `reactiveSession` methods on `Driver`.

The `ReactiveSession` brings reactive API in line with synchronous and asynchronous when it comes to getting an instance on result handle (`ReactiveResult`). Specifically, `run` methods return a `Publisher<ReactiveResult>` that will only emit `ReactiveResult` instance on successful Bolt message (`RUN`) exchange and will emit an error on failure. Furthermore, this change allowed simplification of `ReactiveResult.keys()` method that simply returns a list of strings now.

This update also adds a new reactive backend implementation for Testkit and renames existing implementation to `reactive-legacy`.


michael-simons (approved):
I find everything that we spoke about… Looks great.

gjmwoods (approved):

PR #1211, opened 2022-04-29
Update bookmarks API

injectives:
`Bookmark` objects should represent singular bookmark values. Multiple bookmarks will be represented as a set of bookmarks (`Set<Bookmark>`).

The main API updates are highlighted below.

`Bookmark` updates:
- `value()` - new method
- `values()` - deprecated
- `isEmpty()` - deprecated

`Session` and `AsyncSession` updates:
- `Set<Bookmark> lastBookmarks()` - new method
- `Bookmark lastBookmark()` - deprecated

`ReactiveSession` updates:
- `Set<Bookmark> lastBookmarks()` - new method
- `Bookmark lastBookmark()` - deprecated (has not been released)


driver/src/main/java/org/neo4j/driver/Bookmark.java
```diff
@@ -35,17 +36,39 @@
...
+        return InternalBookmark.parse( Collections.singleton( value ) );
+    }
+
+    /**
+     * Reconstruct bookmark from bookmarks string values.
+     *
      * @param values values obtained from a previous bookmark.
      * @return A bookmark.
      */
+    @Deprecated
```
michael-simons, line 71: The only thing affected under "my" control is OGM and that's good. Proper clear API here.

driver/src/main/java/org/neo4j/driver/Session.java
```diff
@@ -317,18 +318,23 @@ default void executeWriteWithoutResult( Consumer<TransactionContext> contextCons
...
     Bookmark lastBookmark();
 
     /**
-     * Signal that you are done using this session. In the default driver usage, closing and accessing sessions is
-     * very low cost.
+     * Return an immutable set of bookmarks known by this session.
+     *
+     * @return the set of bookmarks.
+     */
+    Set<Bookmark> lastBookmarks();
```
michael-simons, line 339: I am very much not happy about that change as I expressed several times (trello and standup when I remember correctly). This will break us in interesting and unnecessary ways.

Technically, using a new method name is correct. But the concept is bad.

michael-simons (commented):
Really not a fan about returning multiple bookmarks, I still don't understand the technical reasons behind it.

michael-simons (approved):
As discussed in person.

driver/src/main/java/org/neo4j/driver/Session.java
injectives: Just to note, we had more conversations on this topic.

PR #1215, opened 2022-05-09
Migrate RoutingExamplesIT to Testcontainers

injectives:


michael-simons (commented):
Hmm… This isn't doing a real cluster connection anymore. Is that on purpose?

If so, ignore my feedback and move on.
If not, we have a JUNit extension that actually brings up a cluster: https://github.com/michael-simons/junit-jupiter-causal-cluster-testcontainer-extension

injectives:
It is a single node cluster. I have discussed this with @gjmwoods and it has been concluded that it should be sufficient for these particular cases.

injectives:
It is a single node cluster. This should be sufficient for these tests and will make build a bit lighter.

injectives:
I have seen the exciting extension that we might use when we need a multinode cluster. 👍

michael-simons (approved):

driver/src/test/java/org/neo4j/driver/stress/CausalClusteringStressIT.java
```diff
@@ -30,6 +31,7 @@
 import org.neo4j.driver.exceptions.SessionExpiredException;
 import org.neo4j.driver.util.cc.LocalOrRemoteClusterExtension;
 
+@Testcontainers( disabledWithoutDocker = true )
```
gjmwoods, line 34: do we really want silent failure if docker isn't available?

injectives: It would skip, not fail.
This is currently needed for our Windows builds.

PR #1222, opened 2022-05-16
Stop shading dependencies in neo4j-java-driver package

injectives:
`neo4j-java-driver` package no longer includes shaded dependencies.

The new `neo4j-java-driver-all` package includes shaded Netty and Project Reactor dependencies.


michael-simons (approved):
Looks good so far. I am surprised we don’t need to rearrange things more in poms to flip the switch. But than, cool that it works.

PR #1245, opened 2022-06-10
Wrap checked exceptions when rethrowing async exception

injectives:
Driver may throw checked exception without declaring it. This update fixes this behaviour.


michael-simons (approved):
Sweet. I'm glad you didn't forget about this.

michael-simons (approved):
Makes sense to me. Thanks.

PR #1247, opened 2022-06-15
Add support for sending datetime values in UTC

injectives:
This update brings 2 new Bolt structures for transmitting datetime in UTC, specifically tag bytes 49 and 69.

Bolt 5 will only support these structures and will no longer accept tag bytes 46 and 66. Bolt 4.4 and 4.3 will negotiate with server and use the new structures if possible. Previous protocol versions behaviour remains unchanged.


michael-simons (approved):
Essential fix, all good and inline with ADR-05.

Let's take a mental note that we might thinking about a more dedicated `Patch` structure than single arguments in case we need more patches in the future to come.

PR #1256, opened 2022-06-28
Migrate tests depending on Boltkit to use Docker

injectives:
Boltkit has been deprecated and requires additional setup on the host to run tests. This update migrates tests depending on Boltkit to use Docker that is needed for Testkit testing as well.


README.md
```diff
@@ -97,19 +97,18 @@ To use the driver in a project, please use the released driver via Maven Central
...
-Testkit is a tooling that is used to run integration tests for all of the drivers we provide.
-
-Some older tests still rely on [`boltkit`](https://github.com/neo4j-drivers/boltkit), the predecessor to Testkit.
-If `boltkit` is not installed, then all tests that requires `boltkit` will be ignored and will not be executed.
-
-In case you want or can verify contributions with unit tests alone, use the following command to skip all integration and Testkit tests:
-
+The following command may be used to unit test and install artifacts without running integration tests:
 ```
 mvn clean install -DskipITs -P skip-testkit
```
michael-simons, line 102: This still complains about missing connection in 

```
[INFO] Tests run: 8, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 33.129 s - in 
org.neo4j.driver.GraphDatabaseTest
```


```
org.neo4j.driver.exceptions.ServiceUnavailableException: Unable to connect to database management service, ensure the database is running and that there is a working network connection to it.
```

Takes also rather long.

driver/src/test/java/org/neo4j/driver/util/DatabaseExtension.java
```diff
@@ -19,121 +19,227 @@
...
+            key = new File("");
+            network = null;
+            nginx = new GenericContainer<>(DockerImageName.parse("alpine:latest"));
+        }
     }
 
-    public DatabaseExtension(Neo4jSettings settings) {
-        this.settings = settings;
+    @Override
+    public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
```
michael-simons, line 125: Testcontainers provides 

```
@Testcontainers(disabledWithoutDocker = true)
```

I have the impression that a lot of this extension can be replaced by already existing ones. Did you look into this?

michael-simons (commented):
I tried running the full suite and it seems to work.  `org.neo4j.driver.stress.CausalClusteringStressIT` however takes an awful lot of time. 

In general, I am a big fan of getting rid of Boltkit, and having everything orchestrated from Maven. However, I was wondering why those tests are needed at all and wether they can't be orchestrated from Testkit rather than from the driver site. @robsdedude reminded me that Testkit can bring up concrete instances, too. And I am asking this as a big fan of Testcontainers/TestcontainersNeo4j.

+Edit+

I read this

> The general plan is to migrate integration tests to Testkit, but there are many of them and it would take some time. In addition, not all manipulations are currently supported in Testkit.

And my question still applies: Do we need the manipulations from Testkits side or can't we just rely on getting the correct containers?

driver/src/test/java/org/neo4j/driver/util/DatabaseExtension.java
injectives: Yes, I know about this. My intention was to minimise touched files, so I chose this approach. It should be a temporary solution until tests are migrated to Testkit.

README.md
injectives: This is a separate problem that is unrelated to this update. Shall we tackle it separately?

injectives:
> I tried running the full suite and it seems to work. `org.neo4j.driver.stress.CausalClusteringStressIT` however takes an awful lot of time.

Yes, `CausalClusteringStressIT` takes time. This update does not change this. We can look into putting it on a separate profile if we find this is necessary.

injectives:
> In general, I am a big fan of getting rid of Boltkit, and having everything orchestrated from Maven. However, I was wondering why those tests are needed at all and wether they can't be orchestrated from Testkit rather than from the driver site. @robsdedude reminded me that Testkit can bring up concrete instances, too. And I am asking this as a big fan of Testcontainers/TestcontainersNeo4j.

I think the tests are needed at least until we go over them and confirm that such cases are covered by Testkit. Those that are not covered will need to be migrated to Testkit and the ones that are covered will be decommissioned. During migration we usually determine if test case is better off tested with stub server or real one in Testkit.

This update simply lets us move away from deprecated Boltkit until we get to the steps described above.

injectives:
> And my question still applies: Do we need the manipulations from Testkits side or can't we just rely on getting the correct containers?

As far as I remember, Testkit does not currently support server state manipulations. For instance,  stopping, starting, updating configuration or swapping ssl certificates. Some existing integration tests manipulate server state. Whether it makes sense to bring support to Testkit OR change test logic should be determined during migration to Testkit.

README.md
michael-simons: Yeah. I wasn't sure, but since I was testing the build, I mentioned it.

michael-simons (approved):

PR #1265, opened 2022-07-11
Update native-image.properties

injectives:


michael-simons (approved):
Sorry for the hassles, I wonder why this  isn't caught by the test project (I wasnt able to reproduce it there)

PR #1267, opened 2022-07-13
Introduce updateRoutingTableTimeout option

injectives:
Sets maximum amount of time the driver may wait for routing table acquisition.

This option allows setting API response time expectation. It does not limit the time the driver might need when getting routing table.


bigmontz (approved):
⏲️ 💣

PR #1271, opened 2022-07-19
Reduce build log sizes by removing download progress stats

fbiville:


fbiville:
The build log is currently almost at 2MB.
Most of the content is the download stat display.

injectives (approved):

PR #1273, opened 2022-07-20
Defer unknown zone information failures to datetime value access stage

injectives:
If server supplies zone information that is not supported by driver runtime, it will be represented by an unknown datetime value that fails on datetime access. This prevents an immediate failure during data consumption and allows usage of other values.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/messaging/common/CommonValueUnpacker.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/messaging/common/CommonValueUnpacker.java
@@ -24,6 +24,7 @@ import static org.neo4j.driver.Values.point;
 import static org.neo4j.driver.Values.value;
 
 import java.io.IOException;
+import java.time.DateTimeException;
 import java.time.Instant;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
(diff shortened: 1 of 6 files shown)


driver/src/main/java/org/neo4j/driver/internal/messaging/common/CommonValueUnpacker.java
```diff
@@ -374,34 +377,26 @@ private Value unpackLocalDateTime() throws IOException {
...
-        return value(newZonedDateTime(epochSecondLocal, nano, ZoneId.of(zoneIdString)));
-    }
-
-    private Value unpackDateTimeUtcWithZoneId() throws IOException {
-        long epochSecondLocal = unpacker.unpackLong();
-        int nano = Math.toIntExact(unpacker.unpackLong());
-        String zoneIdString = unpacker.unpackString();
-        ZoneId zoneId = ZoneId.of(zoneIdString);
-        return value(newZonedDateTimeUsingUtcBaseline(epochSecondLocal, nano, zoneId));
+    private Value unpackDateTime(boolean unpackOffset, boolean useUtcBaseline) throws IOException {
```
bigmontz: nitpick: the usage of this method like `unpackDateTime(false, true)` doesn't give much context about the expected behaviour.

bigmontz (approved):
🕙 🕦

driver/src/main/java/org/neo4j/driver/internal/messaging/common/CommonValueUnpacker.java
injectives: This is just an internal small method with 2 arguments, but I have added enums for readability.

PR #1275, opened 2022-07-26
Enable compilation failures on warnings for driver module main sources

injectives:
This is primarily an extra protection from unintended warnings.


michael-simons (approved):
Good choice.

PR #1282, opened 2022-08-03
Deprecate RevocationStrategy and introduce RevocationCheckingStrategy

injectives:
The `RevocationStrategy` is an internal type, it has been superseded by `RevocationCheckingStrategy`.


michael-simons (approved):
Good naming choice. Thanks for taking my concerns into consideration.

PR #1295, opened 2022-08-24
Update the new reactive API to use Flow API

injectives:
The new reactive API should use Java Flow API, the legacy deprecated reactive API stays on Reactive Streams API.


michael-simons (approved):
Not reviewed in detail, but tested with the SDN7 5.0.x variant, works as before. Go for it!

PR #1301, opened 2022-08-31
Turn off hostname verification for trustAllCertificates TrustStrategy

injectives:
`Config.TrustStrategy.trustAllCertificates()` should return `TrustStrategy` with hostname verification turned off by default.

This update is part of behaviour unification among the official drivers.


AndyHeap-NeoTech (approved):
The changes look good. Should the security teams advice change the behaviour can easily be re-enabled.

PR #1311, opened 2022-09-13
TestKit Reactive Backend: skip test auto commit error on session close

robsdedude:
Respective test introduced in https://github.com/neo4j-drivers/testkit/pull/522


robsdedude:
I don't think this work will be completed. It's a breaking change that needs coordination between drivers. Maybe we'll come back to it should it every become an actual problem.

PR #1312, opened 2022-09-19
Delete superflous subsitutions.

michael-simons:
As GraalVM native image switched the default to initialize mostly everything to runtime initialization by default (apart from known safe classes https://www.graalvm.org/22.1/reference-manual/native-image/ClassInitialization/#automatic-initialization-of-safe-classes), we don’t have to cater for the default in the driver. The rest is taking care of by the recent netty versions.

In Quarkus we must defer Netty SSL explicitly to runtime (`return new RuntimeInitializedPackageBuildItem("io.netty.handler.ssl“)`), as Quarkus insists of Built-Time-Initialization by default and our SSL config will than trigger the above at a point in time in which the Quarkus built-in substitions are not in place yet.


michael-simons:
Please update TC to use `ghcr.io/graalvm/native-image:ol8-java17-22.2.0` for the verification.

injectives (approved):

PR #1321, opened 2022-10-20
Introduce executeQuery on Driver level

injectives:
This is a new basic high-level API for executing idempotent queries.


driver/src/main/java/org/neo4j/driver/Config.java
```diff
@@ -275,6 +293,19 @@ public static final class ConfigBuilder {
...
         private ConfigBuilder() {}
 
+        /**
+         * Sets a {@link BookmarkManager} implementation for the driver to use on
+         * {@link Driver#executeQuery(Query, QueryConfig)} method and its variants by default.
+         *
+         * @param queryBookmarkManager bookmark manager, must not be {@code null}
+         * @return this builder
+         */
+        public ConfigBuilder withQueryBookmarkManager(BookmarkManager queryBookmarkManager) {
```
gjmwoods, line 309: thinking about the chances users will think configuring this BookmarkManager will apply for the entire driver. Not sure how/if we can reduce that risk.

driver/src/main/java/org/neo4j/driver/Driver.java
```diff
@@ -63,6 +66,109 @@
  * @since 1.0 (Modified and Added {@link AsyncSession} and {@link RxSession} since 2.0)
  */
 public interface Driver extends AutoCloseable {
+    /**
+     * Execute an idempotent query in a managed transaction with automatic retries on retryable errors.
+     * <p>
+     * This is a basic high-level API for executing idempotent queries. {@link #executeQuery(Query, QueryConfig)}
+     * documentation provides more details.
+     *
```
gjmwoods, line 74: perhaps a note back to bookmark configuration and a note about causal consistency guarantees similar to the one on `Session`

gjmwoods (approved):

driver/src/main/java/org/neo4j/driver/Config.java
injectives: A note on sessions has been added.

driver/src/main/java/org/neo4j/driver/Driver.java
injectives: Updated. Please take a look if it matches you expectations.

gjmwoods (approved):

PR #1323, opened 2022-10-22
Add getInstance to TypeSystem

injectives:


michael-simons (approved):
Thanks

(Maybe `getDefault()` or `getDefaultInstance()` is a slightly better name just in case we ever should have another one)

PR #1325, opened 2022-10-25
Introduce new Reactive Streams session

injectives:
This update introduces support for reactive session with Reactive Streams types. While it is similar to the deprecated `RxSession`, it includes the improvements introduced with the `ReactiveSession` that uses Flow API types.

Sample session creation:
```
var session = driver.reactiveSession(org.neo4j.driver.reactivestreams.ReactiveSession.class);
```

### DIFF
--- a/driver/clirr-ignored-differences.xml
+++ b/driver/clirr-ignored-differences.xml
@@ -397,4 +397,16 @@
         <method>org.neo4j.driver.types.TypeSystem getDefault()</method>
     </difference>
 
+    <difference>
+        <className>org/neo4j/driver/Driver</className>
+        <differenceType>7012</differenceType>
+        <method>org.neo4j.driver.BaseReactiveSession reactiveSession(java.lang.Class)</method>
+    </difference>
+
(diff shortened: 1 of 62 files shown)


michael-simons (approved):
I want this. Good stuff. 

Now think to just cut down the `.session()` methods to two (untyped gives you imperative, typed gives you whatever you want). Less is more.

For reference: https://github.com/neo4j-contrib/neo4j-http/commit/6d47ae3c5cba1a1db840afa5ad365e49df256062

injectives:
> I want this. Good stuff.
> 
> Now think to just cut down the `.session()` methods to two (untyped gives you imperative, typed gives you whatever you want). Less is more.
> 
> For reference: [neo4j-contrib/neo4j-http@6d47ae3](https://github.com/neo4j-contrib/neo4j-http/commit/6d47ae3c5cba1a1db840afa5ad365e49df256062)

As we have discussed, I explored updating the session methods and will submit a separate PR for review.

PR #1326, opened 2022-10-26
Introduce new session instantiation methods

injectives:
These new methods provide a consistent API for creating different types of session APIs.

Sample usage:
```
var synchronousSession = driver.session(Session.class);
var asynchronousSession = driver.session(AsyncSession.class);
var reactiveSession1 = driver.session(org.neo4j.driver.reactive.ReactiveSession.class);
var reactiveSession2 = driver.session(org.neo4j.driver.reactivestreams.ReactiveSession.class);
var reactiveSession3 = driver.session(org.neo4j.driver.reactive.RxSession.class);
```

The following methods have been deprecated:
- `asyncSession`
- `reactiveSession`

### DIFF
--- a/driver/clirr-ignored-differences.xml
+++ b/driver/clirr-ignored-differences.xml
@@ -400,13 +400,13 @@
     <difference>
         <className>org/neo4j/driver/Driver</className>
         <differenceType>7012</differenceType>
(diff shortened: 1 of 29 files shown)


michael-simons (approved):
Conceptional great, haven’t look at the implementation. I think it reduces the spread out surface a lot without sacrificing functionality.

PR #1330, opened 2022-10-31
Introduce managed query execution on Driver level

injectives:
This is a new basic high-level API for executing idempotent queries.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/Config.java
+++ b/driver/src/main/java/org/neo4j/driver/Config.java
@@ -75,6 +75,8 @@ public final class Config implements Serializable {
 
     private static final Config EMPTY = builder().build();
 
+    private final BookmarkManager queryBookmarkManager;
+
     /**
      * User defined logging
      */
@@ -102,6 +104,7 @@ public final class Config implements Serializable {
     private final MetricsAdapter metricsAdapter;
 
     private Config(ConfigBuilder builder) {
+        this.queryBookmarkManager = builder.queryBookmarkManager;
         this.logging = builder.logging;
         this.logLeakedSessions = builder.logLeakedSessions;
 
@@ -121,6 +124,21 @@ public final class Config implements Serializable {
 
         this.eventLoopThreads = builder.eventLoopThreads;
         this.metricsAdapter = builder.metricsAdapter;
+    }
+
+    /**
+     * A {@link BookmarkManager} implementation for the driver to use on
+     * {@link Driver#queryTask(String)} method and its variants by default.
+     * <p>
+     * Please note that sessions will not use this automatically, but it is possible to enable it explicitly
+     * using {@link SessionConfig.Builder#withBookmarkManager(BookmarkManager)}.
+     *
+     * @return bookmark manager, must not be {@code null}
+     * @since 5.5
+     */
+    @Experimental
+    public BookmarkManager queryBookmarkManager() {
+        return queryBookmarkManager;
     }
 
     /**
@@ -262,6 +280,8 @@ public final class Config implements Serializable {
      * Used to build new config instances
      */
     public static final class ConfigBuilder {
+        private BookmarkManager queryBookmarkManager =
+                BookmarkManagers.defaultManager(BookmarkManagerConfig.builder().build());
         private Logging logging = DEV_NULL_LOGGING;
         private boolean logLeakedSessions;
         private int maxConnectionPoolSize = PoolSettings.DEFAULT_MAX_CONNECTION_POOL_SIZE;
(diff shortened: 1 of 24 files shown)


driver/src/main/java/org/neo4j/driver/ResultTransformer.java
```diff
@@ -0,0 +1,43 @@
...
+     */
+    void accept(Record record);
+
+    /**
+     * Accepts a {@link ResultSummary} value and finishes the transformation by producing a final value of the target type.
+     *
+     * @param summary the result summary
+     * @return the final value of the target type
+     */
+    T finish(ResultSummary summary);
```
michael-simons: I would default that method.

driver/src/main/java/org/neo4j/driver/ResultTransformerProvider.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.neo4j.driver;
+
+import java.io.Serializable;
+
+@FunctionalInterface
+public interface ResultTransformerProvider<T> extends Serializable {
```
michael-simons: I would not introduce that interface and instead of go with `Supplier`.

I see the reason for a new interface (you want to make clear it's a new instance), but this is something one can suggest but not enforce.
Imagine a scenario in which people want to run multiple queries to the transformer.

driver/src/main/java/org/neo4j/driver/ResultTransformerProviders.java
```diff
@@ -0,0 +1,55 @@
...
+import static org.neo4j.driver.internal.CollectorResultTransformerProvider.EagerResultSummaryHandler;
+import static org.neo4j.driver.internal.CollectorResultTransformerProvider.PassThroughSummaryHandler;
+import static org.neo4j.driver.internal.CollectorResultTransformerProvider.RecordsCollectorProvider;
+
+import java.io.Serializable;
+import java.util.stream.Collector;
+import org.neo4j.driver.internal.CollectorResultTransformerProvider;
+import org.neo4j.driver.summary.ResultSummary;
+
+public class ResultTransformerProviders {
```
michael-simons: `ResultTransformers` is a better name.

The methods would return `Supplier<?>`.

driver/src/test/java/org/neo4j/driver/QueryConfigTest.java
```diff
@@ -0,0 +1,181 @@
...
+                originalConfig.resultTransformerProvider().getClass(),
+                deserializedConfig.resultTransformerProvider().getClass());
+        assertEquals(originalConfig.database(), deserializedConfig.database());
+        assertEquals(originalConfig.impersonatedUser(), deserializedConfig.impersonatedUser());
+        assertEquals(
+                originalConfig.bookmarkManager(defaultManager), deserializedConfig.bookmarkManager(defaultManager));
+    }
+
+    @Test
+    void test() {
```
michael-simons: Excellent stuff! 👍

michael-simons (changes requested):

michael-simons (changes requested):
I agree on your slack remark this being too many overloads. However, it feels more natural to me not having the transformer as part of the config.
I would most like go with
- Having only `Query` objects in the simple api
- Adding `of` method to `Query` class, such as `Query.of(cypher)` and `Query.of(cypher, params)`, so that it is still nicely readable

I don't really understand the "Having a single object that would capture both the collector and the result summary handler." remark, but I feel like I don't want this, it's more cruft.

michael-simons:
Oh, adding some heresy… `execute(query)` instead of executeQuery(query)` :D

michael-simons (approved):
Looks very good to me. I like the naming choice.

driver/src/main/java/org/neo4j/driver/Config.java
```diff
@@ -281,6 +301,24 @@ public static final class ConfigBuilder {
...
+         * <p>
+         * Please note that sessions will not use this automatically, but it is possible to enable it explicitly
+         * using {@link SessionConfig.Builder#withBookmarkManager(BookmarkManager)}.
+         *
+         * @param queryBookmarkManager bookmark manager, must not be {@code null}
+         * @return this builder
+         * @since 5.5
+         */
+        @Experimental
+        public ConfigBuilder withQueryBookmarkManager(BookmarkManager queryBookmarkManager) {
```
gjmwoods: `withQueryTaskBookmarkManager()`?

driver/src/main/java/org/neo4j/driver/Config.java
```diff
@@ -123,6 +126,21 @@ private Config(ConfigBuilder builder) {
...
+     * <p>
+     * Please note that sessions will not use this automatically, but it is possible to enable it explicitly
+     * using {@link SessionConfig.Builder#withBookmarkManager(BookmarkManager)}.
+     *
+     * @return bookmark manager, must not be {@code null}
+     * @since 5.5
+     */
+    @Experimental
+    public BookmarkManager queryBookmarkManager() {
+        return queryBookmarkManager;
```
gjmwoods, line 141: `queryTaskBookmarkManager()`?

gjmwoods (approved):

driver/src/main/java/org/neo4j/driver/Config.java
injectives: Updated. 👍

driver/src/main/java/org/neo4j/driver/Config.java
injectives: Updated. 👍

PR #1332, opened 2022-11-08
Update SessionConfig.Builder.withDatabase documentation

injectives:
This update is intended to make this option and its implications better documented.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/SessionConfig.java
+++ b/driver/src/main/java/org/neo4j/driver/SessionConfig.java
@@ -76,9 +76,10 @@ public final class SessionConfig implements Serializable {
     }
 
     /**
-     * Returns a {@link SessionConfig} for the specified database
+     * Returns a {@link SessionConfig} for the specified database name.
      * @param database the database the session binds to.
      * @return a session config for a session for the specified database.
+     * @see SessionConfig.Builder#withDatabase(String)
      */
     public static SessionConfig forDatabase(String database) {
         return new Builder().withDatabase(database).build();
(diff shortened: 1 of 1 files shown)


gjmwoods (approved):

driver/src/main/java/org/neo4j/driver/SessionConfig.java
```diff
@@ -246,16 +247,38 @@ public Builder withDefaultAccessMode(AccessMode mode) {
         }
 
         /**
-         * Set the database that the newly created session is going to connect to.
+         * Set database name for queries to be executed within session. This also applies to downstream query
+         * executions, like transactions started by session.
```
robsdedude, line 251:
```suggestion
         * executions, like transactions started by the session.
```

driver/src/main/java/org/neo4j/driver/SessionConfig.java
```diff
@@ -246,16 +247,38 @@ public Builder withDefaultAccessMode(AccessMode mode) {
...
          * <p>
-         * For connecting to servers that support multi-databases,
-         * it is highly recommended to always set the database explicitly in the {@link SessionConfig} for each session.
-         * If the database name is not set, then session defaults to connecting to the default database configured in server configuration.
+         * This option has no explicit value by default.
          * <p>
-         * For servers that do not support multi-databases, leave this database value unset. The only database will be used instead.
+         * When no explicit database name is set, driver behaviour depends on connection URI scheme supplied to driver
```
robsdedude, line 255:
```suggestion
         * When no explicit database name is set, driver behaviour depends on the connection URI scheme supplied to driver
```

driver/src/main/java/org/neo4j/driver/SessionConfig.java
```diff
@@ -246,16 +247,38 @@ public Builder withDefaultAccessMode(AccessMode mode) {
...
+         *     fetching for the session, all queries within that session are executed with the explicit database name
+         *     'movies' supplied. If server changes database name for the user, it is only used by subsequent sessions.
+         *     This behaviour requires additional network communication and it is important to ensure that connection
+         *     URI supplied to driver resolves to multiple server endpoints in clustered environments when possible.
+         *     For older Bolt protocol versions the behaviour is the same as described for the <b>bolt</b> scheme above.
+         *     </li>
+         * </ul>
+         * <p>
+         * Setting explicit database name when possible has a benefit of potentially optimising network communication
+         * and ensuring consistent target database name for session.
```
robsdedude, line 277: I would put this part above the detailed explanation. So even if users are too lazy to read all of that, they'd (hopefully) at least pick up on that. Maybe even emphasize it, if javadocs allows for that.

driver/src/main/java/org/neo4j/driver/SessionConfig.java
injectives: It is actually a reference to session in a general sense, which is why `the` is not used. I will see if this can be expressed in a better way.

driver/src/main/java/org/neo4j/driver/SessionConfig.java
injectives: Good point.

PR #1336, opened 2022-11-24
Update SessionConfig.withDatabase documentation (addition)

injectives:


driver/src/main/java/org/neo4j/driver/SessionConfig.java
```diff
@@ -253,6 +253,8 @@ public Builder withDefaultAccessMode(AccessMode mode) {
          * in a straightforward way and potentially simplifies driver logic as well as reduces network communication
          * resulting in better performance.
          * <p>
+         * Usage of Cypher clauses like USE is not a replacement for this option. Driver sends Cypher to the server for processing.
```
robsdedude, line 256:
```suggestion
         * Usage of Cypher clauses like USE is not a replacement for this option. The driver does not parse any Cypher.
```

maybe.

bigmontz (approved):
🥇

driver/src/main/java/org/neo4j/driver/SessionConfig.java
injectives: This actually got paraphrased here: https://github.com/neo4j/neo4j-java-driver/pull/1337

PR #1341, opened 2022-11-30
Release connection on reactive beginTransaction cancellation

injectives:
Each transaction created by the driver requires a network connection. Unfinished transactions may result in connection leaks, meaning that connections acquired from the connection pool are not available for further use.

Subscription cancellation on reactive `beginTransaction` during transaction creation could result in dangling transaction, leading to the connection leak. This update fixes this issue by ensuring that such transactions are rolled back and their connections are returned to the connection pool.


michael-simons:
Excellent work!

image

michael-simons (approved):

PR #1348, opened 2022-12-07
Use moditect plugin to re-add `module-info` after it has been nuked by the shade plugin.

michael-simons:
The shade plugin will not only warn about shading breaking encapsulation, it will also remove existing modules inside the project that does the shading, without any option not doing this. We using the moditect plugin to re-add it after it has been nuked. The shaded dependencies have been omitted from the module declaration of course, as they included.

### DIFF
--- a/bundle/pom.xml
+++ b/bundle/pom.xml
@@ -218,6 +218,43 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.moditect</groupId>
+        <artifactId>moditect-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>add-module-infos</id>
+            <phase>package</phase>
+            <goals>
+              <goal>add-module-info</goal>
+            </goals>
+            <configuration>
+              <overwriteExistingFiles>true</overwriteExistingFiles>
+              <module>
+                <moduleInfoSource><![CDATA[
+                  module org.neo4j.driver {
+                    exports org.neo4j.driver;
+                    exports org.neo4j.driver.async;
+                    exports org.neo4j.driver.reactive;
+                    exports org.neo4j.driver.reactivestreams;
+                    exports org.neo4j.driver.types;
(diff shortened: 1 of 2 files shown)


bundle/pom.xml
```diff
@@ -218,6 +218,43 @@
...
+                    exports org.neo4j.driver.reactivestreams;
+                    exports org.neo4j.driver.types;
+                    exports org.neo4j.driver.summary;
+                    exports org.neo4j.driver.net;
+                    exports org.neo4j.driver.util;
+                    exports org.neo4j.driver.exceptions;
+
+                    requires transitive java.logging;
+                    requires transitive org.reactivestreams;
+                    requires static micrometer.core;
```
injectives, line 248:
```suggestion
                    requires static org.graalvm.sdk;
```

bundle/pom.xml
```diff
@@ -218,6 +218,43 @@
...
+                    requires static org.graalvm.nativeimage.builder;
```
injectives:
```suggestion
                    requires static org.graalvm.sdk;
```

bundle/pom.xml
```diff
@@ -218,6 +218,43 @@
...
+                    requires static org.slf4j;
```
injectives, line 250:
```suggestion
                    requires static org.slf4j;
                    requires static java.management;
```

bundle/pom.xml
injectives: Just to be in sync with https://github.com/neo4j/neo4j-java-driver/blob/5.0/driver/src/main/java/module-info.java

bundle/pom.xml
injectives: Just to be in sync with https://github.com/neo4j/neo4j-java-driver/blob/5.0/driver/src/main/java/module-info.java

injectives (approved):

gjmwoods (approved):

PR #1359, opened 2022-12-20
Extract test-jar content to proper module

fbiville:


fbiville:
This is not ready yet:

- license checks are not configured 
- Javadoc for this module should probably be disabled

The end goal is to have `mvn compile` (or `mvn package`) work out of the box without resorting to `mvn install`.
This is currently not possible since the example module depends on the test sources of the driver module and test sources are only compiled after main sources are compiled (so the main sources of the example module will be compiled before the test sources of the driver module can be compiled/packaged).

PR #1361, opened 2023-01-04
Add new java platform logger

Sineaggi:
With the [java platform logger](https://openjdk.org/jeps/264), we can make the slf4j and jul loggers optional. Users can use implementations of the java platform logging api such as the ones provided by slf4j https://www.slf4j.org/apidocs/org/slf4j/jdk/platform/logging/package-summary.html and log4j https://logging.apache.org/log4j/2.x/log4j-jpl/project-info.html

### DIFF
--- a/driver/src/main/java/module-info.java
+++ b/driver/src/main/java/module-info.java
@@ -35,7 +35,7 @@ module org.neo4j.driver {
     requires io.netty.buffer;
     requires io.netty.codec;
     requires io.netty.resolver;
-    requires static java.logging;
+    requires static transitive java.logging;
     requires transitive org.reactivestreams;
     requires static micrometer.core;
     requires static org.graalvm.nativeimage.builder;
--- a/driver/src/main/java/module-info.java
+++ b/driver/src/main/java/module-info.java
@@ -35,7 +35,7 @@ module org.neo4j.driver {
     requires io.netty.buffer;
     requires io.netty.codec;
     requires io.netty.resolver;
-    requires transitive java.logging;
+    requires static java.logging;
     requires transitive org.reactivestreams;
     requires static micrometer.core;
     requires static org.graalvm.nativeimage.builder;
--- a/driver/src/main/java/org/neo4j/driver/Logging.java
+++ b/driver/src/main/java/org/neo4j/driver/Logging.java
@@ -24,6 +24,7 @@ import java.util.logging.ConsoleHandler;
 import java.util.logging.Level;
 import org.neo4j.driver.internal.logging.ConsoleLogging;
 import org.neo4j.driver.internal.logging.JULogging;
+import org.neo4j.driver.internal.logging.JavaPlatformLogging;
 import org.neo4j.driver.internal.logging.Slf4jLogging;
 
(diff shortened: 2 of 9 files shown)


michael-simons:
Nice work!

driver/src/main/java/org/neo4j/driver/Logging.java
```diff
@@ -118,13 +122,27 @@ static Logging slf4j() {
...
+    static Logging javaPlatformLogging() {
+        return new JavaPlatformLogging();
+    }
+
     /**
      * Create logging implementation that uses {@link java.util.logging}.
      *
      * @param level the log level.
      * @return new logging implementation.
+     * @throws IllegalStateException if SLF4J is not available.
```
injectives: Shouldn't `SLF4J` be replaced with something like `java.util.logging`?

injectives:
@Sineaggi, thanks for the contribution. 👍 

Would you be happy to sign the CLA please? More info here: https://github.com/neo4j/neo4j-java-driver/blob/1.6/CONTRIBUTING.md#want-to-contribute

This is why builds are failing at the moment.

Sineaggi:
> Would you be happy to sign the CLA please? More info here: https://github.com/neo4j/neo4j-java-driver/blob/1.6/CONTRIBUTING.md#want-to-contribute

I have signed the CLA. Please let me know if there are any issues.

driver/src/main/java/org/neo4j/driver/Logging.java
Sineaggi: Ah, I missed that. Fixed!

injectives:
> > Would you be happy to sign the CLA please? More info here: https://github.com/neo4j/neo4j-java-driver/blob/1.6/CONTRIBUTING.md#want-to-contribute
> 
> I have signed the CLA. Please let me know if there are any issues.

Thanks! The `whitelist-check` works now.

driver/src/main/java/module-info.java
```diff
@@ -35,7 +35,7 @@
     requires io.netty.buffer;
     requires io.netty.codec;
     requires io.netty.resolver;
-    requires transitive java.logging;
+    requires static java.logging;
```
injectives: This change triggers compilation failure, caused by the following warning: `class java.util.logging.Level in module java.logging is not indirectly exported using requires transitive`.

Sineaggi: Wow interesting. I tried creating a reproducer locally but wasn't able to.

According to what I've seen online we can make it `static` and `transitive`, but I'm not 100% sure. In this case it's true, the logging class is exporting JUL to downstream users (using the logger in the public return types).

Reading through https://nipafx.dev/java-modules-implied-readability/ right now.

Sineaggi: Is it still failing for the same reasons?

injectives:
@Sineaggi, sorry for the delay, but just wanted to let you know that the next major driver version will use the System Logger by default. See the following for more details: https://github.com/neo4j/neo4j-java-driver/pull/1648

PR #1369, opened 2023-01-27
Update QueryTask Javadoc

injectives:


AndyHeap-NeoTech (approved):
Good idea to keep the simple interface explanation.... simple.

PR #1376, opened 2023-02-14
Unmark TypeSystem as experimental

injectives:
The `TypeSystem` has been available for a significant time. This update removes the experimental status from it.


michael-simons (approved):
Excellent, also the deprecations 👍

PR #1377, opened 2023-02-15
Align driver with latest revision of `ExecuteQuery` ADR

fbiville:
Rename `QueryTask` occurrences with `ExecuteQuery`

Rename default `ExecuteQuery`'s default `BookmarkManager` getter

Make default `ExecuteQuery`'s default `BookmarkManager` unchangeable


driver/src/main/java/org/neo4j/driver/ExecuteQueryConfig.java
```diff
@@ -27,15 +27,15 @@
 import org.neo4j.driver.util.Experimental;
 
 /**
- * Query configuration used by {@link Driver#queryTask(String)} and its variants.
+ * Query configuration used by {@link Driver#executeQueryTemplate(String)} and its variants.
  * @since 5.5
  */
 @Experimental
-public final class QueryConfig implements Serializable {
+public final class ExecuteQueryConfig implements Serializable {
```
injectives: Isn't this one named `QueryConfig` in the ADR?

driver/clirr-ignored-differences.xml
```diff
@@ -412,19 +412,13 @@
     <difference>
         <className>org/neo4j/driver/Driver</className>
         <differenceType>7012</differenceType>
-        <method>org.neo4j.driver.QueryTask queryTask(java.lang.String)</method>
```
injectives: This name has previously been agreed with @AndyHeap-NeoTech. I will need to catch-up with him.

driver/src/main/java/org/neo4j/driver/ExecuteQueryConfig.java
fbiville: oops, yes, went a bit overboard, will revert that change

driver/clirr-ignored-differences.xml
fbiville: maybe, but that should be reflected in the ADR.
Cannot we get a more aligned name though? It does not change the overall approach chosen by the Java driver.

injectives (changes requested):
I had a catch-up with @AndyHeap-NeoTech today, who is working on the ADR amendment.
Following that, `QueryTask` and `QueryConfig` should remain as the are.
`BookmarkManager` accessor naming should include `default` word, but I think it will remain configurable upon driver creation (ADR update pending).

driver/src/main/java/org/neo4j/driver/Driver.java
```diff
@@ -64,24 +64,24 @@
...
     /**
-     * Returns an instance of {@link BookmarkManager} used by {@link QueryTask} instances by default.
+     * Returns an instance of {@link BookmarkManager} used by {@link ExecutableQuery} instances by default.
      *
      * @return bookmark manager, must not be {@code null}
      * @since 5.6
      */
     @Experimental
-    BookmarkManager queryTaskBookmarkManager();
+    BookmarkManager defaultExecuteQueryBookmarkManager();
```
michael-simons: Do we really want that on the drivers level? 
Simple API should be simple. There's little value here for the users we are aiming for. If we don't need it, we should not introduce it. 
If it is requested, we can still add.

michael-simons (changes requested):

driver/src/main/java/org/neo4j/driver/Driver.java
injectives: (To bring context) As far as I remember, this was introduced to allow access to default  `BookmarkManager` to make it possible to use the same instance in `SessionConfig`, where it must be set explicitly to opt-in. If we still want this to be possible, there should be some way of doing it. I am happy to explore other implementation options.

AndyHeap-NeoTech: The thinking behind having this method was to allow the user to causally chain with uses of the session API, where a bookmark manager can be supplied on session creation.

AndyHeap-NeoTech: Should have refreshed the page...Dmitriy beat me to it.

AndyHeap-NeoTech (approved):
All good.

AndyHeap-NeoTech (approved):
LGTM

PR #1382, opened 2023-02-22
Improve Javadoc

injectives:
This update also enables failures on warnings for public Javadoc.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/Config.java
+++ b/driver/src/main/java/org/neo4j/driver/Config.java
@@ -75,6 +75,9 @@ public final class Config implements Serializable {
 
     private static final Config EMPTY = builder().build();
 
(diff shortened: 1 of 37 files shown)


driver/src/main/java/org/neo4j/driver/Config.java
```diff
@@ -75,32 +75,77 @@ public final class Config implements Serializable {
 
     private static final Config EMPTY = builder().build();
 
+    /**
+     * The {@link QueryTask} {@link BookmarkManager}.
```
AndyHeap-NeoTech, line 79: Should this now be ExecuteQuery?

gjmwoods (approved):

AndyHeap-NeoTech (approved):

PR #1389, opened 2023-03-06
Seal AuthToken interface

injectives:
Driver effectively does not expect external implementations of `AuthToken` interface. This update makes it more explicit.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/AuthToken.java
+++ b/driver/src/main/java/org/neo4j/driver/AuthToken.java
@@ -18,6 +18,8 @@
  */
 package org.neo4j.driver;
 
+import org.neo4j.driver.internal.security.InternalAuthToken;
+
 /**
  * Token for holding authentication details, such as <em>user name</em> and <em>password</em>.
  * Such a token is required by a {@link Driver} to authenticate with a Neo4j
@@ -27,4 +29,4 @@ package org.neo4j.driver;
  * @see GraphDatabase#driver(String, AuthToken)
  * @since 1.0
  */
-public interface AuthToken {}
+public sealed interface AuthToken permits InternalAuthToken {}
--- a/driver/src/main/java/org/neo4j/driver/internal/async/connection/ChannelConnectorImpl.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/async/connection/ChannelConnectorImpl.java
@@ -31,15 +31,12 @@ import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.time.Clock;
 import org.neo4j.driver.AuthToken;
-import org.neo4j.driver.AuthTokens;
 import org.neo4j.driver.Logging;
-import org.neo4j.driver.exceptions.ClientException;
 import org.neo4j.driver.internal.BoltServerAddress;
 import org.neo4j.driver.internal.ConnectionSettings;
 import org.neo4j.driver.internal.DomainNameResolver;
 import org.neo4j.driver.internal.async.inbound.ConnectTimeoutHandler;
 import org.neo4j.driver.internal.cluster.RoutingContext;
-import org.neo4j.driver.internal.security.InternalAuthToken;
 import org.neo4j.driver.internal.security.SecurityPlan;
 
 public class ChannelConnectorImpl implements ChannelConnector {
@@ -80,7 +77,7 @@ public class ChannelConnectorImpl implements ChannelConnector {
             RoutingContext routingContext,
             DomainNameResolver domainNameResolver) {
         this.userAgent = connectionSettings.userAgent();
-        this.authToken = requireValidAuthToken(connectionSettings.authToken());
+        this.authToken = connectionSettings.authToken();
         this.routingContext = routingContext;
         this.connectTimeoutMillis = connectionSettings.connectTimeoutMillis();
         this.securityPlan = requireNonNull(securityPlan);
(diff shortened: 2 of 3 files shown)


michael-simons (approved):

michael-simons:
You would need to add a warning to the release notes, though. It might be that people are mocking that class with Mockito and that will stop working, too @injectives

mftruso:
AWS provides an example extending `InternalAuthToken` and overriding the `toMap()` method. This is necessary to generate a new [signature](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html) on every request. How would you do this now that `InternalAuthToken` is `final`? https://docs.aws.amazon.com/neptune/latest/userguide/access-graph-opencypher-bolt.html#access-graph-opencypher-bolt-java-iam-auth

```
package software.amazon.neptune.bolt;

import com.amazonaws.DefaultRequest;
import com.amazonaws.Request;
import com.amazonaws.auth.AWS4Signer;
import com.amazonaws.auth.AWSCredentialsProvider;
import com.amazonaws.http.HttpMethodName;
import com.google.gson.Gson;
import lombok.Builder;
import lombok.Getter;
import lombok.NonNull;
import org.neo4j.driver.Value;
import org.neo4j.driver.Values;
import org.neo4j.driver.internal.security.InternalAuthToken;
import org.neo4j.driver.internal.value.StringValue;

import java.net.URI;
import java.util.Collections;
import java.util.HashMap;
import java.util.Map;

import static com.amazonaws.auth.internal.SignerConstants.AUTHORIZATION;
import static com.amazonaws.auth.internal.SignerConstants.HOST;
import static com.amazonaws.auth.internal.SignerConstants.X_AMZ_DATE;
import static com.amazonaws.auth.internal.SignerConstants.X_AMZ_SECURITY_TOKEN;

/**
 * Use this class instead of `AuthTokens.basic` when working with an IAM
 * auth-enabled server. It works the same as `AuthTokens.basic` when using
 * static credentials, and avoids making requests with an expired signature
 * when using temporary credentials. Internally, it generates a new signature
 * on every invocation (this may change in a future implementation).
 *
 * Note that authentication happens only the first time for a pooled connection.
 *
 * Typical usage:
 *
 * NeptuneAuthToken authToken = NeptuneAuthToken.builder()
 *     .credentialsProvider(credentialsProvider)
 *     .region("aws region")
 *     .url("cluster endpoint url")
 *     .build();
 *
 * Driver driver = GraphDatabase.driver(
 *     authToken.getUrl(),
 *     authToken,
 *     config
 * );
 */

public class NeptuneAuthToken extends InternalAuthToken {
  private static final String SCHEME = "basic";
  private static final String REALM = "realm";
  private static final String SERVICE_NAME = "neptune-db";
  private static final String HTTP_METHOD_HDR = "HttpMethod";
  private static final String DUMMY_USERNAME = "username";
  @NonNull
  private final String region;
  @NonNull
  @Getter
  private final String url;
  @NonNull
  private final AWSCredentialsProvider credentialsProvider;
  private final Gson gson = new Gson();

  @Builder
  private NeptuneAuthToken(
      @NonNull final String region,
      @NonNull final String url,
      @NonNull final AWSCredentialsProvider credentialsProvider
  ) {
      // The superclass caches the result of toMap(), which we don't want
      super(Collections.emptyMap());
      this.region = region;
      this.url = url;
      this.credentialsProvider = credentialsProvider;
  }

  @Override
  public Map<String, Value> toMap() {
    final Map<String, Value> map = new HashMap<>();
    map.put(SCHEME_KEY, Values.value(SCHEME));
    map.put(PRINCIPAL_KEY, Values.value(DUMMY_USERNAME));
    map.put(CREDENTIALS_KEY, new StringValue(getSignedHeader()));
    map.put(REALM_KEY, Values.value(REALM));

    return map;
  }

  private String getSignedHeader() {
    final Request<Void> request = new DefaultRequest<>(SERVICE_NAME);
    request.setHttpMethod(HttpMethodName.GET);
    request.setEndpoint(URI.create(url));
    // Comment out the following line if you're using an engine version older than 1.2.0.0
    request.setResourcePath("/opencypher");

    final AWS4Signer signer = new AWS4Signer();
    signer.setRegionName(region);
    signer.setServiceName(request.getServiceName());
    signer.sign(request, credentialsProvider.getCredentials());

    return getAuthInfoJson(request);
  }

  private String getAuthInfoJson(final Request<Void> request) {
    final Map<String, Object> obj = new HashMap<>();
    obj.put(AUTHORIZATION, request.getHeaders().get(AUTHORIZATION));
    obj.put(HTTP_METHOD_HDR, request.getHttpMethod());
    obj.put(X_AMZ_DATE, request.getHeaders().get(X_AMZ_DATE));
    obj.put(HOST, request.getHeaders().get(HOST));
    obj.put(X_AMZ_SECURITY_TOKEN, request.getHeaders().get(X_AMZ_SECURITY_TOKEN));

    return gson.toJson(obj);
  }
}
```

mftruso:
my solution without extending `InternalAuthToken`

```
Driver driver = GraphDatabase.driver(
					new URI("neptune-cluster-uri"),
					authTokenManager,
					Config.defaultConfig()
			 );


public class NeptuneAuthTokenManager implements AuthTokenManager {
    @Override
    public CompletionStage<AuthToken> getToken() {
        AuthToken token = AuthTokens.custom(DUMMY_USERNAME, getSignedHeader(), REALM, SCHEME);
        return CompletableFuture.completedFuture(token);
    }

    @Override
    public boolean handleSecurityException(AuthToken authToken, SecurityException exception) {
        return false;
    }

...
}
```

PR #1390, opened 2023-03-08
Restore Session.reset as a private API

loveleif:
Restores `Session.reset()` as a private API. Intended for use in Cypher Shell to perform query cancellation (user press ctrl+c), which has been broken since reset was removed. We considered using 4.4 driver in 5.x as a work around, but it was decided to not merge because there were some unwanted effects (see https://github.com/neo-technology/neo4j/pull/18409).

Reverts  33447f573986c6117fa98ec7ccc85b787830cd15 with the following modifications:

- `reset()` is now not available in the public API.
- Adapt tests to new APIs.
- Source code for `longRunningStatement.jar` is now included in the project. Note, this module is not included as part of the parent build, it can be run locally to generate a new jar file in case of breaking procedure API changes.

https://trello.com/c/pV2nRf9x/359-mechanism-for-query-cancellation


injectives:
https://github.com/neo4j/neo4j-java-driver/pull/1394

PR #1396, opened 2023-03-24
Introduce notification configuration, severity and category

injectives:
# Objective

Improve and expand the existing notification API.

## NotificationSeverity

A new `NotificationSeverity` type represents a notification severity level.

It includes 2 public constants:
- `INFORMATION`
- `WARNING`

## NotificationCategory

A new `NotificationCategory` type represents a notification category.

It includes 6 public constants:
- `HINT`
- `UNRECOGNIZED`
- `UNSUPPORTED`
- `PERFORMANCE`
- `DEPRECATION`
- `GENERIC`

## Notification

The `Notification` class has been extended with the following 4 methods:
- `Optional<NotificationSeverity> severityLevel()`
- `Optional<String> rawSeverityLevel()`
- `Optional<NotificationCategory> category()`
- `Optional<String> rawCategory()`

The `raw` methods return a `String` representation returned by the server. In case of an unrecognised value, both `severityLevel()` and `category()` methods return an empty `Optional`. This may happen if a new value is introduced in a future server version and a previous driver version does not support it. Additionally, an empty `Optional` may be returned when driver communicates with a server version that does not supply these values.

The `severity()` method has been deprecated in favour of the `rawSeverityLevel()` method.

## NotificationConfig

A new `NotificationConfig` type has been introduced to allow notification preferences management. By default, the server determines what notifications are provided to the driver. However, user can set a minimum severity level and/or a set of disabled notification categories to manage its expectations. This feature is only supported on Bolt protocol version 5.2 and above.

Both the `Config` and the `SessionConfig` support this new configuration.

Sample usage:
```java
// sets minimum notification severity level to WARNING
// and explicitly disables both GENERIC and HINT notification categories
var driverConfig = Config.builder()
        .withNotificationConfig(NotificationConfig.defaultConfig().enableMinimumSeverity(NotificationSeverity.WARNING).disableCategories(Set.of(NotificationCategory.GENERIC, NotificationCategory.HINT)))
        .build();
// disables all notifications
var sessionConfig = SessionConfig.builder()
        .withNotificationConfig(NotificationConfig.disableAll())
        .build();
```

## Bolt

This update includes support for both 5.1 and 5.2 versions. The latter is required for the full support of the new notification updates.

### DIFF
--- a/driver/clirr-ignored-differences.xml
+++ b/driver/clirr-ignored-differences.xml
@@ -433,4 +433,28 @@
         <method>org.neo4j.driver.BookmarkManager queryTaskBookmarkManager()</method>
     </difference>
 
+    <difference>
+        <className>org/neo4j/driver/summary/Notification</className>
+        <differenceType>7012</differenceType>
+        <method>java.util.Optional severityLevel()</method>
+    </difference>
+
+    <difference>
+        <className>org/neo4j/driver/summary/Notification</className>
+        <differenceType>7012</differenceType>
+        <method>java.util.Optional rawSeverityLevel()</method>
+    </difference>
+
+    <difference>
+        <className>org/neo4j/driver/summary/Notification</className>
+        <differenceType>7012</differenceType>
+        <method>java.util.Optional category()</method>
+    </difference>
+
+    <difference>
+        <className>org/neo4j/driver/summary/Notification</className>
+        <differenceType>7012</differenceType>
+        <method>java.util.Optional rawCategory()</method>
+    </difference>
+
 </differences>
--- a/driver/src/main/java/org/neo4j/driver/internal/messaging/v3/BoltProtocolV3.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/messaging/v3/BoltProtocolV3.java
@@ -34,6 +34,7 @@ import java.util.concurrent.CompletionStage;
 import java.util.function.Consumer;
 import org.neo4j.driver.AuthToken;
 import org.neo4j.driver.Bookmark;
+import org.neo4j.driver.NotificationConfig;
 import org.neo4j.driver.Query;
 import org.neo4j.driver.TransactionConfig;
 import org.neo4j.driver.internal.DatabaseBookmark;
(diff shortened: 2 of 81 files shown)


driver/clirr-ignored-differences.xml
```diff
@@ -433,4 +433,28 @@
...
+        <className>org/neo4j/driver/summary/Notification</className>
+        <differenceType>7012</differenceType>
+        <method>java.util.Optional category()</method>
+    </difference>
+
+    <difference>
+        <className>org/neo4j/driver/summary/Notification</className>
+        <differenceType>7012</differenceType>
+        <method>java.util.Optional rawCategory()</method>
+    </difference>
```
bigmontz, line 458: How does it works?

driver/src/main/java/org/neo4j/driver/internal/messaging/v3/BoltProtocolV3.java
```diff
@@ -189,4 +213,8 @@ public BoltProtocolVersion version() {
     protected boolean includeDateTimeUtcPatchInHello() {
         return false;
     }
+
+    protected boolean includeNotificationSettings() {
+        return false;
+    }
```
bigmontz: The notification information should not be swallowed by the driver. If the protocol doesn't support send notification filters and error should be thrown.

driver/clirr-ignored-differences.xml
injectives: https://www.mojohaus.org/clirr-maven-plugin/usage.html#Compatibility_Checks

> The clirr:check goal is used to check for compatibility with previous versions.

In this case it checks against version `5.6.0` and found that 4 new methods have been added to the `Notification` interface, which could break implementations of this interface. In this case, however, it is suggested to accept this change as we don't expect nor support external implementations of the `Notification` interface. We will probably make it `sealed` in upcoming major release.

driver/src/main/java/org/neo4j/driver/internal/messaging/v3/BoltProtocolV3.java
injectives: Updated as per the discussion.

bigmontz (approved):
🔔

PR #1399, opened 2023-03-29
Remove whitelisted Testkit server configs

injectives:
Testkit no longer uses 3.x servers and these configs are not necessary.


AndyHeap-NeoTech (approved):
LGTM

PR #1404, opened 2023-04-05
Include NotificationConfig on BEGIN message for bolt scheme

injectives:
`NotificationConfig` might be ignored on new sessions with `bolt` scheme. This update fixes it.

### DIFF
--- a/bundle/pom.xml
+++ b/bundle/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.neo4j.driver</groupId>
     <artifactId>neo4j-java-driver-parent</artifactId>
-    <version>5.8-SNAPSHOT</version>
+    <version>5.7-SNAPSHOT</version>
     <relativePath>..</relativePath>
   </parent>
 
--- a/driver/pom.xml
+++ b/driver/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.neo4j.driver</groupId>
     <artifactId>neo4j-java-driver-parent</artifactId>
-    <version>5.8-SNAPSHOT</version>
+    <version>5.7-SNAPSHOT</version>
   </parent>
 
   <artifactId>neo4j-java-driver</artifactId>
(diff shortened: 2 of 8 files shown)


bundle/pom.xml
```diff
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.neo4j.driver</groupId>
     <artifactId>neo4j-java-driver-parent</artifactId>
-    <version>5.8-SNAPSHOT</version>
+    <version>5.7-SNAPSHOT</version>
```
AndyHeap-NeoTech, line 9: Is this still valid if we are releasing in 5.8?

driver/pom.xml
```diff
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.neo4j.driver</groupId>
     <artifactId>neo4j-java-driver-parent</artifactId>
-    <version>5.8-SNAPSHOT</version>
+    <version>5.7-SNAPSHOT</version>
```
AndyHeap-NeoTech, line 9: Is this still valid if we are releasing in 5.8?

examples/pom.xml
```diff
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.neo4j.driver</groupId>
     <artifactId>neo4j-java-driver-parent</artifactId>
-    <version>5.8-SNAPSHOT</version>
+    <version>5.7-SNAPSHOT</version>
```
AndyHeap-NeoTech, line 9: Is this still valid if we are releasing in 5.8?

AndyHeap-NeoTech (approved):
Just one question, is the change to the 5.7 snapshot still valid if the fix is going to go out in 5.8?

bundle/pom.xml
injectives: It is fine, we will have a chance to update it later.

injectives:
> Just one question, is the change to the 5.7 snapshot still valid if the fix is going to go out in 5.8?

https://github.com/neo4j/neo4j-java-driver/pull/1404#discussion_r1159509368

PR #1407, opened 2023-04-06
Make Driver Level Queries API GA

injectives:
This update removes the Preview status from the Driver Level Queries API, making it GA.

The `RoutingControl` type has been changed to a sealed interface with public constants. The names have been updated to `WRITE` and `READ`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,10 +78,9 @@ The `neo4j-java-driver-all` includes an explicit module declaration ([module-inf
 To run a simple query, the following can be used:
 ```java
 var authToken = AuthTokens.basic("neo4j", "password");
-try (var driver = GraphDatabase.driver("bolt://localhost:7687", authToken); var session = driver.session()) {
-    var result = session.run("CREATE (n)");
-    var summary = result.consume();
-    System.out.println(summary.counters().nodesCreated());
+try (var driver = GraphDatabase.driver("bolt://localhost:7687", authToken)) {
+    var result = driver.executableQuery("CREATE (n)").execute();
+    System.out.println(result.summary().counters().nodesCreated());
 }
 ```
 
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -78,10 +78,9 @@ The `neo4j-java-driver-all` includes an explicit module declaration ([module-inf
 To run a simple query, the following can be used:
 ```java
 var authToken = AuthTokens.basic("neo4j", "password");
-try (var driver = GraphDatabase.driver("bolt://localhost:7687", authToken); var session = driver.session()) {
-    var result = session.run("CREATE (n)");
-    var summary = result.consume();
-    System.out.println(summary.counters().nodesCreated());
+try (var driver = GraphDatabase.driver("bolt://localhost:7687", authToken)) {
+    var result = driver.executableQuery("CREATE (n)").execute();
```
AndyHeap-NeoTech, line 82: This makes me happy to see. No use of session.run as a first step :)

AndyHeap-NeoTech (approved):
Looks good. Nice to see the driver level API coming out of preview.

README.md
injectives: Yeah, I thought I might as well update the `README`.

injectives:
For general information, it has been decided to change the type of the `RoutingControl` from `enum` to a `sealed interface` with `public` constants. The main reason for this is that the list of values is not considered final at this point. There is already a discussion on a possible new type that might come within 5.7 series. As a side note, it would not be recommended to write switch expressions on such type.

injectives:
For general information, it has been decided to change the type of the `RoutingControl` from `enum` to a `sealed interface` with `public` constants. The main reason for this is that the list of values is not considered final at this point. There is already a discussion on a possible new type that might come within 5.x series. As a side note, it would not be recommended to write switch expressions on such type.

injectives:
For general information, it has been decided to change the type of the RoutingControl from enum to a sealed interface with public constants. The main reason for this is that the list of values is not considered final at this point. There is already a discussion on a possible new type that might come within 5.x series. As a side note, it would not be recommended to write switch expressions on such type.

PR #1408, opened 2023-04-11
Make Bookmark Manager API GA

injectives:
This update removes the Preview status from the Bookmark Manager API, making it GA.


AndyHeap-NeoTech (approved):
LGTM

PR #1424, opened 2023-05-15
Introduce `bolt_agent`

injectives:
This update introduces support for the new `HELLO` message `bolt_agent` field that is mandatory in the Bolt protocol version 5.3.

The `bolt_agent` value is non-configurable and follows a common format among the official drivers.

Java driver `bolt_agent` sample:
```json
{"product": "neo4j-java/5.7.0-bf6c5444e19cef8b6cb7aa7e0cecf80b7f8a19a9", "platform": "Linux; 5.15.49-linuxkit; aarch64", "language": "Java/17.0.6", "language_details": "Eclipse Adoptium; OpenJDK 64-Bit Server VM; 17.0.6+10"}
```

This update also brings support for the Bolt protocol version 5.3.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/Config.java
+++ b/driver/src/main/java/org/neo4j/driver/Config.java
@@ -18,8 +18,8 @@
  */
 package org.neo4j.driver;
 
-import static java.lang.String.format;
 import static org.neo4j.driver.internal.logging.DevNullLogging.DEV_NULL_LOGGING;
+import static org.neo4j.driver.internal.util.DriverInfoUtil.driverVersion;
 
 import java.io.File;
 import java.io.Serial;
@@ -345,7 +345,7 @@ public final class Config implements Serializable {
         private long idleTimeBeforeConnectionTest = PoolSettings.DEFAULT_IDLE_TIME_BEFORE_CONNECTION_TEST;
         private long maxConnectionLifetimeMillis = PoolSettings.DEFAULT_MAX_CONNECTION_LIFETIME;
         private long connectionAcquisitionTimeoutMillis = PoolSettings.DEFAULT_CONNECTION_ACQUISITION_TIMEOUT;
-        private String userAgent = format("neo4j-java/%s", driverVersion());
+        private String userAgent = driverVersion();
         private final SecuritySettings.SecuritySettingsBuilder securitySettingsBuilder =
                 new SecuritySettings.SecuritySettingsBuilder();
         private long routingTablePurgeDelayMillis = RoutingSettings.STALE_ROUTING_TABLE_PURGE_DELAY_MS;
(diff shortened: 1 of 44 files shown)


driver/src/main/java/org/neo4j/driver/Config.java
```diff
@@ -345,7 +345,7 @@ public static final class ConfigBuilder {
         private long idleTimeBeforeConnectionTest = PoolSettings.DEFAULT_IDLE_TIME_BEFORE_CONNECTION_TEST;
         private long maxConnectionLifetimeMillis = PoolSettings.DEFAULT_MAX_CONNECTION_LIFETIME;
         private long connectionAcquisitionTimeoutMillis = PoolSettings.DEFAULT_CONNECTION_ACQUISITION_TIMEOUT;
-        private String userAgent = format("neo4j-java/%s", driverVersion());
+        private String userAgent = driverVersion();
```
bigmontz: Shouldn't it keep as `format("neo4j-java/%s", driverVersion())`? afaik, the userAgent should not changed by the `boltAgent` introduction since the user can have integrations depending on this log.

driver/src/test/java/org/neo4j/driver/internal/messaging/request/HelloMessageTest.java
```diff
@@ -39,10 +40,12 @@ void shouldHaveCorrectMetadata() {
         authToken.put("user", value("Alice"));
         authToken.put("credentials", value("SecretPassword"));
 
-        HelloMessage message = new HelloMessage("MyDriver/1.0.2", authToken, Collections.emptyMap(), false, null);
+        HelloMessage message =
+                new HelloMessage("MyDriver/1.0.2", BoltAgentUtil.VALUE, authToken, Collections.emptyMap(), false, null);
```
bigmontz, line 45: Test with the `BoltAgent` equals to null or with `BoltAgent` with the complete information can be helpful to avoid future regretions.

driver/src/main/java/org/neo4j/driver/internal/util/DriverInfoUtil.java
```diff
@@ -0,0 +1,88 @@
...
+            getProperty("java.vm.version").ifPresent(value -> append(value, languageDetailsBuilder));
+            return languageDetailsBuilder.isEmpty() ? null : languageDetailsBuilder;
+        });
+
+        return new BoltAgent(
+                productInfo,
+                platformBuilder.isEmpty() ? null : platformBuilder.toString(),
+                language.orElse(null),
+                languageDetails.isEmpty() ? null : languageDetails.toString());
+    }
```
bigmontz, line 51: Might worth to add some unit tests for checking if the formats are the ones expected.

bigmontz (changes requested):
The change in the `userAgent` is not supposed to happen in this change.

driver/src/main/java/org/neo4j/driver/Config.java
injectives: Oh, I was not actually going to change that. It is a side-effect of reworking this PR. Thanks for noticing this. Fixed.

driver/src/test/java/org/neo4j/driver/internal/messaging/request/HelloMessageTest.java
injectives: Done.

driver/src/main/java/org/neo4j/driver/internal/util/DriverInfoUtil.java
injectives: Done. I have also added a test for the default user agent to the `ConfigTest`.

bigmontz (approved):
🕴🏼

PR #1425, opened 2023-05-22
Docs: mention performance impact of notification filtering.

robsdedude:


driver/src/main/java/org/neo4j/driver/NotificationConfig.java
```diff
@@ -34,7 +34,9 @@
  * </ul>
  * <p>
  * By default, both options are not activated.
- *
+ * <p>
+ * Disabling categories or severities allows the server to skip analysis for those, which can speed up query execution.
```
injectives, line 38: Please run `mvn spotless:apply` to correct formatting.

PR #1445, opened 2023-07-01
Improve explicit transaction terminated state handling

injectives:
An explicit transaction is considered terminated as soon as a failure occurs. Any usage of the terminated transaction and any of its results must be stopped and the transaction must be closed explicitly.

This update aims to ensure that the driver does not send further Bolt messages in regards to the terminated transaction.

In addition, a new `TransactionTerminatedException` class has been introduced. It is a subclass of the previously used `ClientException`, making this exception more specific. The exception will contain a non-null `code` if it is created based on the server's response. It will not have a code if it is generated by the driver. Depending on a failure cause, the result handles may emit other exceptions respectively, matching the driver's existing behaviour.


driver/src/main/java/org/neo4j/driver/internal/async/NetworkConnection.java
```diff
@@ -61,12 +70,13 @@ public class NetworkConnection implements Connection {
...
     private final Clock clock;
-
-    private final AtomicReference<Status> status = new AtomicReference<>(Status.OPEN);
     private final MetricsListener metricsListener;
     private final ListenerEvent<?> inUseEvent;
 
     private final Long connectionReadTimeout;
+
+    private Status status = Status.OPEN;
+    private UnmanagedTransaction transaction;
```
bigmontz: `UnmanagedTransaction` depends on `Connection` and `Connection` is depending on `UnmanagedTransaction`, this makes theses classes highly coupled. 

The only part of the transaction needed in the connection is the method `executeWithLockedState`. Maybe you can extract an interface called `LockedStateExecutor` used it instead of the `UnmanagedTransaction`. So, the `bindTransaction` method will also change to `bindLockedStateExecutor` or something simillar. 

This also make the Connection/NetworkConnection more testable, since the code won't depend on a concrete implementation.

driver/src/main/java/org/neo4j/driver/internal/async/NetworkConnection.java
```diff
@@ -194,6 +184,25 @@ public BoltProtocol protocol() {
         return protocol;
     }
 
+    @Override
+    public void bindTransaction(UnmanagedTransaction transaction) {
```
bigmontz: Might worth cover it with tests. The code is quite simple, but it's good avoid any future change break some desired behaviour by accident.

driver/src/main/java/org/neo4j/driver/internal/async/NetworkConnection.java
```diff
@@ -208,73 +217,49 @@ private void writeResetMessageIfNeeded(ResponseHandler resetHandler, boolean isS
...
     }
 
-    private void flushInEventLoop() {
-        channel.eventLoop().execute(() -> {
-            channel.flush();
-            registerConnectionReadTimeout(channel);
-        });
-    }
-
     private void writeMessageInEventLoop(Message message, ResponseHandler handler, boolean flush) {
```
bigmontz, line 219: I can't find the unit test code that tests this changes. It might worth to add some unit to `write` and `writeAndFlush` methods to cover cases where there is a `TerminatorAwareExecutor` and it's terminated or not. Also, when the message are in the list of out of the desired list, and so on.

bigmontz (approved):
I approved with some comments. They are not blockers.

driver/src/main/java/org/neo4j/driver/internal/async/NetworkConnection.java
injectives: Indeed. There is no need for full `UnmanagedTransaction` and it was not like that initially. During the investigation and on-going development the full transaction was eventually exposed. I have refactored that bit now. Thank!

driver/src/main/java/org/neo4j/driver/internal/async/NetworkConnection.java
injectives: A test has been added.

driver/src/main/java/org/neo4j/driver/internal/async/NetworkConnection.java
injectives: Added more tests.

PR #1446, opened 2023-07-06
Ensure transaction prevents new query runs when it is closing

injectives:
As a protection measure also prevent `COMMIT` and `ROLLBACK` Bolt messages from being dispatched if transaction is terminated.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/async/UnmanagedTransaction.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/async/UnmanagedTransaction.java
@@ -247,6 +247,10 @@ public class UnmanagedTransaction implements TerminationAwareStateLockingExecuto
                                     + "it has either experienced an fatal error or was explicitly terminated",
                             causeOfTermination);
                 }
+            } else if (commitFuture != null) {
+                throw new ClientException("Cannot run more queries in this transaction, it is being committed");
+            } else if (rollbackFuture != null) {
+                throw new ClientException("Cannot run more queries in this transaction, it is being rolled back");
             }
         });
     }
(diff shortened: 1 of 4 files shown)


driver/src/main/java/org/neo4j/driver/internal/async/UnmanagedTransaction.java
robsdedude, line 1: Is this something that can be tested with TestKit to see if other drivers can be manipulated into sending `COMMIT` or `ROLLBACK` on a failed transaction?

driver/src/main/java/org/neo4j/driver/internal/async/UnmanagedTransaction.java
```diff
@@ -247,6 +247,10 @@ private void ensureCanRunQueries() {
                                     + "it has either experienced an fatal error or was explicitly terminated",
                             causeOfTermination);
                 }
+            } else if (commitFuture != null) {
+                throw new ClientException("Cannot run more queries in this transaction, it is being committed");
+            } else if (rollbackFuture != null) {
+                throw new ClientException("Cannot run more queries in this transaction, it is being rolled back");
```
bigmontz, line 253: I'm just not that sure about the message, since the commit and rollback failed.

bigmontz (approved):
🍄

driver/src/main/java/org/neo4j/driver/internal/async/UnmanagedTransaction.java
bigmontz: I had misread `future` as `failure`.

driver/src/main/java/org/neo4j/driver/internal/async/UnmanagedTransaction.java
injectives: Not sure if we can hit these particular lines with Testkit. Technically, the Java driver is expected to fail sooner and this is just a safety and logical measure.

However, we can indeed test if drivers might send `COMMIT` or `ROLLBACK` after a failure via Testkit. I will take a look.

PR #1448, opened 2023-07-07
Update to using the local variable type inference in tests

injectives:


bigmontz (approved):
`var approved = "APPROVED"`

PR #1451, opened 2023-07-10
Update sub millisecond transaction timeout handling

injectives:
If transaction timeout value has a sub millisecond value, it gets rounded up to next millisecond value.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/messaging/request/TransactionMetadataBuilder.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/messaging/request/TransactionMetadataBuilder.java
@@ -20,13 +20,13 @@ package org.neo4j.driver.internal.messaging.request;
 
 import static java.util.Collections.emptyMap;
 import static org.neo4j.driver.Values.value;
-import static org.neo4j.driver.internal.DatabaseNameUtil.defaultDatabase;
 
 import java.time.Duration;
 import java.util.Map;
 import java.util.Set;
 import org.neo4j.driver.AccessMode;
 import org.neo4j.driver.Bookmark;
+import org.neo4j.driver.Logging;
 import org.neo4j.driver.NotificationConfig;
 import org.neo4j.driver.Value;
 import org.neo4j.driver.internal.DatabaseName;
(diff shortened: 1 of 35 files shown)


driver/src/main/java/org/neo4j/driver/internal/messaging/request/TransactionMetadataBuilder.java
```diff
@@ -87,7 +78,14 @@ public static Map<String, Value> buildMetadata(
             result.put(BOOKMARKS_METADATA_KEY, value(bookmarks.stream().map(Bookmark::value)));
         }
         if (txTimeoutPresent) {
-            result.put(TX_TIMEOUT_METADATA_KEY, value(txTimeout.toMillis()));
+            var millis = txTimeout.toMillis();
+            if (txTimeout.toNanosPart() % 1_000_000 > 0) {
```
thelonelyvulpes, line 82: Another option for this check is to muliply the millis and check against nanos, mod 1,000,000 is probably clearer though.
```suggestion
            if (txTimeout.toNanosPart() != millis * 1_000_000) {
```
This validation could be in some method `checkForSubmillisecondPrecision` too, just to document for yourself in the future.

thelonelyvulpes (approved):
Just some ideas, looks good.

driver/src/test/java/org/neo4j/driver/internal/messaging/request/TransactionMetadataBuilderTest.java
```diff
@@ -119,11 +127,44 @@ void shouldIncludeNotificationConfig() {
...
 
         var expectedMetadata = new HashMap<String, Value>();
         expectedMetadata.put("notifications_minimum_severity", value("WARNING"));
         expectedMetadata.put("notifications_disabled_categories", value(Set.of("UNSUPPORTED")));
         assertEquals(expectedMetadata, metadata);
     }
+
+    @ParameterizedTest
+    @ValueSource(longs = {1, 1_000_001, 100_500_000, 100_700_000, 1_000_000_001})
+    void shouldRoundUpFractionalTimeoutAndLog(long nanosValue) {
```
bigmontz, line 141: Might also test if the round up doesn't happen and nothing is logged when the value is not fractional

bigmontz (approved):
⏲️

driver/src/test/java/org/neo4j/driver/internal/messaging/request/TransactionMetadataBuilderTest.java
injectives: That bit is technically covered by the existing `shouldHaveCorrectMetadata` test. But I guess we can have a dedicated test as well.

PR #1454, opened 2023-07-11
Refactor

injectives:
This update includes general refactoring, like:
- converting to local variables
- removal of empty methods
- removal of redundant throws
- making eligible classes static
- removal of identical override methods
- upgrade to enhanced switch
- update to text block
- using pattern variables
- using records
- using collection factory calls


driver/src/main/java/org/neo4j/driver/internal/DriverFactory.java
```diff
@@ -161,15 +161,11 @@ protected static MetricsProvider getOrCreateMetricsProvider(Config config, Clock
...
-                return new InternalMetricsProvider(clock, config.logging());
-            case MICROMETER:
-                return MicrometerMetricsProvider.forGlobalRegistry();
-        }
-        throw new IllegalStateException("Unknown or unsupported MetricsAdapter: " + metricsAdapter);
+        return switch (metricsAdapter) {
+            case DEV_NULL -> DevNullMetricsProvider.INSTANCE;
+            case DEFAULT -> new InternalMetricsProvider(clock, config.logging());
+            case MICROMETER -> MicrometerMetricsProvider.forGlobalRegistry();
+        };
```
bigmontz, line 168: What happens if `metricAdapter` doesn't match any of the cases?

Should it have a default `throw new IllegalStateException("Unknown or unsupported MetricsAdapter: " + metricsAdapter);`?

examples/src/main/java/org/neo4j/docs/driver/ConfigCustomResolverExample.java
```diff
@@ -44,20 +44,19 @@ public ConfigCustomResolverExample(String virtualUri, AuthToken authToken, Serve
     }
 
     // tag::config-custom-resolver[]
-    private Driver createDriver(String virtualUri, String user, String password, ServerAddress... addresses) {
+    private Driver createDriver(String user, String password, ServerAddress... addresses) {
```
bigmontz: This is an examples, it might not worth to change since the url cal in fact change in real life.

examples/src/main/java/org/neo4j/docs/driver/PassBookmarkExample.java
```diff
@@ -57,9 +57,9 @@ private Result employ(TransactionContext tx, String person, String company) {
     }
 
     // Create a friendship between two people.
-    private Result makeFriends(TransactionContext tx, String person1, String person2) {
+    private Result makeFriends(TransactionContext tx) {
         return tx.run("MATCH (a:Person {name: $person_1}) MATCH (b:Person {name: $person_2}) MERGE (a)-[:KNOWS]->(b)",
-                parameters("person_1", person1, "person_2", person2));
+                parameters("person_1", "Alice", "person_2", "Bob"));
```
bigmontz: I think the example should be changed since the code should show a generic case where different people are friends

bigmontz (changes requested):
The examples should not change and the switch case in the DriverFactory seems wrong

examples/src/main/java/org/neo4j/docs/driver/ConfigCustomResolverExample.java
injectives: Good point, updated. 👍

examples/src/main/java/org/neo4j/docs/driver/PassBookmarkExample.java
injectives: Good point, updated. 👍

bigmontz (approved):
🧹

PR #1457, opened 2023-07-14
Add support for cancellation on reactive session run

injectives:
This update brings support for cancelling subscription on reactive session `run` methods, like:
- `org.neo4j.driver.reactive.ReactiveSession.run(..)`
- `org.neo4j.driver.reactivestreams.ReactiveSession.run(..)`

If subscription is cancelled before an instance of the `ReactiveResult` is published, the driver will rollback the initiated query by sending the Bolt `RESET` message and will release the network connection back to its connection pool. This will be done in the background.

Once the `ReactiveResult` is published, the cancellation with the rollback is no longer possible.

In addition, this update makes the driver wait until subscriber subscribes to the publisher before dispatching the query to the server. Previously, the driver would do that immediately upon calling the `run` method.


bigmontz (approved):
🚀

PR #1461, opened 2023-07-19
TestKit: add verifyConnectivity to feature list

robsdedude:
The driver already supports the API, so does the backend. The backend just never told TestKit about it ;)


robsdedude:
Closing for now. Needs further investigation and discussion.

PR #1485, opened 2023-09-05
Generalize connector implementation to use SocketAddress. This opens …

gjmwoods:
…the possibility to connect to other network implementations (ie Unix Sockets and Local Channels).

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/async/connection/ChannelConnectorImpl.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/async/connection/ChannelConnectorImpl.java
@@ -26,20 +26,22 @@ import io.netty.channel.ChannelOption;
 import io.netty.channel.ChannelPromise;
 import io.netty.resolver.AddressResolverGroup;
 import java.net.InetSocketAddress;
-import java.net.SocketAddress;
+import java.net.UnknownHostException;
 import java.time.Clock;
 import org.neo4j.driver.AuthTokenManager;
 import org.neo4j.driver.Logging;
 import org.neo4j.driver.NotificationConfig;
 import org.neo4j.driver.internal.BoltAgent;
-import org.neo4j.driver.internal.BoltServerAddress;
 import org.neo4j.driver.internal.ConnectionSettings;
 import org.neo4j.driver.internal.DomainNameResolver;
 import org.neo4j.driver.internal.async.inbound.ConnectTimeoutHandler;
 import org.neo4j.driver.internal.cluster.RoutingContext;
 import org.neo4j.driver.internal.security.SecurityPlan;
 
-public class ChannelConnectorImpl implements ChannelConnector {
+/**
+ * A {@link ChannelConnector} that establishes TCP connections.
+ */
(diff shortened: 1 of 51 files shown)


driver/src/main/java/org/neo4j/driver/internal/async/connection/ChannelConnectorImpl.java
```diff
@@ -99,43 +101,43 @@ public ChannelConnectorImpl(
...
-        SocketAddress socketAddress;
+        InetSocketAddress resolvedAddress;
+
         try {
-            socketAddress =
-                    new InetSocketAddress(domainNameResolver.resolve(address.connectionHost())[0], address.port());
-        } catch (Throwable t) {
-            socketAddress = InetSocketAddress.createUnresolved(address.connectionHost(), address.port());
+            resolvedAddress =
+                    new InetSocketAddress(domainNameResolver.resolve(address.getHostString())[0], address.getPort());
```
injectives, line 114: There might be an issue here. The `address.connectionHost()` may be different to the host specified in the original URL. 🤔 
> // Either is equal to the host or is explicitly provided on creation and is expected to be a resolved IP address.

driver/src/main/java/org/neo4j/driver/internal/async/pool/NettyChannelPool.java
```diff
@@ -236,7 +236,7 @@ public String id() {
         return this.id;
     }
 
-    private String poolId(BoltServerAddress serverAddress) {
-        return String.format("%s:%d-%d", serverAddress.host(), serverAddress.port(), this.hashCode());
+    private String poolId(SocketAddress serverAddress) {
+        return String.format("pool-%d-%d", serverAddress.hashCode(), this.hashCode());
```
injectives, line 240: Not sure if we should change this as it is exposed via: `ConnectionPoolMetrics#id()`.

driver/src/main/java/org/neo4j/driver/internal/metrics/InternalConnectionPoolMetrics.java
```diff
@@ -221,6 +224,9 @@ public String toString() {
     // This is used by the Testkit backend
     @SuppressWarnings("unused")
     public ServerAddress getAddress() {
-        return address;
+        if (address instanceof InetSocketAddress inetSocketAddress) {
+            return new BoltServerAddress(inetSocketAddress.getHostString(), inetSocketAddress.getPort());
+        }
+        return null;
```
injectives, line 230: Throwing might be preferable here. 🤔

driver/src/main/java/org/neo4j/driver/internal/async/connection/ChannelConnectorImpl.java
injectives: I've run the relevant Testkit tests and those failed:
- `test_should_request_rt_from_all_initial_routers_until_successful_on_unknown_failure`
- `test_should_request_rt_from_all_initial_routers_until_successful_on_authorization_expired`

This approach would need some solution in this area if we wanted to move forward with it.

PR #1487, opened 2023-09-26
Introduce Api Telemetry

bigmontz:
Neo4j would like to be able to track more information about driver apis usage, so that smarter decisions can be made on improving drivers and their stack.

The collection of metrics must be respectful of users. Thus, the collected metrics are:
1. impossible to be tied back the customer or user
1. transparent to the users that inspect
1. restrained to required metrics

The metric collection is disabled by default in Neo4j. It can be enabled in the server by setting `server.bolt.telemetry.enabled` to `true`.

However, the metric collection is enabled by default in the drivers. It can be disabled in the driver by configuring the driver with `telemetryDisabled` to `true`. Default: `false`.

**Metrics are only collected when enabled both in server and driver instances.**


injectives (commented):
Testkit failed because of flaky tests unrelated to this.

PR #1490, opened 2023-10-04
Add SECURITY and TOPOLOGY notification categories

injectives:


bigmontz (approved):
🔇 

(I think we need some tests for this, especially for checking if the new categories are not treated as unknown when deserialised)

injectives:
We currently rely on the Testkit, which has a pretty good coverage.

driver/src/main/java/org/neo4j/driver/NotificationCategory.java
```diff
@@ -64,6 +64,22 @@ public sealed interface NotificationCategory extends Serializable permits Intern
      */
     NotificationCategory DEPRECATION = new InternalNotificationCategory(Type.DEPRECATION);
 
+    /**
+     * A security category.
+     * <p>
+     * For instance, the security warnings.
+     * @since 5.14
```
robsdedude:
```suggestion
     * For instance, the security warnings.
     * <p>
     * Requires server version 5.14 or newer.
     * @since 5.14
```

driver/src/main/java/org/neo4j/driver/NotificationCategory.java
```diff
@@ -64,6 +64,22 @@ public sealed interface NotificationCategory extends Serializable permits Intern
...
+     */
+    NotificationCategory SECURITY = new InternalNotificationCategory(Type.SECURITY);
+
+    /**
+     * A topology category.
+     * <p>
+     * For instance, the topology notifications.
+     * @since 5.14
```
robsdedude:
```suggestion
     * For instance, the topology notifications.
     * <p>
     * Requires server version 5.14 or newer.
     * @since 5.14
```

robsdedude (approved):

injectives:
> (I think we need some tests for this, especially for checking if the new categories are not treated as unknown when deserialised)

New tests have been added.

PR #1495, opened 2023-10-24
Make AuthToken rotation and session auth support API GA

injectives:
This update promotes the AuthToken rotation and session auth support API from preview to GA status.


bigmontz (approved):
🥇

PR #1532, opened 2024-02-01
Introduce an override AuthToken support to ExecutableQuery

injectives:
This update adds support for specifying an override `AuthToken` for `ExecutableQuery`. This feature requires Bolt 5.1 or greater.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/Driver.java
+++ b/driver/src/main/java/org/neo4j/driver/Driver.java
@@ -19,6 +19,7 @@ package org.neo4j.driver;
 import java.util.concurrent.CompletionStage;
 import org.neo4j.driver.async.AsyncSession;
 import org.neo4j.driver.exceptions.ClientException;
+import org.neo4j.driver.exceptions.UnsupportedFeatureException;
 import org.neo4j.driver.reactive.ReactiveSession;
 import org.neo4j.driver.reactive.RxSession;
 import org.neo4j.driver.types.TypeSystem;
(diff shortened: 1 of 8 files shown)


driver/src/main/java/org/neo4j/driver/internal/InternalExecutableQuery.java
```diff
@@ -107,4 +116,9 @@ public Map<String, Object> parameters() {
     public QueryConfig config() {
         return config;
     }
+
+    // For testing only
+    public AuthToken authToken() {
+        return authToken;
+    }
```
bigmontz: Might worth to downgrade it to protected/package if possible.

driver/src/main/java/org/neo4j/driver/Driver.java
```diff
@@ -214,7 +215,7 @@ default <T extends BaseSession> T session(Class<T> sessionClass, SessionConfig s
      * {@link AuthToken}.
      * <p>
      * This method allows creating a session with a different {@link AuthToken} to the one used on the driver level.
-     * The minimum Bolt protocol version is 5.1. An {@link IllegalStateException} will be emitted on session interaction
+     * The minimum Bolt protocol version is 5.1. An {@link UnsupportedFeatureException} will be emitted on session interaction
```
bigmontz, line 218: Was it wrongly documented?

bigmontz (approved):
🗡️

driver/src/main/java/org/neo4j/driver/internal/InternalExecutableQuery.java
injectives: It is an internal class (module descriptor), but we might as well downgrade to package visibility here, thanks. 👍

driver/src/main/java/org/neo4j/driver/Driver.java
injectives: Unfortunately, yes. Must have been an oversight. It has now been fixed.

PR #1537, opened 2024-03-01
Add clarification to the driver close and closeAsync methods

injectives:


AndyHeap-NeoTech (approved):
A clear explanation of it's behaviour.

PR #1543, opened 2024-03-04
Introduce mTLS support

injectives:
Please note that this feature is in preview.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/ClientCertificate.java
+++ b/driver/src/main/java/org/neo4j/driver/ClientCertificate.java
@@ -14,16 +14,16 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.neo4j.driver.internal.async.connection;
+package org.neo4j.driver;
 
-import io.netty.bootstrap.Bootstrap;
-import io.netty.channel.ChannelFuture;
-import java.util.function.Function;
-import org.neo4j.driver.internal.BoltServerAddress;
+import org.neo4j.driver.internal.InternalClientCertificate;
+import org.neo4j.driver.util.Preview;
 
-public interface ChannelConnector {
-    ChannelFuture connect(
-            BoltServerAddress address,
-            Bootstrap bootstrap,
-            Function<ChannelFuture, ChannelFuture> channelFutureExtensionMapper);
-}
+/**
+ * An opaque container for client certificate used for mTLS.
+ * <p>
+ * Use {@link ClientCertificates} to create new instances.
+ * @since 5.19
+ */
+@Preview(name = "mTLS")
+public sealed interface ClientCertificate permits InternalClientCertificate {}
--- a/driver/src/main/java/org/neo4j/driver/internal/InternalRotatingClientCertificateManager.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/InternalRotatingClientCertificateManager.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [https://neo4j.com]
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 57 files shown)


driver/src/main/java/org/neo4j/driver/ClientCertificate.java
```diff
@@ -0,0 +1,29 @@
...
+import org.neo4j.driver.util.Preview;
+
+/**
+ * An opaque container for client certificate used for mTLS.
+ * <p>
+ * Use {@link ClientCertificates} to create new instances.
+ * @since 5.19
+ */
+@Preview(name = "mTLS")
+public sealed interface ClientCertificate permits InternalClientCertificate {}
```
bigmontz, line 29: I'm thinking if the `hasUpdate` should be placed in the ClientCertificate, or if it should be part of other container. It looks a bit strange the definition of the certificate be placed along with flow control information.

driver/src/main/java/org/neo4j/driver/internal/pki/DerUtils.java
```diff
@@ -0,0 +1,204 @@
...
+        }
+        return len;
+    }
+
+    /**
+     * Because Java...
+     */
+    private static int unsignedByte(ByteBuffer input) {
+        return input.get() & 0xFF;
+    }
```
bigmontz, line 203: ☕

driver/src/main/java/org/neo4j/driver/internal/security/SSLContextManager.java
```diff
@@ -0,0 +1,141 @@
...
+        CompletableFuture<SSLContext> sslContextFuture;
+        CompletionStage<SSLContext> sslContextStage = null;
+        synchronized (this) {
+            if (this.sslContextFuture == null) {
+                this.sslContextFuture = new CompletableFuture<>();
+                sslContextFuture = this.sslContextFuture;
+                var sslContext = this.sslContext;
+                sslContextStage = clientCertificateManager.getCertificate().thenApply(clientCertificate -> {
+                    var certificate = (InternalClientCertificate) clientCertificate;
+                    if (certificate.hasUpdate() || sslContext == null) {
```
bigmontz: Whe is hasUpdate set to be false?

bigmontz (changes requested):
I was following the code and I didn't find the place where the hasUpdate is set to false so the SSLContextManager can stop to reload the same certificate when used in combination with the RotatingClientCertificateManager.

driver/src/main/java/org/neo4j/driver/internal/security/SSLContextManager.java
injectives: It is either set by the user in the `ClientCertificateManager` implementation or by the provided `RotatingClientCertificateManager` implementation.

driver/src/main/java/org/neo4j/driver/internal/InternalRotatingClientCertificateManager.java
```diff
@@ -0,0 +1,63 @@
...
+                    clientCertificate.certificate(),
+                    clientCertificate.privateKey(),
+                    clientCertificate.password(),
+                    false);
+            clientCertificateStage = CompletableFuture.completedStage(clientCertificate);
+            return stage;
+        } else {
+            return clientCertificateStage;
+        }
+    }
```
bigmontz: Might worth to add unit tests for this class since this is not covered by integration or testkit ones.

bigmontz (approved):
🔐

driver/src/main/java/org/neo4j/driver/internal/InternalRotatingClientCertificateManager.java
injectives: Updated.

driver/src/main/java/org/neo4j/driver/ClientCertificate.java
injectives: Updated.

bigmontz (approved):
🔐

PR #1545, opened 2024-03-25
add Netty leak detector extension

sullis:
Add the Netty Leak Detector JUnit extension to detect
Netty resource leaks

https://github.com/nettyplus/netty-leak-detector-junit-extension


injectives:
Apologies for the delayed response. We were working on extracting the lower level networking into a separate library and it involved moving the Netty dependency as well. Due to those changes this update wouldn't fit without moving it as well.

PR #1555, opened 2024-06-01
Add support for GQL-status objects

injectives:
This update adds support for GQL-status objects defined in the GQL standard.


driver/src/main/java/org/neo4j/driver/Config.java
```diff
@@ -304,15 +315,45 @@ public long fetchSize() {
...
+    public Optional<NotificationSeverity> minimumNotificationSeverity() {
+        return Optional.ofNullable(((InternalNotificationConfig) notificationConfig).minimumSeverity());
+    }
+
+    /**
+     * Returns a set of disabled notification classifications.
+     * @return the {@link Set} of disabled {@link NotificationClassification}
+     * @since 5.22.0
+     */
+    public Set<NotificationClassification> disabledNotificationClassifications() {
```
bigmontz, line 343: Methods and new types introduced should be marked as preview, we wan't to let our interface decisions open for while.

driver/src/main/java/org/neo4j/driver/Config.java
```diff
@@ -304,15 +315,45 @@ public long fetchSize() {
 
     /**
      * Returns notification config.
+     *
      * @return the notification config
      * @since 5.7
+     * @deprecated superseded by {@link #minimumNotificationSeverity()} and
+     * {@link #disabledNotificationClassifications()}
      */
+    @Deprecated
```
bigmontz: We should not deprecate the old methods since the new ones still in preview. The deprecation should be done when we move the new methods to GA

driver/src/main/java/org/neo4j/driver/internal/summary/InternalNotification.java
```diff
@@ -16,86 +16,86 @@
...
+                .map(type -> switch (type) {
+                    case HINT -> NotificationCategory.HINT;
+                    case UNRECOGNIZED -> NotificationCategory.UNRECOGNIZED;
+                    case UNSUPPORTED -> NotificationCategory.UNSUPPORTED;
+                    case PERFORMANCE -> NotificationCategory.PERFORMANCE;
+                    case DEPRECATION -> NotificationCategory.DEPRECATION;
+                    case SECURITY -> NotificationCategory.SECURITY;
+                    case TOPOLOGY -> NotificationCategory.TOPOLOGY;
+                    case GENERIC -> NotificationCategory.GENERIC;
+                });
```
bigmontz, line 44: How do you handle unknown here?

driver/src/test/java/org/neo4j/driver/internal/util/MetadataExtractorTest.java
```diff
@@ -303,20 +313,87 @@ void shouldBuildResultSummaryWithNotifications() {
...
+        var gqlStatusObjectsIterator = summary.gqlStatusObjects().iterator();
+        var firstGqlStatusObject = (Notification) gqlStatusObjectsIterator.next();
+        var secondGqlStatusObject = (Notification) gqlStatusObjectsIterator.next();
+        assertEquals(firstNotification, firstGqlStatusObject);
+        assertEquals(secondNotification, secondGqlStatusObject);
+    }
+
+    @Test
+    @SuppressWarnings({"deprecation", "OptionalGetWithoutIsPresent"})
+    void shouldBuildResultSummaryWithGqlStatusObjects() {
```
bigmontz, line 349: It might be good to also exercises filling the default for  missing values like the diagnostic_record, diagnostic_record.OPERATION, and so on.

driver/src/main/java/org/neo4j/driver/summary/GqlStatusObject.java
```diff
@@ -0,0 +1,45 @@
...
+     * The GQLSTATUS description.
+     * @return the GQLSTATUS description
+     */
+    String statusDescription();
+
+    /**
+     * Returns the diagnostic record.
+     * @return the diagnostic record
+     */
+    Map<String, Value> diagnosticRecord();
```
bigmontz, line 46: We should have a way to simple way of printing the value for human inspection.

bigmontz (changes requested):
We should not deprecate the current apis yet and the new apis should be released as preview.

driver/src/main/java/org/neo4j/driver/summary/GqlStatusObject.java
injectives: We get this from Java and our `Value` implementations. See the example below:
```
System.out.println(firstGqlStatusObject.diagnosticRecord());

{_classification="SECURITY", OPERATION_CODE="", OPERATION="", _severity="WARNING", CURRENT_SCHEMA="/", _position={column: 424242, offset: 42, line: 4242}}
```

driver/src/test/java/org/neo4j/driver/internal/util/MetadataExtractorTest.java
injectives: Good point, this have been added. 👍

driver/src/main/java/org/neo4j/driver/internal/summary/InternalNotification.java
injectives: An empty `Optional` is returned in such case.

driver/src/main/java/org/neo4j/driver/Config.java
injectives: Good point, done. 👍

driver/src/main/java/org/neo4j/driver/Config.java
injectives: Good point, updated. 👍

driver/src/main/java/org/neo4j/driver/summary/ResultSummary.java
```diff
@@ -82,12 +84,24 @@ public interface ResultSummary {
...
      */
     List<Notification> notifications();
 
+    /**
+     * Returns a sequenced set of GQL-status objects resulting from the request execution.
+     *
+     * @return the sequenced set of GQL-status objects
+     * @since 5.22.0
+     */
+    Set<GqlStatusObject> gqlStatusObjects();
```
bigmontz, line 105: Missing @preview here.

driver/src/main/java/org/neo4j/driver/internal/summary/InternalNotification.java
bigmontz: This is the way you already did with the legacy Notification, so this is fine.

driver/src/main/java/org/neo4j/driver/summary/ResultSummary.java
injectives: Thanks!

bigmontz (approved):
🔔

driver/src/main/java/org/neo4j/driver/summary/Notification.java
```diff
@@ -28,7 +30,7 @@
  * @since 1.0
  */
 @Immutable
-public interface Notification {
+public interface Notification extends GqlStatusObject {
```
meistermeier, line 33: Why not use the `GqlStatusObject` (resp. the `InternalGqlStatusObject`) as a delegate within `(Internal)Notification`?
The inheritance seems to create more dependency and replacing `Notification` with `GqlStatusObject` in the future becomes much more complicated than just "extracting" the delegate object and make it a first level entity.

driver/src/main/java/org/neo4j/driver/NotificationCategory.java
```diff
@@ -17,50 +17,48 @@
...
-import org.neo4j.driver.internal.InternalNotificationCategory;
-import org.neo4j.driver.internal.InternalNotificationCategory.Type;
 
 /**
  * Notification category.
  *
  * @since 5.7
  */
-public sealed interface NotificationCategory extends Serializable permits InternalNotificationCategory {
+public sealed interface NotificationCategory extends Serializable permits NotificationClassification {
```
meistermeier, line 26: I don't like the interface permits but also depends on implementation (here enum reference). But on the other hand this should get deprecated and later just the `NotificationClassification` stays, or?

driver/src/main/java/org/neo4j/driver/Config.java
```diff
@@ -768,6 +810,42 @@ public ConfigBuilder withNotificationConfig(NotificationConfig notificationConfi
...
+         * Sets a set of disabled classifications for notifications produced by the server.
+         *
+         * @param disabledNotificationClassifications the set of disabled notification classifications
+         * @return this builder
+         * @since 5.22.0
+         */
+        @Preview(name = "GQL-status object")
+        public ConfigBuilder withDisabledNotificationClassifications(
+                Set<NotificationClassification> disabledNotificationClassifications) {
+            var disabledCategories = disabledNotificationClassifications == null
```
meistermeier, line 840: I don't know if it's worth extracting the conversion of categories. It is the same code as in the `GqlNotificationConfig`.

meistermeier (commented):
Completely ignored the bits around the communication layer because I have not much knowledge here.
Some remarks on dependencies and inheritance.

driver/src/main/java/org/neo4j/driver/NotificationCategory.java
injectives: Exactly. The `NotificationClassification` should replace the `NotificationCategory`.

driver/src/main/java/org/neo4j/driver/summary/Notification.java
injectives: Closing as we have discussed this now.

driver/src/main/java/org/neo4j/driver/Config.java
injectives: Closing as we have discussed this now.

PR #1559, opened 2024-07-06
Add preview support for GQL Errors

injectives:


driver/src/main/java/org/neo4j/driver/exceptions/Neo4jException.java
```diff
@@ -25,48 +32,130 @@
...
+            Map<String, Value> diagnosticRecord,
+            Throwable cause) {
         super(message, cause);
+        this.gqlStatus = Objects.requireNonNull(gqlStatus);
+        Objects.requireNonNull(statusDescription);
+        if (message != null) {
+            this.statusDescription = String.format("error: %s. %s", statusDescription, message);
+        } else {
+            this.statusDescription = String.format("error: %s.", statusDescription);
+        }
```
bigmontz: Status description will come already formatted from the database.

driver/src/main/java/org/neo4j/driver/internal/messaging/v57/MessageReaderV57.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    protected void unpackFailureMessage(ResponseMessageHandler output) throws IOException {
+        var params = unpacker.unpackMap();
+        var gqlError = unpackGqlError(params);
+        output.handleFailureMessage(gqlError);
+    }
+
+    private GqlError unpackGqlError(Map<String, Value> params) throws IOException {
+        var gqlStatus = params.get("gql_status").asString();
+        var statusDescription = params.get("status_description").asString();
```
bigmontz:
```suggestion
        var statusDescription = params.get("description").asString();
```

driver/src/main/java/org/neo4j/driver/internal/messaging/v57/MessageReaderV57.java
```diff
@@ -0,0 +1,107 @@
...
+        var code = params.get("neo4j_code").asString();
+        var message = params.get("status_message").asString();
```
bigmontz:
```suggestion
        var message = params.get("message").asString();
```

driver/src/main/java/org/neo4j/driver/internal/messaging/v57/MessageReaderV57.java
```diff
@@ -0,0 +1,107 @@
...
+    protected void unpackFailureMessage(ResponseMessageHandler output) throws IOException {
+        var params = unpacker.unpackMap();
+        var gqlError = unpackGqlError(params);
+        output.handleFailureMessage(gqlError);
+    }
+
+    private GqlError unpackGqlError(Map<String, Value> params) throws IOException {
+        var gqlStatus = params.get("gql_status").asString();
+        var statusDescription = params.get("status_description").asString();
+        var code = params.get("neo4j_code").asString();
```
bigmontz: This value, `neo4j_code`, is not present in `causes`. Since GqlCauses is something new we add, we won't need to keep the backwards compatibility here and have the code. We can also define the causes as GqlException and extend the Neo4jException from there. 

Does it make sense?

injectives: Yeah, I am aware they won't have the `neo4j_code`. I will update that bit. They will be just `Neo4jException` objects with `N/A` code to avoid breaking updates.

driver/src/main/java/org/neo4j/driver/exceptions/Neo4jException.java
```diff
@@ -78,4 +162,91 @@ public Neo4jException(String code, String message, Throwable cause) {
...
+    /**
+     * Returns the GQL error cause.
+     *
+     * @return an {@link Optional} of GQL error cause
+     * @since 5.26.0
+     */
+    @Preview(name = "GQL-error")
+    public Optional<Neo4jException> gqlCause() {
+        return findFirstGqlCause(this, Neo4jException.class);
+    }
```
bigmontz, line 239: `GQLCause` object doesn't have `code`. Some drivers abstracted `GQLException` as all the fields but code and make `Neo4jException extends GQLException` and add the `code`. This is not a blocker, but something it is good to be aware.

bigmontz (approved):
💣 💥

driver/src/main/java/org/neo4j/driver/exceptions/Neo4jException.java
injectives: Lets consider this during the preview phase.

PR #1562, opened 2024-07-23
Add Bolt 5.6

injectives:


driver/src/main/java/org/neo4j/driver/internal/util/MetadataExtractor.java
```diff
@@ -394,6 +394,8 @@ private static GqlStatusObject extractGqlStatusObject(Value value) {
             return new InternalGqlStatusObject(status, description, diagnosticRecord);
         } else {
             var title = value.get("title").asString();
+            var notificationDescription =
+                    value.containsKey("description") ? value.get("description").asString() : description;
```
bigmontz, line 399: I think it might worth to update the `MetadataExtractorTest` with a test which the description is set. The `shouldBuildResultSummaryWithGqlStatusObjects` already test the case it isn't.

bigmontz (approved):
🔔

driver/src/main/java/org/neo4j/driver/internal/util/MetadataExtractor.java
injectives: Updated. 👍

PR #1566, opened 2024-08-02
Add host-only factory using default port to ServerAddress

robsdedude:


driver/src/main/java/org/neo4j/driver/net/ServerAddress.java
```diff
@@ -47,4 +47,14 @@ public interface ServerAddress {
     static ServerAddress of(String host, int port) {
         return new BoltServerAddress(host, port);
     }
+
+    /**
+     * Create a new address with the given host using the default bolt port.
+     *
+     * @param host the host portion. Should not be {@code null}.
+     * @return new server address with the specified host and default bolt port.
+     */
```
injectives, line 57:
```suggestion
     * @since 5.24.0
     */
```

injectives (approved):

PR #1580, opened 2024-11-07
Update RxResultCursorImpl

injectives:


driver/src/main/java/org/neo4j/driver/internal/cursor/RxResultCursorImpl.java
```diff
@@ -47,428 +51,362 @@
...
 
     @Override
-    public void onRecord(Value[] fields) {
-        var record = new InternalRecord(runSummary.keys(), fields);
-        synchronized (this) {
-            updateRecordState(RecordState.HAD_RECORD);
-            decrementDemand();
-        }
-        recordConsumer.accept(record, null);
+    public synchronized boolean isDone() {
```
AndyHeap-NeoTech, line 170: Minor thought. Perhaps this should be named isSuccessful or something like that as it only returns true on the SUCCEEDED case. Not important though.

AndyHeap-NeoTech (approved):
One very minor comment on a method name that I'm not even sure needs changing :) The refactor looks good, approved.

PR #1583, opened 2024-11-13
Update handling of cancellation state in reactive result

injectives:


meistermeier (approved):
Approval based on Spring Data Neo4j tests running fine again.

PR #1600, opened 2025-01-01
Introduce home database resolution cache

injectives:
This update adds support for Bolt protocol 5.8 and introduces home database resolution cache that works with Bolt protocol 5.8+ when routing scheme is used. The cache is maintained by the driver internally and is not exposed in the API, its purpose is to reduce the number of Bolt exchanges when home database resolution is needed.

### DIFF
--- a/bolt-api-routed/src/main/java/org/neo4j/driver/internal/bolt/routedimpl/impl/cluster/RoutingTableHandlerImpl.java
+++ b/bolt-api-routed/src/main/java/org/neo4j/driver/internal/bolt/routedimpl/impl/cluster/RoutingTableHandlerImpl.java
@@ -208,4 +208,12 @@ public class RoutingTableHandlerImpl implements RoutingTableHandler {
     public RoutingTable routingTable() {
         return routingTable;
     }
+
+    @Override
+    public synchronized boolean staleRoutingTable(AccessMode mode) {
+        if (refreshRoutingTableFuture != null) {
+            return true;
+        }
+        return routingTable.isStaleFor(mode);
+    }
 }
--- a/driver/src/main/java/org/neo4j/driver/internal/homedb/HomeDatabaseCacheImpl.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/homedb/HomeDatabaseCacheImpl.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [https://neo4j.com]
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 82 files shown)


bolt-api-routed/src/main/java/org/neo4j/driver/internal/bolt/routedimpl/impl/cluster/RoutingTableHandlerImpl.java
```diff
@@ -208,4 +208,12 @@ public boolean isRoutingTableAged() {
     public RoutingTable routingTable() {
         return routingTable;
     }
+
+    @Override
+    public synchronized boolean staleRoutingTable(AccessMode mode) {
```
StephenCathcart, line 213: Would `isRoutingTableStale` be more fitting?

driver/src/main/java/org/neo4j/driver/internal/homedb/HomeDatabaseCacheImpl.java
```diff
@@ -0,0 +1,100 @@
+/*
```
StephenCathcart, line 1: Discuss general concurrency

driver/src/main/java/org/neo4j/driver/internal/homedb/HomeDatabaseCacheKey.java
```diff
@@ -0,0 +1,43 @@
...
+            var tokenMap = ((InternalAuthToken) overrideAuthToken).toMap();
+            var scheme = tokenMap.get(InternalAuthToken.SCHEME_KEY);
+            if (scheme.asString().equals("basic")) {
+                key = new MapHomeDatabaseCacheKey(Collections.singletonMap(
+                        InternalAuthToken.PRINCIPAL_KEY, tokenMap.get(InternalAuthToken.PRINCIPAL_KEY)));
+            } else {
+                key = new MapHomeDatabaseCacheKey(tokenMap);
+            }
+        } else {
+            key = DriverHomeDatabaseCacheKey.INSTANCE;
```
StephenCathcart, line 39: Is this the "DEFAULT" key as per the ADR?

StephenCathcart (approved):
Good job! 🎉

bolt-api-routed/src/main/java/org/neo4j/driver/internal/bolt/routedimpl/impl/cluster/RoutingTableHandlerImpl.java
StephenCathcart: Discussed

driver/src/main/java/org/neo4j/driver/internal/homedb/HomeDatabaseCacheImpl.java
StephenCathcart: Discussed, all good 👍

driver/src/main/java/org/neo4j/driver/internal/homedb/HomeDatabaseCacheKey.java
StephenCathcart: - yes

PR #1605, opened 2025-01-27
Add support for Bolt Protocol Handshake Manifest v1

injectives:
This update brings support for a new Bolt handshake and also removes support for Bolt protocol versions 4 and 4.1.


StephenCathcart (approved):
Nice implementation 👍

PR #1608, opened 2025-01-27
Update `Neo4jException.gqlCause`

injectives:
This update fixes `Neo4jException.gqlCause` that did not return subtypes of `Neo4jException`. In addition, only GQL causes from non interrupted error chains will be returned.


StephenCathcart (approved):
Change and tests look good 👍

PR #1622, opened 2025-03-05
Update retry delay multiplier in ExponentialBackoffRetryLogic

injectives:
This fixes a bug in ExponentialBackoffRetryLogic and ensures that the retry delay works as expected.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/retry/ExponentialBackoffRetryLogic.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/retry/ExponentialBackoffRetryLogic.java
@@ -325,8 +325,8 @@ public class ExponentialBackoffRetryLogic implements RetryLogic {
         if (initialRetryDelayMs < 0) {
             throw new IllegalArgumentException("Initial retry delay should >= 0: " + initialRetryDelayMs);
         }
-        if (multiplier < 1.0) {
-            throw new IllegalArgumentException("Multiplier should be >= 1.0: " + multiplier);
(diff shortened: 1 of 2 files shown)


driver/src/main/java/org/neo4j/driver/internal/retry/ExponentialBackoffRetryLogic.java
```diff
@@ -45,14 +45,14 @@ public class ExponentialBackoffRetryLogic implements RetryLogic {
     public static final long DEFAULT_MAX_RETRY_TIME_MS = SECONDS.toMillis(30);
 
     private static final long INITIAL_RETRY_DELAY_MS = SECONDS.toMillis(1);
-    private static final double RETRY_DELAY_MULTIPLIER = 1;
+    private static final double RETRY_DELAY_MULTIPLIER = 2.0;
```
AndyHeap-NeoTech, line 48: I'll comment here as it is not possible to comment on a non-changed line. 

Line 328 should probably check that multiplier that is supplied is >= 2.0 rather than 1.0.

AndyHeap-NeoTech (approved):
Approved with one comment

driver/src/main/java/org/neo4j/driver/internal/retry/ExponentialBackoffRetryLogic.java
injectives: I have updated the check. There is also a new test that verifies that the default value is `2.0`.

PR #1629, opened 2025-03-17
perf: Optimise Bolt values unpacking

injectives:
This update optimises Bolt values unpacking by implementing specific type mappings in `ValueFactory`.


StephenCathcart (approved):
👍

PR #1635, opened 2025-04-15
docs: Update Driver#close() and Driver#closeAsync() documentation

injectives:
The objective is to improve `Driver#close()` and `Driver#closeAsync()` methods documentation and to emphasise that it should be called only when all driver interactions are finished.


robsdedude (approved):
:cherries::hair_pick:

PR #1636, opened 2025-04-22
fix: Ensure reactive transaction subsequent runs work during streaming

injectives:
When using reactive transaction, it is possible for one thread to initiate multiple query runs with subsequent results streaming, while the driver thread would be supplying the respective results and handling the streaming. Since all of this uses the same Bolt connection, it is important to make sure it is shared correctly to avoid mismanagement of the Bolt message handling.


MaxAake (approved):
This looks good to me 👍🏻

PR #1637, opened 2025-04-22
feat(object-mapping): Add support for mapping list value to array

injectives:
The main objective of this update is to add support for mapping list value to array. Given that `org.neo4j.driver.Values#value(Object)` method allows creating list value from array (excluding `byte[]`), it makes sense to support the mapping back from list value to array.

Example:
```java
var array = new float[] {0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10};
var values = Values.value(array);
assertArrayEquals(array, values.as(float[].class));
```

As `org.neo4j.driver.Values#value(Object)` supports mapping `char[]` and `short[]` to list values, the supported mappings from string and integer values have also been extended.

Updated list of supported mappings:
- `TypeSystem#LIST` -> `List`, `T[]` as long as list elements may be mapped to the array component type (for example, `char[]`, `boolean[]`, `String[]`, `long[]`, `int[]`, `short[]`, `double[]`, `float[]`)
- `TypeSystem#STRING` -> `String`, `char`, `Character`
- `TypeSystem#INTEGER` -> `long`, `Long`, `int`, `Integer`, `short`, `Short`, `double`, `Double`, `float`, `Float`

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/Record.java
+++ b/driver/src/main/java/org/neo4j/driver/Record.java
@@ -129,6 +129,8 @@ public interface Record extends MapAccessorWithDefaultValue {
      * {@code null} value (this includes primitive types), an alternative constructor that excludes it must be
      * available.
      * <p>
+     * The mapping only works for types with directly accessible constructors, not interfaces or abstract types.
+     * <p>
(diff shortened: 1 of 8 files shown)


MaxAake (approved):
Good stuff!

PR #1638, opened 2025-04-23
feat(object-mapping): Add support for mapping java.lang.Record to value

injectives:
Please note that this is a [feature preview](https://github.com/neo4j/neo4j-java-driver/blob/5.0/README.md#preview-features).

This update adds support for mapping `java.lang.Record` to map value that may be useful when writing data to Neo4j.

The following new method has been added to `org.neo4j.driver.Values`:
```java
Value value(java.lang.Record record)
```

It returns a map value based on record components of a given `java.lang.Record`.

Example (similar to the [Neo4j Movies Database](https://github.com/neo4j-graph-examples/movies)):
```java
// assuming the following Java record
public record Movie(String title, String tagline, long released) {}
// a new movie may be created in the following way
var movie = new Movie("title", "tagline", 2025);
var movieValue = Values.value(movie);
driver.executableQuery("CREATE (:Movie $movie)")
         .withParameters(Map.of("movie", movieValue))
         .execute();
```

Because the driver methods accepting a `Map<String, Object>` as query parameters automatically map values to value instances, it is possible to avoid mapping movie explicitly:
```java
var movie = new Movie("title", "tagline", 2025);
driver.executableQuery("CREATE (:Movie $movie)")
         .withParameters(Map.of("movie", movie))
         .execute();
```

Assuming movie titles being unique, it is possible to update the created movie in the following way:
```java
var updatedMovie = new Movie("title", "updated tagline", 2024);
driver.executableQuery("""
                 MATCH (movie:Movie {title: $movie.title})
                 SET movie += $movie
                 """)
         .withParameters(Map.of("movie", updatedMovie))
         .execute();
```

The `Property` annotation may be used to override the record component name.
```java
public record Movie(String title, String tagline, @Property("releasedYear") long released) {}
```

Note that those record components that have `null` value will be excluded from the map value.

It is also important to understand that sending all properties over network may not always be desirable and will depend on a use-case.

This new mapping as also available via the `Value value(Object value)` method, but with a note that it is still in preview status, not GA status.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/Values.java
+++ b/driver/src/main/java/org/neo4j/driver/Values.java
@@ -456,6 +456,7 @@ public final class Values {
      * @see java.lang.Record
      * @see java.lang.reflect.RecordComponent
      * @see Property
+     * @throws ClientException when mapping fails
      * @since 5.28.5
      */
     @Preview(name = "Object mapping")
--- a/driver/src/main/java/org/neo4j/driver/Values.java
+++ b/driver/src/main/java/org/neo4j/driver/Values.java
@@ -445,8 +445,8 @@ public final class Values {
      * It is also important to understand that sending all properties over network may not always be desirable and will
      * depend on a use-case.
      * <p>
-     * In addition, please note that while this mapping may build nested structures, like map of maps, there may be
-     * limitations on how those may be used by the database. Please read the Neo4j Cypher Manual for more up-to-date
+     * In addition, please note that while this mapping allows nested structures, like map of maps, there may be
+     * limitations on how those are supported by the database. Please read the Neo4j Cypher Manual for more up-to-date
      * details. For example, at the time of writing, it is not possible to store maps as properties
      * (see the following <a href="https://neo4j.com/docs/cypher-manual/current/values-and-types/property-structural-constructed/#constructed-types">page</a>).
      *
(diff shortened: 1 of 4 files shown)


AndyHeap-NeoTech (approved):

MaxAake (approved):
✅ Very nice addition!

driver/src/main/java/org/neo4j/driver/Values.java
```diff
@@ -386,6 +395,98 @@ public static Value value(Stream<Object> stream) {
...
+     * }
+     * </pre>
+     * Note that those record components that have {@code null} value will be excluded from the map value.
+     * <p>
+     * It is also important to understand that sending all properties over network may not always be desirable and will
+     * depend on a use-case.
+     * <p>
+     * In addition, please note that while this mapping may build nested structures, like map of maps, there may be
+     * limitations on how those may be used by the database. Please read the Neo4j Cypher Manual for more up-to-date
+     * details. For example, at the time of writing, it is not possible to store maps as properties
```
meistermeier:
```suggestion
     * In addition, please note that while this mapping allows nested structures, like map of maps, there may be
     * limitations on how those are supported by the database. Please read the Neo4j Cypher Manual for more up-to-date
     * details. For example, at the time of writing, it is not possible to store maps as properties
```
Trying to reduce the recurrence of "may be"s.

meistermeier (approved):

PR #1646, opened 2025-05-12
feat(bookmarks): delete deprecated multi value support in Bookmark

injectives:
BREAKING CHANGE: the `org.neo4j.driver.Bookmark` no longer supports having multiple bookmark values, multiple values are supportted by simply having multiple instances of `Bookmark`


MaxAake (approved):
Looks good!

PR #1648, opened 2025-05-13
feat(logging): add support for System.Logger and deprecate Logging

injectives:
BREAKING CHANGE: the default driver logging implementation has been changed from `Logging#none()` to `Logging.systemLogging()`.

The logging abstraction has been deprecated in favour of the `System.Logger`. Specifically, this applies to the following::
- `org.neo4j.driver.Logging`
- `org.neo4j.driver.Logger`
- `org.neo4j.driver.Config.ConfigBuilder#withLogging(Logging)`
- `org.neo4j.driver.Config#logging()`

Once the logging abstraction is deleted, the driver is expected to use the `System.Logger` only.

### DIFF
--- a/benchkit-backend/src/main/java/neo4j/org/testkit/backend/Config.java
+++ b/benchkit-backend/src/main/java/neo4j/org/testkit/backend/Config.java
@@ -17,12 +17,10 @@
 package neo4j.org.testkit.backend;
 
 import java.net.URI;
-import java.util.logging.Level;
 import org.neo4j.driver.AuthToken;
 import org.neo4j.driver.AuthTokens;
-import org.neo4j.driver.Logging;
 
-public record Config(int port, URI uri, AuthToken authToken, Logging logging) {
+public record Config(int port, URI uri, AuthToken authToken) {
     static Config load() {
         var env = System.getenv();
         var port = Integer.parseInt(env.getOrDefault("TEST_BACKEND_PORT", "9000"));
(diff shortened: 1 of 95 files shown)


michael-simons (approved):
It works in my test, but while I said I really don't mind for the future JUL or System.logger, do you think it is wise to replace JUL with System.logger as part of the deprecation per default?
It would be good to offer the System.logger variant, but keep the JUL logger as default (since you didn't delete it (and if you would, I wouldn't have approved the PR))


Especially as we stated this including and mentioning even the level

> * Provide a logging implementation for the driver to use. Java logging framework {@link java.util.logging} with {@link Level#INFO} is used by default.
> * Callers are expected to either implement {@link Logging} interface or provide one of the existing implementations available from static factory
> * methods in the {@link Logging} interface.

I don't think it's too hard to construct a scenario in which people are relying on that, and they will see a bit of a surprise, without warning.

In the end, I leave it to your judgement. 
Maybe I overthink the switch, but we have seen the oddest issues with changing the logging in the core product, and the issues created in some constellations with bridges, fascade etc being on the class path or not (both variants caused trouble).

injectives:
@michael-simons, thank you for having a look at this PR! 👍 

> do you think it is wise to replace JUL with System.logger as part of the deprecation per default?

The current default is actually no logging (`Logging#none()`). This has been the case since August 2021, basically since `4.4.0` release as part of the drivers unification process. Unfortunately, that part of documentation was not updated back then. This PR finally updates it.

So, we are not actually changing from JUL to System Logger, we are changing from no logging to System Logger.

Sineaggi:
@michael-simons FWIW the System.Logger defaults to JUL if JUL is available (i.e. the logger module hasn't been jlinked out) and that's the only logger available. Otherwise, in production systems most devs would use Logback/Log4j, both of which offer System.Logger implementations.

PR #1649, opened 2025-05-16
feat(session): delete deprecated session methods

injectives:
BREAKING CHANGE: this update removes deprecated session methods on the `Driver` level. In addition, this also removes the experimental `defaultTypeSystem()` method.


StephenCathcart (approved):
👍

PR #1653, opened 2025-06-12
feature(bom): introduce BOM and update to Bolt Connection 4.0.0

injectives:
This update introduces a Bill of Materials (BOM) for Neo4j Java Driver with the following Maven coordinates:
```
<groupId>org.neo4j.driver</groupId>
<artifactId>neo4j-java-driver-bom</artifactId>
```

In addition, it adds support for loading `BoltConnectionProviderFactory` using `java.util.ServiceLoader` and updates Bolt Connection to `4.0.0`. This support is an internal implementation detail and is NOT a public feature that should be used programmatically. Only official implementations of `BoltConnectionProviderFactory` are supported and the extend of such support is individual to each implementation. The `neo4j-java-driver-all` bundles everything in the same artifact.


MaxAake (approved):
✅

PR #1657, opened 2025-06-24
perf(value): optimise value handling for Bolt Connection

injectives:
The objective of this update is to reduce the amount of memory allocations needed when Bolt Connection values are used. Driver internal values now implement Bolt Connection values directly and no longer need a wrapper adapter. In addition, lists are used instead of arrays to prevent array copying.


MaxAake (approved):
LGTM!

PR #1660, opened 2025-06-26
fix(retry): make RetryLogic executor threads daemon

injectives:
This update makes sure `RetryLogic` executor threads are daemon.


StephenCathcart (approved):
👍

PR #1667, opened 2025-07-03
feat(gql-status-object): Introduce GqlNotification

injectives:
BREAKING CHANGE: this update separates the new GQL Status Object hierarchy from the legacy `Notification` type by introducing a new `GqlNotification` interface that extends the `GqlStatusObject`.

The update is based on the feedback received on the GQL Status Object feature, which is currently in preview. This should make it easier to understand the new GQL Status Objects and migrate from using the legacy `Notification` to the new types.

Eventually, the new hierarchy is expected to replace the legacy `Notification` that is expected to be deprecated and phased out at a later date.

### DIFF
--- a/driver/clirr-ignored-differences.xml
+++ b/driver/clirr-ignored-differences.xml
@@ -824,4 +824,29 @@
         <method>org.neo4j.driver.types.Type VECTOR()</method>
     </difference>
 
+    <difference>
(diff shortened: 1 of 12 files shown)


AndyHeap-NeoTech (approved):
LGTM

AndyHeap-NeoTech (approved):
👍

PR #1669, opened 2025-07-07
feat(gql-status-object): make GQL Status Object GA

injectives:
This update makes GQL Status Object GA and deprecates Notification.


MaxAake (approved):
LGTM! 👍🏻

PR #1672, opened 2025-07-09
feat(gql-error): add GQLSTATUS finders

injectives:
In some situations it might be needed to check if GQL error or any error in its GQL error chain has a specific GQLSTATUS.

This update adds 2 new methods to `Neo4jException` to make it easier:
- `boolean containsGqlStatus(String gqlStatus)`
- `Optional<Neo4jException> findByGqlStatus(String gqlStatus)`


MaxAake (approved):
🌸 ✅

PR #1676, opened 2025-07-13
feat(gql-status-object): add legacy notification fields

injectives:
This update is for Neo4j private use only. It adds the following 3 legacy notification fields to `InternalGqlNotification`:
- `code`
- `title`
- `description`

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/summary/InternalGqlNotification.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/summary/InternalGqlNotification.java
@@ -17,6 +17,7 @@
 package org.neo4j.driver.internal.summary;
 
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import org.neo4j.driver.NotificationClassification;
 import org.neo4j.driver.NotificationSeverity;
@@ -31,6 +32,11 @@ public final class InternalGqlNotification extends InternalGqlStatusObject imple
     private final NotificationClassification classification;
     private final String rawClassification;
 
+    // private Neo4j use only
+    private final String code;
+    private final String title;
+    private final String description;
+
     public InternalGqlNotification(
             String gqlStatus,
             String statusDescription,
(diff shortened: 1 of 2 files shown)


driver/src/main/java/org/neo4j/driver/internal/summary/InternalGqlNotification.java
```diff
@@ -73,11 +85,47 @@ public Optional<String> rawClassification() {
...
     @Override
     public String toString() {
         return "InternalGqlNotification{" + "gqlStatus='"
                 + gqlStatus + '\'' + ", statusDescription='"
                 + statusDescription + '\'' + ", diagnosticRecord="
-                + diagnosticRecord + '}';
+                + diagnosticRecord + '\'' + ", code='"
+                + code + '\'' + ", title='"
+                + title + '\'' + ", description='"
+                + description + '}';
```
MaxAake, line 114: I'm unsure if it's correct to add the internal use properties to the toString. Won't this mean these 3 irrelevant (to the user) properties will be added to any printing of the Notification?

driver/src/main/java/org/neo4j/driver/internal/summary/InternalGqlNotification.java
```diff
@@ -73,11 +85,47 @@ public Optional<String> rawClassification() {
...
+    }
+
+    // private Neo4j use only
```
MaxAake, line 117: Is there any documentation tag that could be used to further clarify that these aren't for customer use? I know it's an internal class but it's still present in the API documentation.

I'm unsure about the ramifications for doing this in Java but perhaps marking them deprecated?

MaxAake (commented):
💬

driver/src/main/java/org/neo4j/driver/internal/summary/InternalGqlNotification.java
MaxAake: resolved in private conversation

driver/src/main/java/org/neo4j/driver/internal/summary/InternalGqlNotification.java
MaxAake: resolved in private conversation

MaxAake (approved):
✅

PR #1677, opened 2025-07-14
feat(gql-status-object): add legacy notification fields

injectives:
This update is for Neo4j private use only. It adds the following 3 legacy notification fields to `InternalGqlNotification`:
- `code`
- `title`
- `description`


MaxAake (approved):
✅

PR #1681, opened 2025-07-21
fix(object-mapping): try making record components accessible

injectives:
In some cases reading components of `record` may fail as their accessors may not be accessible. This update makes sure the driver tries to make them accessible. For instance, this is useful when using local `record` in a stricter runtime. The `ObjectMappingTests` actually had a test for this, but it did not fail when it was run as a unit test. The tests have been moved to integration tests without any other update.


MaxAake (approved):
✅

PR #1685, opened 2025-08-27
feat(acquisition): apply acquisition timeout to all steps

injectives:
BREAKING CHANGE: `org.neo4j.driver.Config.ConfigBuilder#withConnectionAcquisitionTimeout(long, TimeUnit)` no longer accepts negative values, matching the handling of `org.neo4j.driver.Config.ConfigBuilder#withConnectionTimeout(long, TimeUnit)`. In addition, the acquisition timeout is applied for the entire connection acquisition process that may include client-side routing and other logic. Previously, the connection acquisition timeout applied individually per every acquisition from individual connection pool and only when all connections were busy.


MaxAake (approved):
✅ Looks good!

PR #1686, opened 2025-08-28
feat(observability): allow setting null provider

injectives:
Setting `null` provider used to throw `IllegalArgumentException`. This update fixes that issue.


MaxAake (approved):
✅

PR #1687, opened 2025-09-02
docs(retries): add a note about implicit transaction

injectives:
This update adds a note about implicit transaction to `RetryableException`.


AndyHeap-NeoTech (approved):
A clear explanation.

AndyHeap-NeoTech (approved):

PR #1689, opened 2025-09-03
fix(acquisition): avoid infinite connection timeout if there is a limit

injectives:
As acquisition timeout effectively includes connection timeout, the latter should not be left infinite when the former is limited.


MaxAake (approved):
✅

PR #1690, opened 2025-09-04
feat(native): add support for Netty native transports

injectives:
This update brings support for using Netty native transports.

Providing that a valid native dependency is added, it will be discovered and used instead of the default `NIO` transport.

For example:
```
<dependency>
    <groupId>io.netty</groupId>
    <artifactId>netty-transport-native-io_uring</artifactId>
    <classifier>linux-x86_64</classifier>
</dependency>
```


MaxAake (approved):
✅

PR #1691, opened 2025-09-08
feat(unsupportedType): add support for Bolt Unsupported Type

injectives:
This update adds support for a new `TypeSystem#UNSUPPORTED()` type. The unsupported type may occur when a new type is introduced in the Neo4j server and the driver is connected over an older Bolt Protocol version that does not support this type. A new `UnsupportedType` object provides information about the unsupported type and the minimum Bolt version needed to support it. However, the `UnsupportedType` object itself requires at least Bolt Protocol 6.0.

### DIFF
--- a/driver/clirr-ignored-differences.xml
+++ b/driver/clirr-ignored-differences.xml
@@ -935,4 +935,10 @@
         <differenceType>8001</differenceType>
     </difference>
 
+    <difference>
+        <className>org/neo4j/driver/types/TypeSystem</className>
+        <differenceType>7012</differenceType>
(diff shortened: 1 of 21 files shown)


michael-simons:
So we are not giving the user access to the raw binary data it seems? Fine with me, thought it would be different though

michael-simons (approved):

injectives:
> So we are not giving the user access to the raw binary data it seems? Fine with me, thought it would be different though

No, unfortunately this would not be available on Bolt level. The reasons are in ADR-033 that is still open.

MaxAake (approved):
✅

PR #1696, opened 2025-09-11
feat(tfo): add experimental support for TFO

injectives:
This update adds experimental support for TCP Fast Open.

A working TCP Fast Open setup enables the driver to start one of the following during TCP handshake by sending early data:
- Bolt Handshake (for connections not needing encryption)
- TLS Handshake (for connections needing encryption)

The TCP Fast Open is only supported when all of the following is true:
- The system the driver is running on supports and has TCP Fast Open enabled.
- One of the following connection URI schemes is used:
  - neo4j
  - neo4j+s
  - neo4j+ssc
  - bolt
  - bolt+s
  - bolt+ssc
- One of the following Netty Native Transports that is compatible with the system the driver is running on is added to the runtime by the user:
 - netty-transport-native-io_uring (Netty 4.2+ only)
 - netty-transport-native-epoll
 - netty-transport-native-kqueue

Even when all of the above is true, it is important to make sure that the endpoint the driver connects to also supports and has TCP Fast Open enabled. Otherwise, an extra slight overhead is likely and end-to-end TCP Fast Open benefits will not be available.

At present, this is an opt-in `Config` option available via `ConfigBuilder#withTcpFastOpenEnabled(boolean)`. It is `false` by default.

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/Config.java
+++ b/driver/src/main/java/org/neo4j/driver/Config.java
@@ -40,6 +40,7 @@ import org.neo4j.driver.internal.observation.DriverObservationProvider;
 import org.neo4j.driver.internal.retry.ExponentialBackoffRetryLogic;
 import org.neo4j.driver.net.ServerAddressResolver;
 import org.neo4j.driver.observation.ObservationProvider;
+import org.neo4j.driver.util.Experimental;
 import org.neo4j.driver.util.Immutable;
 import org.neo4j.driver.util.Preview;
 import org.neo4j.driver.util.Resource;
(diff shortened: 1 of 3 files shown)


driver/src/main/java/org/neo4j/driver/Config.java
```diff
@@ -874,6 +899,47 @@ public ConfigBuilder withTelemetryDisabled(boolean telemetryDisabled) {
...
+         *     <ul>
+         *         <li>netty-transport-native-io_uring (Netty 4.2+ only)</li>
+         *         <li>netty-transport-native-epoll</li>
+         *         <li>netty-transport-native-kqueue</li>
+         *     </ul>
+         *     </li>
+         * </ul>
+         * <p>
+         * Even when all of the above is true, it is important to make sure that the endpoint the driver connects to also
+         * supports and has TCP Fast Open enabled.
```
MaxAake:
```suggestion
         * Even when all of the above is true; if the endpoint the driver connects to does not also have TCP Fast Open enabled then TCP Fast Open won't function and extra network overhead will be incurred.
```

MaxAake (commented):
✍🏻 ❓

driver/src/main/java/org/neo4j/driver/Config.java
injectives: Updated, thanks!

MaxAake (approved):
✅

PR #1699, opened 2025-09-15
feat(tfo): update config naming

injectives:
BREAKING CHANGE: the configuration option has been renamed to highlight the attempt nature of it.


michael-simons (approved):
Nice that we are aligned on this one. I think the `wither` is ok as well, it's consistent with what you have. Go ahead! :)

PR #1708, opened 2025-10-11
feature(observation): Set driver dependency as provided

injectives:
Driver should not be added transitively, observation modules should just be additive.


michael-simons (approved):
Makes a lot of sense 👍

PR #1730, opened 2026-03-06
ProfiledPlan: expose existence for all metrics

robsdedude:
Deprecate `ResultSummary#profile()` returning a `ProfiledPlan` in favor of `ResultSummary#queryProfile` returning a `Profile`.

The main difference between the profile interfaces is that the new one exposes whether stats (such as `dbHits`, `time`, etc.) have been recorded by returning optional types. Further, `ProfiledPlan#records()` corresponds to the new `Profile#rows()` to better align with Neo4j's embedded API, Bolt, and the raw arguments attached to the profile steps.

According to a chat with the Cypher runtime team, the absence of the value is not equivalent to it being zero. Therefore it makes sense to express this information in the API. The information is encoded in the bolt protocol through absence or existence of the respective fields. The driver logic has been changed to use this information instead of checking for fallback zero values.

Closes: DRIVERS-282

Depends on:
 * https://github.com/neo4j-drivers/testkit/pull/705


robsdedude:
Please don't merge yet. I'd like to wait for this work to be ready for merging across all drivers first.

driver/src/main/java/org/neo4j/driver/summary/ResultSummary.java
```diff
@@ -85,9 +85,22 @@ public interface ResultSummary {
...
+    /**
+     * This describes how the database did execute your query.
+     * <p>
+     * If the query you executed {@link #hasProfile() was profiled}, the query plan will contain detailed
+     * information about what each step of the plan did. That more in-depth version of the query plan becomes
+     * available here.
+     *
+     * @return profiled query plan for the executed query if available, otherwise null
+     */
+    Profile queryProfile();
```
injectives: Given that `Profile` may be missing, it is better to have `Optional<Profile>` as return type and deprecate `hasProfile()` method.

driver/src/main/java/org/neo4j/driver/summary/ProfiledPlan.java
```diff
@@ -74,5 +76,5 @@ public interface ProfiledPlan extends Plan {
     long time();
 
     @Override
-    List<ProfiledPlan> children();
+    List<? extends ProfiledPlan> children();
```
injectives: I would leave it unchanged in deprecated class.

driver/src/main/java/org/neo4j/driver/summary/Profile.java
```diff
@@ -0,0 +1,74 @@
...
+package org.neo4j.driver.summary;
+
+import java.util.List;
+import java.util.OptionalDouble;
+import java.util.OptionalLong;
+
+/**
+ * This is the same as a regular {@link Plan} - except this plan has been executed, meaning it also contains detailed information about how much work each
+ * step of the plan incurred on the database.
+ * @since 6.1
```
injectives:
```suggestion
 * @since 6.2.0
```

driver/src/main/java/org/neo4j/driver/summary/ResultSummary.java
```diff
@@ -85,9 +85,22 @@ public interface ResultSummary {
...
 
+    /**
+     * This describes how the database did execute your query.
+     * <p>
+     * If the query you executed {@link #hasProfile() was profiled}, the query plan will contain detailed
+     * information about what each step of the plan did. That more in-depth version of the query plan becomes
+     * available here.
+     *
+     * @return profiled query plan for the executed query if available, otherwise null
+     */
```
injectives, line 118:
```suggestion
     * @since 6.2.0
     */
```

driver/src/main/java/org/neo4j/driver/summary/ResultSummary.java
injectives: I would also make it a default method:
```java
default Optional<Profile> queryProfile() {
    return Optional.empty();
}
```

injectives: You might want to update the `plan()` and `hasPlan()` methods resprectively.

driver/src/main/java/org/neo4j/driver/summary/Profile.java
```diff
@@ -0,0 +1,74 @@
...
+     * @return the ratio of page cache hits to total number of lookups or 0 if no data is available
+     */
+    OptionalDouble pageCacheHitRatio();
+
+    /**
+     * Returns the amount of time spent in the associated execution step if it was recorded.
+     *
+     * @return the amount of time spent in the associated execution step if it was recorded
+     */
+    OptionalLong time();
```
injectives: I wonder if `Optional<Duration>` would be a more natural fit for this.

driver/src/main/java/org/neo4j/driver/summary/ResultSummary.java
robsdedude: > You might want to update the plan() and hasPlan() methods resprectively.

The means I have to rename `plan()` to be able to change its signature with a deprecation cycle. 

  1. Should we? Doing so, would mean that Java will be the odd one out with a different name.
  2. If yes, what name? `queryPlan()`?
  3. If no, do you still want me to make `queryProfile()` return an `Optional`? Because then it'd be `null`able `plan()`, but `Optional` `queryProfile()` which is unsatisfyingly inconsistent imo.

driver/src/main/java/org/neo4j/driver/summary/ProfiledPlan.java
robsdedude: Thanks. Leftover from a previous design.

driver/src/main/java/org/neo4j/driver/summary/Profile.java
robsdedude: You tell me if it is in Java land.

For context:

 * all other drivers expose a raw number
 * SSR and Query API will want this back as a number
 * I reckon turning this into a `Duration` won't be hard for users to do

Side note: the docs don't even state the unit in which the time is measured... I guess that should be improved at least.

injectives: In my view, the language has a dedicated type for expressing duration, so we should use it. It can be turned back to number quite easily.

driver/src/main/java/org/neo4j/driver/summary/ResultSummary.java
injectives: Yeah, I would just add `queryPlan()` to keep things consistent.

driver/src/main/java/org/neo4j/driver/internal/summary/InternalProfile.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Optional;
+import java.util.OptionalDouble;
+import java.util.OptionalLong;
+import java.util.function.Function;
+import org.neo4j.driver.Value;
+import org.neo4j.driver.summary.Profile;
+
+public class InternalProfile extends InternalPlan<Profile> implements Profile {
+
+    private final OptionalLong dbHits;
```
injectives: We try to avoid using `Optional` as fields and parameters. Some inspection solutions raise this as an issue as well. For example: https://www.jetbrains.com/help/inspectopedia/OptionalUsedAsFieldOrParameterType.html#-l19x0b_3

driver/src/main/java/org/neo4j/driver/internal/summary/InternalResultSummary.java
```diff
@@ -170,7 +192,7 @@ public int hashCode() {
                 queryType,
                 counters,
                 plan,
-                profile,
+                profiledPlan,
```
injectives:
```suggestion
                profiledPlan,
                profile,
```

Same as https://github.com/neo4j/neo4j-java-driver/pull/1730/changes#r3350974965

driver/src/main/java/org/neo4j/driver/internal/summary/InternalResultSummary.java
injectives, line 172:
```suggestion
                && Objects.equals(profiledPlan, that.profiledPlan)
                && Objects.equals(profile, that.profile)
```

I suggest having both here while there are 2 fields. Otherwise, we would have to document that `Profile` is always equals to `ProfiledPlan` to prevent confusion, but it still feels more fragile.

driver/src/main/java/org/neo4j/driver/internal/summary/InternalResultSummary.java
```diff
@@ -184,7 +206,7 @@ public String toString() {
                 + databaseInfo + ", queryType="
                 + queryType + ", counters="
                 + counters + ", plan="
-                + plan + ", profile="
+                + plan + ", queryProfile="
```
injectives, line 198:
```suggestion
                + plan + ", profile="
                + profiledPlan + ", queryProfile="
```

We tend to include most fields in `String` representation.

driver/src/main/java/org/neo4j/driver/internal/summary/InternalResultSummary.java
```diff
@@ -37,7 +39,11 @@ public class InternalResultSummary implements ResultSummary {
     private final QueryType queryType;
     private final SummaryCounters counters;
     private final Plan plan;
-    private final ProfiledPlan profile;
+
+    @SuppressWarnings("deprecation")
+    private final ProfiledPlan profiledPlan;
```
injectives: It would be nice if we could avoid having `ProfiledPlan` as a field, but just implement a wrapper for `Profile` that we could create on-demand when the `profile` method is called. In such case, the comments regarding `equals`, hashCode` and `toString` would no longer be applicable.

driver/src/test/java/org/neo4j/driver/integration/SummaryIT.java
```diff
@@ -80,8 +81,8 @@ void shouldContainBasicMetadata() {
         assertThat(summary.queryType(), equalTo(QueryType.READ_ONLY));
         assertThat(summary.query().text(), equalTo(query));
         assertThat(summary.query().parameters(), equalTo(parameters));
-        assertFalse(summary.hasPlan());
-        assertFalse(summary.hasProfile());
+        assertTrue(summary.queryPlan().isEmpty());
+        assertTrue(summary.queryProfile().isEmpty());
```
injectives, line 92:
```suggestion
        assertFalse(summary.hasPlan());
        assertFalse(summary.hasProfile());
        assertTrue(summary.queryPlan().isEmpty());
        assertTrue(summary.queryProfile().isEmpty());
```

It probably makes sense to keep testing the deprecated methods too while they are available.

testkit-backend/src/main/java/neo4j/org/testkit/backend/messages/requests/SummaryUtil.java
```diff
@@ -154,26 +158,40 @@ private static Summary.Plan toPlan(Plan plan) {
...
+                .pageCacheHits(boxOptionalLong(profile.pageCacheHits()))
+                .pageCacheMisses(boxOptionalLong(profile.pageCacheMisses()))
+                .pageCacheHitRatio(boxOptionalDouble(profile.pageCacheHitRatio()))
+                .time(boxOptionalProfileTime(profile.time()))
+                .children(
+                        profile.children().stream().map(SummaryUtil::toProfile).collect(Collectors.toList()))
                 .build();
     }
 
+    private static Long boxOptionalLong(OptionalLong value) {
```
injectives: Same as https://github.com/neo4j/neo4j-java-driver/pull/1730/changes#r3350906825

driver/src/main/java/org/neo4j/driver/summary/ProfiledPlan.java
injectives: For some reason it is still showing a change here.

robsdedude: Must have missed it. Sorry. Now it should be gone for good.

driver/src/main/java/org/neo4j/driver/summary/ResultSummary.java
```diff
@@ -85,9 +92,34 @@ public interface ResultSummary {
...
     ProfiledPlan profile();
 
+    /**
+     * This describes how the database will execute your query.
+     *
+     * @return query plan for the executed query if available
+     * @since 6.2.0
+     */
+    default Optional<Plan> queryPlan() {
+        return Optional.empty();
```
injectives:
```suggestion
        return Optional.ofNullable(plan());
```

driver/src/main/java/org/neo4j/driver/summary/ResultSummary.java
```diff
@@ -85,9 +92,34 @@ public interface ResultSummary {
...
+     * This describes how the database did execute your query.
+     * <p>
+     * If the query you executed was profiled, the query plan will contain detailed information about
+     * what each step of the plan did. That more in-depth version of the query plan becomes available here.
+     *
+     * @return profiled query plan for the executed query if available
+     * @since 6.2.0
+     */
+    default Optional<QueryProfile> queryProfile() {
+        return Optional.empty();
```
injectives:
```suggestion
        throw new UnsupportedOperationException("queryProfile is not supported");
```

injectives (approved):

PR #1732, opened 2026-03-16
feat(resolver): Send resolved address in ROUTE routing context

injectives:
Sending resolved address in ROUTE routing context helps when connecting to DBMS that has `dbms.routing.default_router=SERVER` mode enabled as returned routing table includes the address that the driver could establish connection with.


MaxAake (approved):
✅

PR #1733, opened 2026-03-17
TestKit backend: omit profile's hasPageCacheStats

injectives:
When serializing a `ProfiledPlan` in the TestKit backend, skip `hasPageCacheStats` as that's not corresponding to a key in the raw profile as transmitted via bolt.


MaxAake (approved):
✅

PR #1738, opened 2026-03-27
perf: Use a single BoltConnectionProvider for all connection pools

injectives:
Using a single `BoltConnectionProvider` is sufficient.


MaxAake (approved):
✅

PR #1740, opened 2026-03-30
fix(query): Allow null parameters map

injectives:
This update makes sure the following does not throw an exception:
```java
new Query("MATCH (n) RETURN n", (Map<String, Object>) null)
```

It matches the existing handling for `Value` parameters:
```java
new Query("MATCH (n) RETURN n", (Value) null)
```

Part of: DRIVERS-299


MaxAake (approved):
✅

PR #1751, opened 2026-04-14
fix(bolt): unpack DateTimeZoneId with valid zone handling

injectives:
This update fixes a bug where an incorrect `ZonedDateTime` could be returned, especially during DST periods.


MaxAake (approved):
✅

PR #1755, opened 2026-04-14
build(deps): update dependencies

injectives:


MaxAake (approved):
The flaky 4.2 test is more likely to be related to the previous commit, so this is approved

injectives:
> The flaky 4.2 test is more likely to be related to the previous commit, so this is approved

No, it has been flaky for a while. Needs a separate investigation.

PR #1759, opened 2026-04-29
feat(idempotent): add support for autocommit retries

injectives:
This update introduces automatic retries for auto-commit queries executed via session `run` methods. At present, the driver performs a single immediate retry, and only for idempotent errors.

This behaviour is enabled by default, but can be disabled at both the driver and session levels.

Driver configuration:
```java
var config = Config.builder()
        .withAutoCommitRetriesDisabled(true) // Defaults to false
        .build();
```

Session configuration:
```java
var config = SessionConfig.builder()
        .withAutoCommitRetriesMode(AutoCommitRetriesMode.DISABLED) // Defaults to AutoCommitRetriesMode.DEFAULT that delegates to driver configuration
        .build();
```

Closes: DRIVERS-308

### DIFF
--- a/driver/src/main/java/org/neo4j/driver/internal/async/NetworkSession.java
+++ b/driver/src/main/java/org/neo4j/driver/internal/async/NetworkSession.java
@@ -299,9 +299,7 @@ public class NetworkSession {
             return CompletableFuture.completedStage(null)
                     .thenCompose(ignored -> {
                         var messages = new ArrayList<Message>(2);
-                        apiTelemetryWork
-                                .getTelemetryMessageIfEnabled(connection)
-                                .ifPresent(messages::add);
+                        telemetryEnabled.ifPresent(messages::add);
                         messages.add(newRunMessage(connection, query, parameters, config));
                         return connection.writeAndFlush(responseHandler, messages, parentObservation);
                     })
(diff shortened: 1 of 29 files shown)


driver/src/main/java/org/neo4j/driver/internal/async/NetworkSession.java
```diff
@@ -219,51 +266,85 @@ public CompletionStage<RxResultCursor> runRx(
...
+                    this::handleNewBookmark,
+                    runFailed,
+                    this::handleDatabaseName,
+                    apiTelemetryWork);
+            return CompletableFuture.completedStage(null)
+                    .thenCompose(ignored -> {
+                        var messages = new ArrayList<Message>(2);
+                        apiTelemetryWork
+                                .getTelemetryMessageIfEnabled(connection)
+                                .ifPresent(messages::add);
```
MaxAake:
```suggestion
                        telemetryEnabled
                                .ifPresent(messages::add);
```
Unless I'm missing something, this looks like a double call to `getTelemetryMessageIfEnabled()`

MaxAake (changes requested):
One small comment

driver/src/main/java/org/neo4j/driver/internal/async/NetworkSession.java
injectives: Thanks!

MaxAake (approved):
✅

PR #1762, opened 2026-05-06
Ensure deep array quality for underlying vector array

gjmwoods:


driver/src/main/java/org/neo4j/driver/internal/AbstractArrayVector.java
```diff
@@ -81,4 +82,16 @@ private T arraycopy(T elements) {
...
+        if (array instanceof byte[] value) return Arrays.hashCode(value);
+        if (array instanceof short[] value) return Arrays.hashCode(value);
+        if (array instanceof int[] value) return Arrays.hashCode(value);
+        if (array instanceof long[] value) return Arrays.hashCode(value);
+        if (array instanceof float[] value) return Arrays.hashCode(value);
+        if (array instanceof double[] value) return Arrays.hashCode(value);
+        if (array instanceof char[] value) return Arrays.hashCode(value);
+        if (array instanceof boolean[] value) return Arrays.hashCode(value);
+        return Arrays.deepHashCode((Object[]) array);
+    }
```
injectives:
```suggestion

    abstract int elementsHashCode();
```

I think it may be better to have an abstract method here.

driver/src/main/java/org/neo4j/driver/internal/AbstractArrayVector.java
```diff
@@ -61,12 +62,12 @@ public boolean equals(Object o) {
...
         return length == that.length
                 && Objects.equals(elementType, that.elementType)
-                && Objects.equals(elements, that.elements);
+                && Objects.deepEquals(elements, that.elements);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(elementType, length, elements);
+        return Objects.hash(elementType, length, arrayHashCode(elements));
```
injectives:
```suggestion
        return Objects.hash(elementType, length, elementsHashCode());
```

driver/src/main/java/org/neo4j/driver/internal/AbstractArrayVector.java
```diff
@@ -61,12 +62,12 @@ public boolean equals(Object o) {
         var that = (AbstractArrayVector<?>) o;
         return length == that.length
                 && Objects.equals(elementType, that.elementType)
-                && Objects.equals(elements, that.elements);
+                && Objects.deepEquals(elements, that.elements);
```
injectives:
```suggestion
                && elementsEquals();
```

As we are going for `abstract` method for `hashcode`, it may be worth having a similar one for elements equality and to use the respective `Arrays.equals()` for implementations.

injectives (approved):

PR #1766, opened 2026-06-09
feat(uuid): add support for Neo4j UUID and Bolt 6.1

injectives:
This update brings support for Neo4j UUID value and Bolt 6.1 that is required to support it.

Example:

```java
var result = driver.executableQuery("RETURN $value")
        .withParameters(Map.of("value", UUID.randomUUID())) // java.util.UUID value
        .execute();

var uuidValue = result.records().get(0).get("$value");
uuidValue.type(); // TypeSystem.UUID()
uuidValue.asUUID(); // java.util.UUID value
```

Resolves:  DRIVERS-363


MaxAake (approved):
✅

PR #1777, opened 2026-08-04
fix(uuid): Fix `Value#asObject()` of `UUID` value

injectives:
This update fixes a bug in `Value#asObject()` implementation of `UUID` value that attempted to map it to a `String`.


MaxAake (approved):
✅

PR #1781, opened 2026-08-17
feat(prop-enc): introduce Neo4j Property Encryption

injectives:
This update introduces End-to-End Neo4j Property Encryption, a new preview feature that allows Neo4j properties to be encrypted and decrypted on the driver side.

Property values are encrypted and encoded into bytes. The resulting bytes represent the encrypted value and its associated metadata. Users can choose how and where to store those bytes, provided the chosen representation is supported as a Neo4j property type. To decrypt a value, users retrieve the encoded bytes from the format in which they stored them and provide those bytes to the driver for decryption.

The feature is designed to be easy to use, with an opinionated approach to encryption and decryption. All official Neo4j drivers are expected to support the feature, enabling interoperability between drivers when they are configured with compatible encryption profiles and key management components.

## Property Encryption

Neo4j Property Encryption is exposed through `PropertyEncryption`, which is available directly from the `Driver` object. This makes the API easy to access, particularly in existing applications where the driver object is already available.

Asynchronous and reactive variants are also available.

Example:
```java
var propertyEncryption = driver.propertyEncryption(); // PropertyEncryption

propertyEncryption = driver.propertyEncryption(PropertyEncryption.class); // PropertyEncryption, same as above
var asyncPropertyEncryption = driver.propertyEncryption(AsyncPropertyEncryption.class); // AsyncPropertyEncryption
var reactivePropertyEncryption = driver.propertyEncryption(org.neo4j.driver.encryption.reactive.ReactivePropertyEncryption.class); // ReactivePropertyEncryption using java.util.concurrent.Flow types
var reactiveStreamsPropertyEncryption = driver.propertyEncryption(org.neo4j.driver.encryption.reactivestreams.ReactivePropertyEncryption.class); // ReactivePropertyEncryption using org.reactivestreams types
```

The following example shows how values are encrypted and decrypted. The concepts used by the example, including encryption profiles and key aliases, are explained in the sections below.

Values are encrypted by creating a `PropertyEncryptionRequest` and providing it to `PropertyEncryption` or one of its alternative variants. `PropertyEncryptionRequest#builder()` returns a staged builder for creating a `PropertyEncryptionRequest`.

Example:

```java
var secretValue = "secret";
var userId = 1000;
var encryptionRequest = PropertyEncryptionRequest.builder()
        .fromValue(secretValue) // the value to encrypt
        .withAAD(userId) // an optional AAD
        .usingKeyAlias("users-key") // the key to encrypt with, this is explained later
        .build();
[...]

PR #1782, opened 2026-09-18
build(testkit): base image on Debian bookworm

RichardIrons-neo4j:
Switch from bullseye to bookworm.


injectives:
@RichardIrons-neo4j, thank you! As discussed, #1781 included a similar update.

